PR #171, opened 2020-06-11
ParlAI Chat Demo Components Refactor

pringshia:
This PR is a heavy refactor of the front-end code for ParlAI chat live tasks.

# Changes:

### JS
- Introduced constants in place of strings, e.g. `CONNECTION_STATUS.DISCONNECTED_SERVER` and `INPUT_MODE.WAITING`.
- Moved several components into separate files to make `Core.jsx` less overwhelming
- Wrapper components such as LeftPane now act as true wrapper components by rendering `children`. This makes usage more readable and compositional:

    ```jsx
    <LeftPane stretch={appSettings.isCoverPage}>
      <CustomTaskDescription
        chatTitle={taskConfig.chat_title}
        taskDescription={taskConfig.task_description}
      />
    </LeftPane>
    <RightPane>
      <ChatPane inputMode={inputMode} messages={messages} />
      <ResponsePane inputMode={inputMode} onMessageSend={onMessageSend} />
    </RightPane>
    ```

- Changed pythonic `snake_case` to conventional React `camelCase` in appropriate places.
- Introduced React context to avoid intense prop-drilling happening in previous code. Child components can now call `useContext` to get at global state.
    - **`useContext(MephistoContext)`** gives props available via the useMephistoTask hook
    - **`useContext(AppContext)`** gives global app-specific state, like `taskContext` and `appSettings.volume`

### Styles
- Updated scrolling/resize functionality to be purely CSS instead of JS using flexbox and absolution positioning on the scrollable flex child.
- Moved some inline styles to CSS files
- Cleaned up the height code so the pane height only needs to be specified once on the container element (instead of also on the `LeftPane` and `RightPane`)

### Dev-Experience
- Added ESLint to `mephisto-task`
- Added ESLint to the `parlai-chat-demo` example project

# Future Work

- I want to move `MephistoContext` into the `mephisto-task` package. I took a stab at it but this messed up rendering and on every heart beat, the whole app was being re-rendered, causing things like the chat input to get wiped out on every heartbeat. After a two-hour timebox, i'm deferring this to be tackled in a separate PR.
- Look into removing the build process from the example folder, and only house custom code in the example folder instead.
- Convert some of the other remaining components to be pure function components as well.


examples/parlai_chat_task_demo/webapp/src/app.jsx
```diff
@@ -164,34 +150,37 @@ function MainApp() {
...
         }}
-        allDoneCallback={() => {
-          destroy();
-          handleSubmit({});
+      >
+        <div className="container-fluid" id="ui-container">
+          <BaseFrontend
+            inputMode={inputMode}
+            messages={messages}
+            onMessageSend={handleMessageSend}
```
JackUrb, line 169: One question - why are these props rather than app context? I'm not sure when you should be writing one verses the other.

pringshia: Good question. This is largely subjective, and you can use either approach to send data down.

I chose these as the props in this case because I like the reduction of the app as a mental model as a simple chat container... it shows messages, it calls a callback on new message sends, and it has a prop to toggle it's "mode". All of the other props seemed "contextual" and not relevant to the core functionality of the app at a reductionist level, so I moved those to the app context. I can also see `onTaskComplete` being here but I kept it in the context since I found myself having to do a lot of prop drilling again to get it to the components that need it.

pringshia:
Good point, let me also copy these changes over into `server/blueprints/parlai_chat/`.

PR #172, opened 2020-06-11
[WIP] dynamic launch_units with an issue of non-updating status

saraEbrahim:
I modified the `launch_units` in `task_launcher.py` to be limited with a `max_num_concurrent_units` attribute in `TaskLauncher`.
The idea is to keep track of launched and un-launched units in dictionary (to be able to pop the `COMPLETED` units from the launched dictionary).

What I'm trying to know is when the task is changed its status from `CREATED` to `COMPLETED` because it works fine in the first 2 units but the 2nd unit fails to update its status to `COMPLETED` even if the task physically done (open the url, choose the rating in `static_test_script.py` test case and push 'Submit').

### DIFF
--- a/mephisto/core/local_database.py
+++ b/mephisto/core/local_database.py
@@ -228,7 +228,7 @@ class LocalMephistoDB(MephistoDB):
     """
 
     def __init__(self, database_path=None):
-        logger.info(f"database path: {database_path}")
+        logger.debug(f"database path: {database_path}")
         self.conn: Dict[int, Connection] = {}
         self.table_access_condition = threading.Condition()
         super().__init__(database_path)
--- a/mephisto/core/local_database.py
+++ b/mephisto/core/local_database.py
@@ -228,6 +228,7 @@ class LocalMephistoDB(MephistoDB):
     """
 
     def __init__(self, database_path=None):
+        logger.info(f"database path: {database_path}")
         self.conn: Dict[int, Connection] = {}
         self.table_access_condition = threading.Condition()
         super().__init__(database_path)
--- a/mephisto/core/task_launcher.py
+++ b/mephisto/core/task_launcher.py
@@ -46,7 +46,7 @@ class TaskLauncher:
         db: "MephistoDB",
         task_run: "TaskRun",
         assignment_data_list: List[InitializationData],
-        max_num_concurrent_units: int = 2,
+        max_num_concurrent_units: int = 0,
     ):
         """Prepare the task launcher to get it ready to launch the assignments"""
         self.db = db
--- a/mephisto/core/task_launcher.py
+++ b/mephisto/core/task_launcher.py
@@ -10,7 +10,12 @@
 # interface could be like an iterator. This class will launch tasks
 # as if the loader is an iterator.
 
-from mephisto.data_model.assignment import Assignment, Unit, InitializationData
+from mephisto.data_model.assignment import (
+    Assignment,
+    Unit,
+    InitializationData,
+    AssignmentState,
+)
 
 from typing import Dict, Optional, List, Any, TYPE_CHECKING
 
@@ -20,6 +25,7 @@ if TYPE_CHECKING:
     from mephisto.data_model.task import TaskRun
     from mephisto.data_model.database import MephistoDB
 
+import threading
 from mephisto.core.logger_core import get_logger
 
 logger = get_logger(name=__name__, verbose=True, level="info")
@@ -37,6 +43,7 @@ class TaskLauncher:
         db: "MephistoDB",
         task_run: "TaskRun",
         assignment_data_list: List[InitializationData],
+        max_num_concurrent_units: int = 1,
     ):
         """Prepare the task launcher to get it ready to launch the assignments"""
         self.db = db
@@ -45,6 +52,10 @@ class TaskLauncher:
         self.assignments: List[Assignment] = []
         self.units: List[Unit] = []
         self.provider_type = task_run.get_provider().PROVIDER_TYPE
+        self.max_num_concurrent_units = max_num_concurrent_units
+        self.num_running_units = 0
+        self.launched_units: Dict[Any, Any] = {}
+        self.unlaunched_units: Dict[Any, Any] = {}
 
         run_dir = task_run.get_run_dir()
         os.makedirs(run_dir, exist_ok=True)
@@ -83,10 +94,39 @@ class TaskLauncher:
                 )
                 self.units.append(Unit(self.db, unit_id))
 
+        for unit in self.units:
+            self.unlaunched_units[unit.db_id] = unit
+
+    def generate_units(self):
+        for unit in self.units:
+            yield unit
+
+    def _launch_limited_units(self, units_generator, url):
+        while len(self.unlaunched_units.keys()) != 0:
+            if self.num_running_units < self.max_num_concurrent_units:
+                unit = next(units_generator)
+                self.launched_units[unit.db_id] = unit
+                self.unlaunched_units.pop(unit.db_id)
+                unit.launch(url)
+                self.num_running_units += 1
+            for _, unit in self.launched_units.copy().items():
+                if unit.get_status() == AssignmentState.COMPLETED:
+                    self.num_running_units -= 1
+                    self.launched_units.pop(unit.db_id)
+
     def launch_units(self, url: str) -> None:
         """launch any units registered by this TaskLauncher"""
-        for unit in self.units:
-            unit.launch(url)
+        # define units generator
+        units_generator = self.generate_units()
+
+        if self.max_num_concurrent_units == 0:  # launch all units (no limit)
+            for _ in self.units:
+                next(units_generator).launch(url)
+        else:  # start the thread that launches units according to the max_unit limit
+            thread = threading.Thread(
+                target=self._launch_limited_units, args=(units_generator, url)
+            )
+            thread.start()
 
     def expire_units(self) -> None:
         """Clean up all units on this TaskLauncher"""
--- a/mephisto/core/task_launcher.py
+++ b/mephisto/core/task_launcher.py
@@ -32,6 +32,8 @@ from mephisto.core.logger_core import get_logger
 
 logger = get_logger(name=__name__, verbose=True, level="info")
 
+UNIT_GENERATOR_WAIT_SECONDS = 10
+
 
 class TaskLauncher:
     """
@@ -45,7 +47,7 @@ class TaskLauncher:
         db: "MephistoDB",
         task_run: "TaskRun",
         assignment_data_list: List[InitializationData],
-        max_num_concurrent_units: int = 1,
+        max_num_concurrent_units: int = 2,
     ):
         """Prepare the task launcher to get it ready to launch the assignments"""
         self.db = db
@@ -55,7 +57,7 @@ class TaskLauncher:
         self.units: List[Unit] = []
         self.provider_type = task_run.get_provider().PROVIDER_TYPE
         self.max_num_concurrent_units = max_num_concurrent_units
-        self.launched_units: List[Unit] = []
+        self.launched_units: Dict[str, Unit] = {}
         self.unlaunched_units: List[Unit] = []
 
         run_dir = task_run.get_run_dir()
@@ -100,25 +102,27 @@ class TaskLauncher:
         """ units generator which checks that only 'max_num_concurrent_units' running at the same time,
         i.e. in the LAUNCHED or ASSIGNED states """
         while True:
-            for i in range(len(self.launched_units)):
-                unit = self.launched_units[i]
+            units_id_to_remove = []
+            for db_id, unit in self.launched_units.items():
                 status = unit.get_status()
                 if (
                     status != AssignmentState.LAUNCHED
                     and status != AssignmentState.ASSIGNED
                 ):
-                    self.launched_units.pop(i)
+                    units_id_to_remove.append(db_id)
+            for db_id in units_id_to_remove:
+                self.launched_units.pop(db_id)
 
             num_avail_units = self.max_num_concurrent_units - len(self.launched_units)
             for i in range(len(self.unlaunched_units)):
                 if i < num_avail_units:
                     unit = self.unlaunched_units[i]
-                    self.launched_units.append(unit)
+                    self.launched_units[unit.db_id] = unit
                     self.unlaunched_units.pop(i)
                     yield unit
                 else:
                     break
-            time.sleep(10)
+            time.sleep(UNIT_GENERATOR_WAIT_SECONDS)
             if not self.unlaunched_units:
                 break
 
--- a/mephisto/core/task_launcher.py
+++ b/mephisto/core/task_launcher.py
@@ -35,16 +35,6 @@ logger = get_logger(name=__name__, verbose=True, level="info")
 UNIT_GENERATOR_WAIT_SECONDS = 10
 
 
-class LimitedDict(dict):
-    def __init__(self, limit):
-        self.limit = limit
-        super().__init__()
-
-    def __setitem__(self, key, value):
-        super().__setitem__(key, value)
-        assert self.limit, len(self.keys())
-
-
 class TaskLauncher:
     """
     This class is responsible for managing the process of registering
@@ -57,7 +47,7 @@ class TaskLauncher:
         db: "MephistoDB",
         task_run: "TaskRun",
         assignment_data_list: List[InitializationData],
-        max_num_concurrent_units: int = 2,
+        max_num_concurrent_units: int = 0,
     ):
         """Prepare the task launcher to get it ready to launch the assignments"""
         self.db = db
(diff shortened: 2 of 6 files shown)


mephisto/core/task_launcher.py
```diff
@@ -83,13 +96,46 @@ def create_assignments(self) -> None:
...
         for unit in self.units:
+            self.unlaunched_units[unit.db_id] = unit
+
+    def generate_units(self):
+        """ units generator which checks that only 'max_num_concurrent_units' running at the same time,
+        i.e. in the LAUNCHED or ASSIGNED states """
+        while True:
+            for _, launched_unit in self.launched_units.copy().items():
+                if (
+                    launched_unit.get_status() != AssignmentState.LAUNCHED
```
JackUrb, line 108: `get_status()` incurs at the least a database hit, and at the current most a round trip to determine the status of an MTurk. Prefer getting this value into a variable first and then comparing.

mephisto/core/task_launcher.py
```diff
@@ -45,6 +54,10 @@ def __init__(
         self.assignments: List[Assignment] = []
         self.units: List[Unit] = []
         self.provider_type = task_run.get_provider().PROVIDER_TYPE
+        self.max_num_concurrent_units = max_num_concurrent_units
+        self.num_running_units = 0
```
JackUrb, line 58: We don't need this value, as we can always check `len(self.launched_units)` and builtins already do this bookkeeping to keep the runtime of `len` constant and low.

mephisto/core/task_launcher.py
```diff
@@ -45,6 +54,10 @@ def __init__(
...
+        self.launched_units: Dict[Any, Any] = {}
```
JackUrb, line 59: Would be a `Dict[str, Unit]` as used here.

mephisto/core/task_launcher.py
```diff
@@ -45,6 +54,10 @@ def __init__(
...
+        self.unlaunched_units: Dict[Any, Any] = {}
```
JackUrb, line 60: As noted below, this may be better as a `list`.

mephisto/core/task_launcher.py
```diff
@@ -83,13 +96,46 @@ def create_assignments(self) -> None:
...
             unit.launch(url)
 
+    def launch_units(self, url: str) -> None:
+        """launch any units registered by this TaskLauncher"""
+        thread = threading.Thread(target=self._launch_limited_units, args=(url,))
+        thread.start()
+
     def expire_units(self) -> None:
         """Clean up all units on this TaskLauncher"""
+
```
JackUrb, line 138: remove newline.

mephisto/core/task_launcher.py
```diff
@@ -83,13 +96,46 @@ def create_assignments(self) -> None:
...
+                    self.launched_units.pop(launched_unit.db_id)
+
+            num_avail_units = self.max_num_concurrent_units - self.num_running_units
+            for i, item in enumerate(self.unlaunched_units.copy().items()):
+                db_id, unit = item
+                if i < num_avail_units:
+                    self.launched_units[unit.db_id] = unit
+                    self.unlaunched_units.pop(unit.db_id)
+                    self.num_running_units += 1
+                    yield unit
```
JackUrb, line 121: This is definitely functional, but rather than incurring the cost of a dict copy, you could instead bound your loop on `range(num_avail_units)` and select that many units. Alternately you could have a break in the `else` condition of the given if (instead of continuing the loop).

Ultimately though, this code is simplified by replacing the `unlaunched_units` dict with a list. We don't have an explicit need to directly index specific units from the list of units currently unlaunched, and as such it's a bag that we can take any value out of. Using a `list` then lets you simply use `unlaunched_units.pop()` for `num_avail_units` times.

mephisto/core/task_launcher.py
```diff
@@ -83,13 +96,46 @@ def create_assignments(self) -> None:
...
+            time.sleep(10)
```
JackUrb, line 122: Prefer to keep magic numbers in constants near the top of the file.

mephisto/data_model/task.py
```diff
@@ -261,7 +261,12 @@ def get_valid_units_for_worker(self, worker: "Worker") -> List["Unit"]:
             is_self_set = map(lambda u: u.worker_id == worker.db_id, unit_set)
             if not any(is_self_set):
                 units += unit_set
-        valid_units = [u for u in units if u.get_assigned_agent() is None]
+        valid_units = [
+            u
+            for u in units
+            if u.get_assigned_agent() is None
```
JackUrb, line 267: You don't need to check `u.get_assigned_agent()` being `None`, as `get_assigned_agent` will only ever return `None` for `Unit`'s that are in the `CREATED` and `LAUNCHED` statuses. As such, this check is redundant.

mephisto/core/task_launcher.py
JackUrb: Newline is still here

mephisto/core/task_launcher.py
JackUrb: Constant is still here. A definition for `UNIT_GENERATOR_WAIT_SECONDS = 10` would be appropriate near the head of the file.

mephisto/core/task_launcher.py
```diff
@@ -82,14 +94,47 @@ def create_assignments(self) -> None:
...
+        i.e. in the LAUNCHED or ASSIGNED states """
+        while True:
+            for i in range(len(self.launched_units)):
+                unit = self.launched_units[i]
+                status = unit.get_status()
+                if (
+                    status != AssignmentState.LAUNCHED
+                    and status != AssignmentState.ASSIGNED
+                ):
+                    self.launched_units.pop(i)
```
JackUrb, line 110: This loop isn't quite doing what you want it to do, and has a chance of running over the end. Take the instance where you have 4 launched units, and the first 2 are being popped. Your list will go through the following steps:
```
i = 0
[<completed unit>, <completed_unit>, <launched unit>, <launched unit>]
           ^
pop unit 0

i = 1
[<completed_unit>, <launched unit>, <launched unit>]
                           ^
# We skipped one of the completed units!

i = 2
[<completed_unit>, <launched unit>, <launched unit>]
                                          ^

i = 3
[<completed_unit>, <launched unit>, <launched unit>]
                                                      ^
# We're out of bounds!!
```

Your original implementation of `launched_units` using `dict`s was actually fine, the only one that needed to change was `unlaunched_units`. It's important to consider whether or not direct indexing is important or useful when making choices on which data structure you need. 

For `unlaunched_units`, order doesn't matter, and it doesn't matter which one we take. Thus a `list` is appropriate as we can always remove from the front for free, and we can create slices for free as well. 
For `launched_units`, we don't know when a unit will be completed, and we don't know from what index we'll need to remove from the `launched_units` data structure. Removing from a random index in a list costs `n`, and can lead to array index bugs as you have here. Removing from a `dict` is constant time. Admittedly you run into an issue where you can't just iterate over the dict's values due to the bounds changing, _but_ you can avoid this by having the iterating loop _find_ the elements to remove (creating a temporary list of `db_id`s to pop), and then iterating over that list immediately after. That prevents you from needing to duplicate the dict every iteration.

mephisto/core/task_launcher.py
```diff
@@ -10,20 +10,40 @@
...
+
+class LimitedDict(dict):
+    def __init__(self, limit):
+        self.limit = limit
+        super().__init__()
+
+    def __setitem__(self, key, value):
+        super().__setitem__(key, value)
+        assert self.limit, len(self.keys())
+
```
JackUrb, line 46: You can initialize this in the test rather than here.

mephisto/core/task_launcher.py
```diff
@@ -10,20 +10,40 @@
...
+
+class LimitedDict(dict):
+    def __init__(self, limit):
+        self.limit = limit
+        super().__init__()
+
+    def __setitem__(self, key, value):
+        super().__setitem__(key, value)
+        assert self.limit, len(self.keys())
```
JackUrb, line 45: I'm not sure what you're trying to do with this assert - the `assert` syntax is `assert (conditional), (message for if conditional fails)`

mephisto/core/task_launcher.py
```diff
@@ -37,6 +57,7 @@ def __init__(
         db: "MephistoDB",
         task_run: "TaskRun",
         assignment_data_list: List[InitializationData],
+        max_num_concurrent_units: int = 2,
```
JackUrb, line 60: This should default to 0, such that the feature isn't used when not asked for.

mephisto/core/local_database.py
```diff
@@ -228,6 +228,7 @@ class LocalMephistoDB(MephistoDB):
     """
 
     def __init__(self, database_path=None):
+        logger.info(f"database path: {database_path}")
```
JackUrb, line 231: Nit: this is probably `logger.debug` level.

mephisto/core/task_launcher.py
```diff
@@ -10,20 +10,30 @@
...
+    Unit,
+    InitializationData,
+    AssignmentState,
+)
 
 from typing import Dict, Optional, List, Any, TYPE_CHECKING
 
 import os
+import time
+from datetime import datetime
```
JackUrb, line 24: Nit: unneeded

mephisto/core/task_launcher.py
```diff
@@ -82,14 +96,57 @@ def create_assignments(self) -> None:
                     task_run.sandbox,
                 )
                 self.units.append(Unit(self.db, unit_id))
+                self.unlaunched_units[unit_id] = Unit(self.db, unit_id)
+
+    def generate_units(self):
+        """ units generator which checks that only 'max_num_concurrent_units' running at the same time,
+        i.e. in the LAUNCHED or ASSIGNED states """
+        while True:
```
JackUrb, line 104: Generally it's good to avoid `while True` loops - here it'd be worthwhile to have a flag as to whether or not `expire_units` has been called, and exit this loop if it has.

test/core/test_task_launcher.py
```diff
@@ -52,10 +65,12 @@ def test_init_on_task_run(self):
...
         self.assertEqual(launcher.provider_type, MockProvider.PROVIDER_TYPE)
 
-    def test_create_launch_expire_assignments(self):
+    def test_create_expire_assignments(self):
         """Initialize a launcher on a task run, then create the assignments"""
         mock_data_array = self.get_mock_assignment_data_array()
-        launcher = TaskLauncher(self.db, self.task_run, mock_data_array)
+        launcher = TaskLauncher(
+            self.db, self.task_run, mock_data_array, max_num_concurrent_units=1
+        )
```
JackUrb, line 73: You shouldn't be editing existing tests, but rather adding new ones to test the new functionality.

saraEbrahim: yeah the additional parameter was meant for a test case in an existing one. But there were few lines in this test which fails the new way of units' statuses (it tested whether all of the units had `LAUNCHED` status after calling `launcher.launch_units()`

test/core/test_task_launcher.py
```diff
@@ -74,20 +87,31 @@ def test_create_launch_expire_assignments(self):
         for assignment in launcher.assignments:
             self.assertEqual(assignment.get_status(), AssignmentState.CREATED)
 
-        launcher.launch_units("dummy-url:3000")
-
-        for unit in launcher.units:
-            self.assertEqual(unit.get_db_status(), AssignmentState.LAUNCHED)
-        for assignment in launcher.assignments:
-            self.assertEqual(assignment.get_status(), AssignmentState.LAUNCHED)
-
```
JackUrb, line 83: Nit - these should be restored, and the test should still be `test_create_launch_expire_assignments`

test/core/test_task_launcher.py
```diff
@@ -74,20 +87,31 @@ def test_create_launch_expire_assignments(self):
...
         launcher.expire_units()
 
         for unit in launcher.units:
             self.assertEqual(unit.get_db_status(), AssignmentState.EXPIRED)
         for assignment in launcher.assignments:
             self.assertEqual(assignment.get_status(), AssignmentState.EXPIRED)
 
+    def test_launch_assignments(self):
```
JackUrb, line 97: A better name would be `test_launch_assignments_with_concurrent_unit_cap`.

test/core/test_task_launcher.py
```diff
@@ -74,20 +87,31 @@ def test_create_launch_expire_assignments(self):
...
+        """Initialize a launcher on a task run, then create the assignments"""
+        mock_data_array = self.get_mock_assignment_data_array()
+        launcher = TaskLauncher(
+            self.db, self.task_run, mock_data_array, max_num_concurrent_units=1
+        )
+        launcher.launched_units = LimitedDict(launcher.max_num_concurrent_units)
+        launcher.create_assignments()
+        launcher.launch_units("dummy-url:3000")
+
+        for unit in launcher.units:
```
JackUrb, line 107: This should be looping until all of the units are in the `COMPLETED` state, asserting that it's taken less than some amount of time to get there.

```
while not fully complete
  for all units
    if unit is launched, set it to complete
  sleep
  assert not over time limit
```

test/core/test_task_launcher.py
```diff
@@ -74,20 +87,31 @@ def test_create_launch_expire_assignments(self):
...
         for unit in launcher.units:
             self.assertEqual(unit.get_db_status(), AssignmentState.EXPIRED)
         for assignment in launcher.assignments:
             self.assertEqual(assignment.get_status(), AssignmentState.EXPIRED)
 
+    def test_launch_assignments(self):
+        """Initialize a launcher on a task run, then create the assignments"""
+        mock_data_array = self.get_mock_assignment_data_array()
+        launcher = TaskLauncher(
+            self.db, self.task_run, mock_data_array, max_num_concurrent_units=1
```
JackUrb, line 101: I still think it's worthwhile to test with `max_num_concurrent_units` as 2, and also as 4 and 5. You can put these all in a loop so long as you call `tearDown` and `setUp` at the end of each iteration.

test/core/test_task_launcher.py
```diff
@@ -74,20 +87,31 @@ def test_create_launch_expire_assignments(self):
...
+        )
+        launcher.launched_units = LimitedDict(launcher.max_num_concurrent_units)
+        launcher.create_assignments()
+        launcher.launch_units("dummy-url:3000")
+
+        for unit in launcher.units:
+            if unit.get_status() == AssignmentState.LAUNCHED:
+                unit.set_db_status(AssignmentState.COMPLETED)
+                time.sleep(10)
```
JackUrb, line 110: This internal sleep can be much shorter (`0.1`) if you're looping on the condition of advancing any launched units to completing until all are completed.

test/core/test_task_launcher.py
saraEbrahim: for some reason `tearDown` crashes with database `KeyError` in ln 256 `core/local_database.py` but works fine without calling `setUp` only.

JackUrb: When does this crash occur - you'll need to call `tearDown` right before `setUp` or else the test is going to leave some excess content in the `tmp/` directory.

test/core/test_task_launcher.py
```diff
@@ -88,6 +101,33 @@ def test_create_launch_expire_assignments(self):
...
+                self.db,
+                self.task_run,
+                mock_data_array,
+                max_num_concurrent_units=max_num_units,
+            )
+            launcher.launched_units = LimitedDict(launcher.max_num_concurrent_units)
+            launcher.create_assignments()
+            launcher.launch_units("dummy-url:3000")
+
+            while set([u.get_status() for u in launcher.units]) != {
```
JackUrb, line 119: In theory this test will loop infinitely if the system doesn't work. As such you should get a `start_time` before the loop, and have a `self.assert...` inside the loop ensuring that the current time isn't greater than some amount later than that `start_time`.

test/core/test_task_launcher.py
saraEbrahim: weird .. the error occurs when the `setUp` call is at the beginning of the for loop, but disappears when it's called right before `tearDown()`, why would that happen?

JackUrb: Unsure, but you can think of it as a lock - once `tearDown()` has been called, if you make any calls before `setUp` the resources won't exist to execute the call.

test/core/test_task_launcher.py
```diff
@@ -88,6 +103,37 @@ def test_create_launch_expire_assignments(self):
...
+                AssignmentState.COMPLETED
+            }:
+                for unit in launcher.units:
+                    if unit.get_status() == AssignmentState.LAUNCHED:
+                        unit.set_db_status(AssignmentState.COMPLETED)
+                    time.sleep(0.1)
+                self.assertEqual(launcher.launched_units.exceed_limit, False)
+                curr_time = time.time()
+                if curr_time - start_time > MAX_WAIT_TIME_UNIT_LAUNCH:
```
JackUrb, line 131: this should be an assert, rather than a break. Otherwise the test will pass even if this is exceeded.

test/core/test_task_launcher.py
saraEbrahim: I really missed the last line here. so sorry for that. But glad to know why.

PR #182, opened 2020-06-24
Support for JSON and JSONL file data and Subtasks in Static Tasks

mwillwork:
Submitting this for visibility and so Jing can use it. Feel free to tell me to change it! :)

### DIFF
--- a/mephisto/server/blueprints/abstract/static_task/static_blueprint.py
+++ b/mephisto/server/blueprints/abstract/static_task/static_blueprint.py
@@ -20,6 +20,7 @@ from mephisto.core.registry import register_mephisto_abstraction
 import os
 import time
 import csv
+import json
 
 from typing import ClassVar, List, Type, Any, Dict, Iterable, TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


mephisto/server/blueprints/abstract/static_task/static_blueprint.py
```diff
@@ -135,6 +162,13 @@ def add_args_to_group(cls, group: "ArgumentGroup") -> None:
             default=1,
             type=int,
         )
+        group.add_argument(
+            "--subtasks-per-unit",
+            dest="subtasks_per_unit",
+            type=int,
+            default=5,
+            help="number of subtasks/comparisons to do per unit",
+        )
```
JackUrb, line 171: This last part is probably meant for your multi-subtask-per-unit blueprint.

mephisto/server/blueprints/static_task/static_html_blueprint.py
```diff
@@ -58,17 +58,23 @@ def __init__(self, task_run: "TaskRun", opts: Any):
 
     @classmethod
     def assert_task_args(cls, opts: Any) -> None:
+        print(f'data_json: {opts}')
```
JackUrb, line 61: nit: log

mephisto/server/blueprints/abstract/static_task/static_blueprint.py
mwillwork: Yeah let me know: I would say that having subtasks is a pretty common ask for us, so that's why I didn't put it in the task-specific blueprint I created.

PR #188, opened 2020-06-30
[WIP] adding tasks/assignments dynamically

saraEbrahim:
I've started with Jack's first suggestion in the issue#99: 

- create a main thread that is responsible of running both: `try_launch_units()` and `try_create_assignments()` threads.
- I doubt that `generate_assignment()` with its current version will work if `self.assignment_data_list` is changing its size (which is what I expect as the required task is to allow dynamic size of this list) but I need to think about how to test it to know if it works.

### DIFF
--- a/mephisto/core/operator.py
+++ b/mephisto/core/operator.py
@@ -289,6 +289,7 @@ class Operator:
                 if task_run.get_is_completed():
                     self.supervisor.shutdown_job(tracked_run.job)
                     tracked_run.architect.shutdown()
+                    tracked_run.task_launcher.shutdown()
                     del self._task_runs_tracked[task_run.db_id]
             time.sleep(2)
 
@@ -297,6 +298,7 @@ class Operator:
         self.is_shutdown = True
         for tracked_run in self._task_runs_tracked.values():
             logger.info("expiring units")
+            tracked_run.task_launcher.shutdown()
             tracked_run.task_launcher.expire_units()
         try:
             remaining_runs = self._task_runs_tracked.values()
--- a/mephisto/core/operator.py
+++ b/mephisto/core/operator.py
@@ -270,8 +270,7 @@ class Operator:
             raise e
 
         launcher = TaskLauncher(self.db, task_run, initialization_data_array)
-        launcher.create_assignments()
-        launcher.launch_units(task_url)
+        launcher.start(task_url)
 
         self._task_runs_tracked[task_run.db_id] = TrackedRun(
             task_run=task_run,
--- a/mephisto/core/operator.py
+++ b/mephisto/core/operator.py
@@ -246,9 +246,7 @@ class Operator:
             initialization_data_array = blueprint.get_initialization_data()
             # TODO(#99) extend
             if not isinstance(initialization_data_array, list):
-                raise NotImplementedError(
-                    "Non-list initialization data is not yet supported"
-                )
+                logger.info("Non-list initialization data is now supported")
 
             # Link the job together
             job = self.supervisor.register_job(
@@ -270,7 +268,8 @@ class Operator:
             raise e
 
         launcher = TaskLauncher(self.db, task_run, initialization_data_array)
-        launcher.start(task_url)
+        launcher.create_assignments()
+        launcher.launch_units(task_url)
 
         self._task_runs_tracked[task_run.db_id] = TrackedRun(
             task_run=task_run,
--- a/mephisto/core/operator.py
+++ b/mephisto/core/operator.py
@@ -244,9 +244,6 @@ class Operator:
 
             blueprint = BlueprintClass(task_run, task_args)
             initialization_data_array = blueprint.get_initialization_data()
-            # TODO(#99) extend
-            if not isinstance(initialization_data_array, list):
-                logger.info("Non-list initialization data is now supported")
 
             # Link the job together
             job = self.supervisor.register_job(
--- a/mephisto/core/task_launcher.py
+++ b/mephisto/core/task_launcher.py
@@ -37,9 +37,9 @@ GENERATOR_WAIT_SECONDS = 0.5
 
 
 class GeneratorType(enum.Enum):
-    none = 0
-    unit = 1
-    assignment = 2
+    NONE = 0
+    UNIT = 1
+    ASSIGNMENT = 2
 
 
 class TaskLauncher:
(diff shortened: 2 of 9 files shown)


mephisto/core/operator.py
```diff
@@ -270,8 +270,7 @@ def parse_and_launch_run(
             raise e
 
         launcher = TaskLauncher(self.db, task_run, initialization_data_array)
-        launcher.create_assignments()
-        launcher.launch_units(task_url)
+        launcher.start(task_url)
```
JackUrb, line 273: You shouldn't be updating the flow of how the operator tackles this just yet - your implementation should work under the separated flow (by only handling components that are necessary to run).

mephisto/core/operator.py
```diff
@@ -246,9 +246,7 @@ def parse_and_launch_run(
             initialization_data_array = blueprint.get_initialization_data()
             # TODO(#99) extend
             if not isinstance(initialization_data_array, list):
-                raise NotImplementedError(
-                    "Non-list initialization data is not yet supported"
-                )
+                logger.info("Non-list initialization data is now supported")
```
JackUrb, line 249: You can drop this log now that the support is present.

mephisto/core/task_launcher.py
```diff
@@ -17,22 +17,29 @@
...
 
 UNIT_GENERATOR_WAIT_SECONDS = 10
-ASSIGNMENT_GENERATOR_WAIT_SECONDS = 0.5
+GENERATOR_WAIT_SECONDS = 0.5
+
+
+class GeneratorType(enum.Enum):
+    none = 0
+    unit = 1
+    assignment = 2
```
JackUrb, line 42: Nit - enum keys are usually `ALL_CAPS`.

mephisto/core/task_launcher.py
```diff
@@ -140,28 +174,33 @@ def generate_units(self):
...
-                for unit in self.generate_units():
-                    unit.launch(url)
-            else:
-                for _ in range(len(self.unlaunched_units.items())):
-                    db_id, unit = self.unlaunched_units.popitem()
-                    unit.launch(url)
-                time.sleep(0.5)
+    def _launch_limited_units(self, url: str) -> None:
+        """ use units' generator to launch limited number of units according to (max_num_concurrent_units)"""
+        while self.did_launch_units:
```
JackUrb, line 187: What is this flag actually doing?

mephisto/core/task_launcher.py
```diff
@@ -46,74 +53,95 @@ def __init__(
...
+            self.finished_generators = True
+
+    def create_assignments(self) -> None:
+        """ Create an assignment and associated units for the generated assignment data """
+        if self.generator_type == GeneratorType.none:
+            for data in self.assignment_data_iterable:
+                self._create_single_assignment(data)
+        else:
+            self.did_create_assignments = True
+            self.did_launch_units = True
```
JackUrb, line 144: You can actually have the `assignments` thread be launched in the same way as the `units` thread at this point, as both need to be run separately anyways.

mephisto/core/task_launcher.py
saraEbrahim: it was somehow making sure the units generator keeps working .. but now I realized it is not needed at all. Thanks!

mephisto/core/task_launcher.py
```diff
@@ -45,63 +53,87 @@ def __init__(
...
-                    task_run.provider_type,
-                    task_run.task_type,
-                    task_run.sandbox,
-                )
-                self.units.append(Unit(self.db, unit_id))
+            self.units.append(Unit(self.db, unit_id))
+            with self.unlaunched_units_access_condition:
                 self.unlaunched_units[unit_id] = Unit(self.db, unit_id)
-                self.keep_launching = True
+            self.keep_launching_units = True
```
JackUrb, line 112: This should only be set once when `launch_assignments` is first called, or else an errant call here could lead to a race condition against `expire_units` where the launching continues.

mephisto/core/task_launcher.py
```diff
@@ -45,63 +53,87 @@ def __init__(
...
+
+    def _try_generating_assignments(self) -> None:
+        """ Try to generate more assignments from the assignments_data_iterator"""
+        while not self.finished_generators:
+            try:
+                data = next(self.assignment_data_iterable)
+                self._create_single_assignment(data)
+            except StopIteration:
+                self.finished_generators = True
+            time.sleep(GENERATOR_WAIT_SECONDS)
```
JackUrb, line 122: This should be labelled `ASSIGNMENT_GENERATOR_WAIT_SECONDS` to match the unit one.

mephisto/core/task_launcher.py
```diff
@@ -45,63 +53,87 @@ def __init__(
...
+
+    def create_assignments(self) -> None:
+        """ Create an assignment and associated units for the generated assignment data """
+        if self.generator_type == GeneratorType.NONE:
+            for data in self.assignment_data_iterable:
+                self._create_single_assignment(data)
+        else:
+            main_thread = threading.Thread(target=self._try_generating_assignments, args=())
```
JackUrb, line 130: nit: Variable naming should describe what this thread actually is - this is the `assignment_gen_thread` or something.

mephisto/core/task_launcher.py
saraEbrahim: I'll modify it but can we discuss how the race condition can happen over the meeting?

saraEbrahim:
@JackUrb I added the `thread.join()` for both threads at the end of `expire_units()` but `test/core/test_operator.py` still fails (there are 3 threads not 1 as expected), that's because `expire_units()` in `TaskLauncher` is never called from `operator.py`; at this line:
https://github.com/facebookresearch/Mephisto/blob/10a9d109d3f708a2f91d919964f9cb907710e942/mephisto/core/operator.py#L307
the number of items in `self._task_runs_tracked` dictionary is zero.

JackUrb:
Ah this is a fair point - the correct course of action here would be to move the flag-setting and joining to a new `shutdown()` function, and call that from both `operator.shutdown()` (in the line before the `tracked_run.task_launcher.expire_units()` call) and within `operator._track_and_kill_runs()` after the `tracked_run.architect.shutdown()` call. 

Having a `shutdown` tends to be a better design for cleaning up allocated resources anyways

mephisto/core/task_launcher.py
```diff
@@ -32,8 +32,8 @@
 
 logger = get_logger(name=__name__, verbose=True, level="debug")
 
-UNIT_GENERATOR_WAIT_SECONDS = 10
-ASSIGNMENT_GENERATOR_WAIT_SECONDS = 0.5
+UNIT_GENERATOR_WAIT_SECONDS = 1
+ASSIGNMENT_GENERATOR_WAIT_SECONDS = 0.05
```
JackUrb, line 36: Why change these times?

saraEbrahim: there was a test in the `test_operator.py` making sure that consecutive assignments finish after certain amount of time, when they are as large as before it fails, so I changed it and was about to ask you (how can we define these numbers efficiently? and should the test in the `test_operator.py` be changed later?)

JackUrb: I'd update the `test_operator` to use a timeout style test rather than a direct check, like the style of:
```
start_time = time.time()
while (condition is not met):
  assert time.time - start_time is not more than some amount
```
as the timeouts defined initially here seem better than these really rapid ones.

JackUrb:
So do the operator tests pass now? Have you made that change as well?

saraEbrahim:
Now `test_operator.py` passes .. I just had to increase the `TIMEOUT_TIME` to match the `UNIT_GENERATOR_WAIT_SECONDS`.

saraEbrahim:
@JackUrb  Thank you for bearing with me and mentoring me and teaching me alot :)

PR #192, opened 2020-07-03
Directly softblock MTurk workers

### LINKED ISSUE #183: [Feature Request] Soft-block MTurk workers by MTurk ID directly
The very basic idea is to implement a `MTurkRequester.block_worker_by_mturk_id()` method, and call that. It would rely on the underlying methods in `mturk_utils` and the `MTurkDatastore`

First it would get the client, as in [here](https://github.com/facebookresearch/Mephisto/blob/890a6c94b1fa81198d0cbcf230c69c7f199d186e/mephisto/providers/mturk/mturk_requester.py#L52). Then it would get the corresponding MTurk qualification id from the local map, as in [here](https://github.com/facebookresearch/Mephisto/blob/890a6c94b1fa81198d0cbcf230c69c7f199d186e/mephisto/providers/mturk/mturk_worker.py#L75). Then it can block the worker directly with that, using [this](https://github.com/facebookresearch/Mephisto/blob/master/mephisto/providers/mturk/mturk_utils.py#L260). 

CC @EricMichaelSmith @jxmsML

JackUrb:
# Overview
This includes a utility function that allows directly softblocking MTurk workers. It also includes a script that lets you block from the command line.

Closes #183

cc @mwillwork @EricMichaelSmith @jxmsML 

# Testing
Ran:
```
> python mephisto/scripts/mturk/soft_block_workers_by_mturk_id.py 
```
and assigned a block qualification to myself directly from MTurk worker id.

### DIFF
--- a/mephisto/providers/mturk/utils/script_utils.py
+++ b/mephisto/providers/mturk/utils/script_utils.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import List, Optional
+
+from mephisto.providers.mturk.mturk_utils import give_worker_qualification
+from mephisto.core.local_database import LocalMephistoDB
+from mephisto.data_model.requester import Requester
+
+
+def direct_soft_block_mturk_workers(
+    worker_list: List[str], 
+    soft_block_qual_name: str, 
+    requester_name: Optional[str] = None,
+):
+    db = LocalMephistoDB()
+    reqs = db.find_requesters(requester_name=requester_name, provider_type="mturk")
+    requester = reqs[0]
+
+    mturk_qual_details = requester.datastore.get_qualification_mapping(
+        soft_block_qual_name
+    )
+    if mturk_qual_details is not None:
+        # Overrule the requester, as this qualification already exists
+        requester = Requester(db, mturk_qual_details["requester_id"])
+        qualification_id = mturk_qual_details["mturk_qualification_id"]
+    else:
+        qualification_id = requester._create_new_mturk_qualification(
+            soft_block_qual_name
+        )
+    
+    mturk_client = requester._get_client(requester._requester_name)
+    for worker_id in worker_list:
+        give_worker_qualification(
+            mturk_client,
+            worker_id,
+            qualification_id,
+            value=1,
+        )
+
+
+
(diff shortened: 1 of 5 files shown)


jxmsML:
Thanks for introducing this util!!

mephisto/providers/mturk/utils/script_utils.py
```diff
@@ -0,0 +1,44 @@
...
+def direct_soft_block_mturk_workers(
+    db: "MephistoDB",
+    worker_list: List[str],
+    soft_block_qual_name: str,
+    requester_name: Optional[str] = None,
+):
+    """
+    Directly assign the soft blocking MTurk qualification that Mephisto 
+    associates with soft_block_qual_name to all of the workers in worker_list.
+    If requester_name is not provided, it will use the first mturk requester
```
EricMichaelSmith, line 24: Hmm - is there a usecase for softblocking the first MTurk requester in the database? Otherwise it might be safer to check that `requester_name` is not None

mephisto/providers/mturk/utils/script_utils.py
```diff
@@ -0,0 +1,44 @@
...
+    from mephisto.data_model.database import MephistoDB
+
+def direct_soft_block_mturk_workers(
+    db: "MephistoDB",
+    worker_list: List[str],
+    soft_block_qual_name: str,
+    requester_name: Optional[str] = None,
+):
+    """
+    Directly assign the soft blocking MTurk qualification that Mephisto
```
EricMichaelSmith, line 22: Maybe it's worth noting that `worker_list` here should be a list of MTurk IDs, not a list of worker IDs as listed in the Mephisto database?

mephisto/scripts/mturk/soft_block_workers_by_mturk_id.py
```diff
@@ -0,0 +1,26 @@
...
+reqs = db.find_requesters(provider_type="mturk")
+names = [r.requester_name for r in reqs]
+print("Available Requesters: ", names)
+
+requester_name = input("Select a requester to soft block from: ")
+soft_block_qual_name = input("Provide a soft blocking qualification name: ")
+
+workers_to_block = []
+while True:
+    new_id = input("MTurk Worker Id to soft block (blank to continue): ")
```
EricMichaelSmith, line 21: nit: maybe clarify that blank will just soft-block all workers entered thus far?

mephisto/providers/mturk/utils/script_utils.py
JackUrb: You can register as many of them as you want (such that if you're running different projects with different budgets or accounts, you can swap just by providing different names). In the case that you have multiple accounts, but don't specify  here, it'll use the most recently registered account to assign the qualification. Later on, assigning that qualification through mephisto (with `worker.grant_qualification`) will automatically use the correct account to grant from though.

PR #195, opened 2020-07-07
Allow displaying images for ACUTE-Evals

EricMichaelSmith:
If a message has an "image_src" field (consisting of a serialized image string), show it in a chat bubble.

Also:
- Add the option to include additional task description text at the bottom of the left pane for ACUTE-Evals
- Fix in passing through "pairings_filepath" in `examples/acute_eval_demo/example_script.py`
- Small addition to .gitignore

### DIFF
--- a/examples/acute_eval_demo/example_script.py
+++ b/examples/acute_eval_demo/example_script.py
@@ -83,7 +83,7 @@ extra_args = {
     "annotations_per_pair": args["annotations_per_pair"],
     "random_seed": 42,  # random seed
     "subtasks_per_unit": args[
-        "subtask_per_unit"
+        "subtasks_per_unit"
     ],  # num comparisons to show within one unit
     "num_matchup_pairs": args[
         "num_matchup_pairs"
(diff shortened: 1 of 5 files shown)


mephisto/server/blueprints/acute_eval/webapp/src/components/core_components.jsx
```diff
@@ -79,18 +79,28 @@ function ChatMessage({ message, model, is_primary_speaker }) {
...
+    );
+  }
 }
 
 function MessageList({ task_data, index }) {
   let messageList;
 
   if (task_data.pairing_dict === undefined) {
-    messageLists = (
+    messageList = (
```
EricMichaelSmith, line 103: I assume this is supposed to be `messageList`? Because otherwise this variable is never used

examples/acute_eval_demo/example_script.py
```diff
@@ -76,7 +76,7 @@
 )
 
 extra_args = {
-    "pairings_filepath": f"{TASK_DIRECTORY}/pairings.jsonl",
+    "pairings_filepath": args['pairings_filepath'],
```
JackUrb, line 79: Nice catch!

PR #196, opened 2020-07-07
Single file ParlAI chat builds

JackUrb:
# Overview
You can now specify a full ParlAI chat frontend using just one `main.js` file. See the file `parlai_chat_task_demo/custom_simple/main.js`. A new flag `--custom-source-dir` tells mephisto to pick up the files in a custom directory, and use them to build the ParlAI chat task frontend. All build files are put into a subdirectory of the specified directory, which can be ignored when committing code. Even better, the source is only rebuilt when any of the user's custom files are altered.

# Implementation
Most of the updates are present in the `ParlAIChatTaskBuilder` in `build_and_return_custom_bundle`. The basic flow is that it checks to see if there is an existing custom build from the given dir, then to see if the existing build is up to date, and then only builds after that if there have been changes. At the moment it pulls from `TARGET_BUILD_FILES`, and these are the only files one can use to customize their task through this process.

The build is put into `{custom_source_dir}/pre_build/*`, which is a directory that Mephisto users should choose to put in their `.gitignore`'s if using Mephisto in their project.

# Testing 
Screen Shot 2020-07-07 at 7 46 37 PM

Made edits to the `main.js` file, asserted that that caused a rebuild, but otherwise no build was needed to relaunch.

# Discussion
The folder names really aren't ideal, but I just wanted to get it working. @pringshia we can talk about something better later. Same goes for the option names, we'll likely want to condense.

# Next steps
I want to make it possible to pull from a custom component directory as well, and have those imports work, but as this is an uncommon case, I'll leave it for future work.

cc @mwillwork - This should be a more optimal writing and building process for the majority of ParlAI users. (plus thanks to @pringshia  that `main.js` file is beautiful now 🤩)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@ mephisto/server/blueprints/**/build/*
 examples/**/build/*
 **/*.log
 **/build/*
-**/pre_build/*
+**/_generated/*
 
 # temporary files for inspiration from ParlAI-MTurk
 old_code/*
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,9 @@ tmp/*
 mephisto/server/**/package-lock.json
 mephisto/server/blueprints/**/build/*
 examples/**/build/*
+**/*.log
+**/build/*
+**/pre_build/*
 
 # temporary files for inspiration from ParlAI-MTurk
 old_code/*
--- a/examples/parlai_chat_task_demo/README.md
+++ b/examples/parlai_chat_task_demo/README.md
@@ -17,6 +17,12 @@ Further, you can try running a task using a customized frontend bundle with:
 python examples/parlai_chat_task_demo/parlai_test_script.py --use-custom-task true
 ```
 
+### Command line arguments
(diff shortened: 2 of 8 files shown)


pringshia:
This is great Jack! I tried running locally and it worked like a charm. This will be great in helping reduce the barrier-to-entry for folks. The build-if-changed portion is also very nice, I can re-run the task in seconds.

pringshia:
By the folder names in the discussion, which folder names are you referring to specifically? `pre_build` and `custom_simple`?

JackUrb:
After taking another look at this it's mostly just `pre_build`, as having a `pre_build/build` directory feels weird to me.

pringshia:
How about `_generated`? The underscore will let it drift to the top of folder structures and also hint to the end user that it is more of a "system" artifact. The "generated" word is also a hint that this is derived content and need not be checked into source control.

JackUrb:
I like the `_generated` idea a lot - committed and ready for full review/merge.

pringshia:
Looks great to me! Love this.

PR #205, opened 2020-07-22
Making taskConfig configurable

JackUrb:
# Overview
This PR makes it possible to easily configure the frontend arguments passed to `taskConfig` via the `extra_args` dict passed to the operator when launching a run. By setting a `task_config` key on in `extra_args`, anything in that dict will be available in `taskConfig` on the frontend.

# Implementation
Mostly just adds a pull from `task_config` to the core `Blueprint` class, but also requires a change to the `Operator` flow to ensure that the `Blueprint` _with_ the task args is already cached in the task run so that the architect can refer to it properly.

# Testing
Edited the static react task to see if I could pass the dict `{test: "test"}`:
Screen Shot 2020-07-22 at 2 10 27 PM
Screen Shot 2020-07-22 at 2 22 18 PM

cc @douwekiela

### DIFF
--- a/examples/static_react_task/README.md
+++ b/examples/static_react_task/README.md
@@ -39,6 +39,10 @@ which if you look at the `webapp/src/components/core_components.jsx` file you ca
 ```
 This is the flow for providing data to a static react task.
 
+#### Adding taskConfig
+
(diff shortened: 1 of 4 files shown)


pringshia:
Nice!

PR #208, opened 2020-07-23
Fixing integer value mturk qual specification

JackUrb:
# Overview
MTurk-specific qualifications have a few possible fields for values that need to be copied over, and `mturk_utils` was expecting to need to copy `LocaleValues` or `IntegerValue`, but not `IntegerValues`. This PR adds the missing key.

# Testing
Launched on sandbox with an `IntegerValues` qualification.

cc @douwekiela

### DIFF
--- a/mephisto/providers/mturk/mturk_utils.py
+++ b/mephisto/providers/mturk/mturk_utils.py
@@ -326,7 +326,11 @@ def convert_mephisto_qualifications(
         if converted["Comparator"] is None:
             converted["Comparator"] = qualification["comparator"]
 
-        if converted["IntegerValue"] is None and converted["LocaleValues"] is None:
+        if (
+            converted["IntegerValue"] is None
+            and converted["IntegerValues"] is None
+            and converted["LocaleValues"] is None
+        ):
             value = qualification["value"]
             if isinstance(value, list):
                 converted["IntegerValues"] = value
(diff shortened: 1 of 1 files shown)


mephisto/providers/mturk/mturk_utils.py
```diff
@@ -326,14 +327,21 @@ def convert_mephisto_qualifications(
...
-        if converted["IntegerValue"] is None and converted["LocaleValues"] is None:
+        if (
+            converted["IntegerValue"] is None
+            and converted["IntegerValues"] is None
+            and converted["LocaleValues"] is None
+        ):
             value = qualification["value"]
             if isinstance(value, list):
                 converted["IntegerValues"] = value
+                del converted["IntegerValue"]
```
pringshia, line 338: should this be the plural key being deleted...

mephisto/providers/mturk/mturk_utils.py
```diff
@@ -326,14 +327,21 @@ def convert_mephisto_qualifications(
...
             elif isinstance(value, int):
                 converted["IntegerValue"] = value
+                del converted["IntegerValues"]
```
pringshia, line 341: and this be the singular key being deleted?

mephisto/providers/mturk/mturk_utils.py
JackUrb: The idea here is only one of `IntegerValues`, `IntegerValue`, or `LocaleValues` should be set. As such, if it's a list, we delete the singular, and if it's the singular we delete the list.

PR #209, opened 2020-07-23
Various Acute eval state fixes

JackUrb:
# Overview
A few Acute eval bugs have come up lately:
- Sometimes the task build is asking for permissions that a user may not have
- Not saving the task duration
- Sometimes the last evaluation is not saved
- Some errant scripts seemed to have been able to edit valid `AgentStates's while in-flight (potentially breaking them during the save).

This PR fixes those

# Implementation
- `subprocess.call(["npm", "link", "mephisto-task"])` was only required before `mephisto-task` was published, and isn't necessary anymore. This was responsible for the permission issue.
-  Added the step of saving task duration
- Implemented a frontend change that should prevent the clearest possible cause of saving `{}` as the last value.
- To prevent possible `AgentState` corruption, includes a change that only returns a copy of the internal state when requested via accessor functions.

# Testing
Ran Acute eval locally, didn't see any of the above issues.

### DIFF
--- a/mephisto/server/blueprints/acute_eval/webapp/src/components/core_components.jsx
+++ b/mephisto/server/blueprints/acute_eval/webapp/src/components/core_components.jsx
@@ -232,15 +232,15 @@ class EvalResponse extends React.Component {
   }
 
   checkValidData() {
+    let response_data = {
+      speakerChoice: this.state.speakerChoice,
+      textReason: this.state.textReason,
+    };
     if (this.state.speakerChoice !== "" && this.state.textReason.length > 4) {
-      let response_data = {
-        speakerChoice: this.state.speakerChoice,
-        textReason: this.state.textReason,
-      };
       this.props.onValidDataChange(true, response_data);
       return;
     }
-    this.props.onValidDataChange(false, {});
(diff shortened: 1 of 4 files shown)


EricMichaelSmith:
Thanks @JackUrb ! I'm happy to rubber-stamp but others might know more about this than I do

pringshia:
LGTM

mephisto/server/blueprints/acute_eval/webapp/src/components/core_components.jsx
```diff
@@ -232,15 +232,15 @@ class EvalResponse extends React.Component {
...
+      speakerChoice: this.state.speakerChoice,
+      textReason: this.state.textReason,
+    };
     if (this.state.speakerChoice !== "" && this.state.textReason.length > 4) {
-      let response_data = {
-        speakerChoice: this.state.speakerChoice,
-        textReason: this.state.textReason,
-      };
       this.props.onValidDataChange(true, response_data);
       return;
```
pringshia, line 241: tiniest of nits - i'd probably just get rid of this `return` and wrap the `onValidDataChange` call below in an `else` block

PR #222, opened 2020-08-04
operator shut down is aware of finished assignments thread

saraEbrahim:
Regarding the 3rd point of this thread, this PR is for the `operator.shut_down()` to shut down and update the database of the completed run only if the assignments generator thread is done (if it exists)

### DIFF
--- a/mephisto/core/operator.py
+++ b/mephisto/core/operator.py
@@ -290,12 +290,10 @@ class Operator:
             runs_to_check = list(self._task_runs_tracked.values())
             for tracked_run in runs_to_check:
                 task_run = tracked_run.task_run
-                if not tracked_run.task_launcher.assignment_thread_done:
+                task_run.update_completion_progress(tracked_run.task_launcher)
+                if not task_run.get_is_completed():
                     continue
                 else:
-                    task_run.assignment_generators_done = True
-
-                if task_run.get_is_completed():
                     self.supervisor.shutdown_job(tracked_run.job)
                     tracked_run.architect.shutdown()
                     tracked_run.task_launcher.shutdown()
--- a/mephisto/core/operator.py
+++ b/mephisto/core/operator.py
@@ -279,7 +279,7 @@ class Operator:
             architect=architect,
             job=job,
         )
-        task_run.update_assignment_generator_done(False)
+        task_run.update_completion_progress(status=False)
 
         return task_run.db_id
 
@@ -292,7 +292,7 @@ class Operator:
             runs_to_check = list(self._task_runs_tracked.values())
             for tracked_run in runs_to_check:
                 task_run = tracked_run.task_run
-                task_run.update_completion_progress(tracked_run.task_launcher)
+                task_run.update_completion_progress(task_launcher=tracked_run.task_launcher)
                 if not task_run.get_is_completed():
                     continue
                 else:
--- a/mephisto/core/operator.py
+++ b/mephisto/core/operator.py
@@ -279,6 +279,7 @@ class Operator:
             architect=architect,
             job=job,
         )
+        task_run.assignments_generator_done = False
         return task_run.db_id
 
     def _track_and_kill_runs(self):
--- a/mephisto/core/operator.py
+++ b/mephisto/core/operator.py
@@ -279,7 +279,8 @@ class Operator:
             architect=architect,
             job=job,
         )
-        task_run.assignments_generator_done = False
+        task_run.update_assignment_generator_done(False)
+
         return task_run.db_id
 
     def _track_and_kill_runs(self):
--- a/mephisto/core/operator.py
+++ b/mephisto/core/operator.py
@@ -290,6 +290,11 @@ class Operator:
             runs_to_check = list(self._task_runs_tracked.values())
             for tracked_run in runs_to_check:
                 task_run = tracked_run.task_run
+                if not tracked_run.task_launcher.assignment_thread_done:
+                    continue
+                else:
+                    task_run.assignment_generators_done = True
+
                 if task_run.get_is_completed():
                     self.supervisor.shutdown_job(tracked_run.job)
                     tracked_run.architect.shutdown()
--- a/mephisto/core/task_launcher.py
+++ b/mephisto/core/task_launcher.py
@@ -73,7 +73,7 @@ class TaskLauncher:
         self.unlaunched_units_access_condition = threading.Condition()
         if isinstance(self.assignment_data_iterable, types.GeneratorType):
             self.generator_type = GeneratorType.ASSIGNMENT
-            self.assignments_thread = False
+            self.assignment_thread_done = False
         else:
             self.generator_type = GeneratorType.NONE
         run_dir = task_run.get_run_dir()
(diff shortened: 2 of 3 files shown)


mephisto/core/operator.py
```diff
@@ -290,6 +290,11 @@ def _track_and_kill_runs(self):
             runs_to_check = list(self._task_runs_tracked.values())
             for tracked_run in runs_to_check:
                 task_run = tracked_run.task_run
+                if not tracked_run.task_launcher.assignment_thread_done:
+                    continue
+                else:
+                    task_run.assignment_generators_done = True
```
JackUrb, line 296: It's somewhat strange that the operator is directly checking and setting attributes for the task run. The semantics here would be better in calling `task_run.update_completion_progress(task_launcher)` or something similar, and then let the `get_is_completed()` call still be responsible for the implicit continue in this loop. The `update_completion_progress` method should ideally only rely on `task_launcher`'s functions, rather than direct accesses to its attributes.

mephisto/core/task_launcher.py
```diff
@@ -68,9 +68,12 @@ def __init__(
         self.unlaunched_units: Dict[str, Unit] = {}
         self.keep_launching_units: bool = False
         self.finished_generators: bool = False
+        self.assignment_thread_done: bool = True
+
         self.unlaunched_units_access_condition = threading.Condition()
         if isinstance(self.assignment_data_iterable, types.GeneratorType):
             self.generator_type = GeneratorType.ASSIGNMENT
+            self.assignments_thread = False
```
JackUrb, line 76: What is this added for?

saraEbrahim: it is a typo, it should be `self.assignment_thread_done` flag.

saraEbrahim:
Hey @JackUrb 
I didn't understand this comment:

> In this case, one may re-initialize a task_run from the database after it's already been completed, in which case the `assignment_generators_done` flag may not properly reflect the state things are in. Also, `sync_completion_status` will only ever be able to trigger in the thread where things are run from, and as such it's important to note this now.

Can you explain it a bit more?

JackUrb:
Of course - the `TaskRun` object can be initialized outside of a run, say while someone is trying to run analysis on a completed run or something of the sort. As such a new `TaskRun` will be initialized that is entirely out of the context of any operator, and it wont ever have the `assignment_generators_done ` flag set. 

If you were to initialize that `TaskRun` in a process that wasn't the same process running the `Operator`, you'll end up in a scenario where you have `task_run_ex` (the examined `TaskRun`), and `task_run_op` (the `TaskRun` that the `Operator` retains a reference to). Both these objects refer to the same "Task Run", but only `task_run_op.sync_completion_status()` will ever actually be able to change the completion status as `task_run_op` is the only one that will have `assignment_generators_done` change. This isn't necessarily _wrong_ behavior, but it should be noted in `sync_completion_status`'s description that only `TaskRun`s that are currently run by an `operator` have the context to update the completion status.

saraEbrahim:
@JackUrb so should I just note this in `sync_completion_status`' description or should I try avoid that behaviour?

mephisto/core/task_launcher.py
```diff
@@ -188,6 +192,9 @@ def launch_units(self, url: str) -> None:
         )
         self.units_thread.start()
 
+    def get_assignment_thread_flag(self) -> bool:
```
JackUrb, line 195: Nit: ideally class methods should be named closer to the intent that they are trying to match, because down the road it's not necessarily clear what the `assignment_thread_flag` is supposed to be representing. `get_assignments_are_all_created` for instance is more in-line with the intent for this flag.

mephisto/core/operator.py
```diff
@@ -279,6 +279,7 @@ def parse_and_launch_run(
             architect=architect,
             job=job,
         )
+        task_run.assignments_generator_done = False
```
JackUrb, line 282: Nit - this should also use the setter `update_completion_progress` rather than a direct access, as we are trying to decouple the two and the less the operator needs to touch directly, the better.

mephisto/data_model/task.py
```diff
@@ -199,6 +199,7 @@ def __init__(
         self.provider_type = row["provider_type"]
         self.task_type = row["task_type"]
         self.sandbox = row["sandbox"]
+        self.assignments_generator_done: bool = None
```
JackUrb, line 202: 👍

mephisto/data_model/task.py
```diff
@@ -380,7 +386,12 @@ def sync_completion_status(self) -> None:
             for status in AssignmentState.incomplete():
                 if statuses[status] > 0:
                     has_incomplete = True
-            if not has_incomplete:
+            if (
+                not has_incomplete
+                and self.assignments_generator_done is None
+                or not has_incomplete
+                and self.assignments_generator_done
```
JackUrb, line 393: This conditional can be made simpler, what are you trying to capture in this logic?

mephisto/data_model/task.py
```diff
@@ -362,6 +363,11 @@ def get_assignment_statuses(self) -> Dict[str, int]:
             for status in AssignmentState.valid()
         }
 
+    def update_completion_progress(self, task_launcher) -> None:
```
JackUrb, line 366: Nit: should be used to both initialize `assignments_generator_done` to `False` as well as to set it `True` later. You can use the `task_launcher` to determine the state in both cases.

mephisto/data_model/task.py
saraEbrahim: I'm trying to let both `TaskRun` (the one that's run by the `Operator` and the one that is independent) to access this line.

JackUrb: I can't follow this logic very clearly though as written considering it has a number of `and`s and `or`s interlinked, with two of the same `not has_incomplete` call. Going to take steps to simplify.
```python
if not has_incomplete and self.assignments_generator_done is None or not has_incomplete and self.assignments_generator_done: # The logic in here is very difficult to parse given the many statements and no divisions
if (not has_incomplete and self.assignments_generator_done is None) or (not has_incomplete and self.assignments_generator_done): # My guess is that this is your intended behavior, as python gives precedence to and over or, but this isn't easily seen without additional division.
if (not has_incomplete) and (self.assignments_generator_done is None or self.assignments_generator_done): # Given the above, it's possible to pull out the not has_incomplete term
if (not has_incomplete) and (self.assignments_generator_done is not False): # The latter term of the above can also be simplified, as we know assignments_generator_done only has 3 possible states.
if not has_incomplete and self.assignments_generator_done is not False:
```

mephisto/data_model/task.py
```diff
@@ -363,9 +363,12 @@ def get_assignment_statuses(self) -> Dict[str, int]:
...
         }
 
+    def update_assignment_generator_done(self, status) -> None:
+        self.assignments_generator_done = False
+
     def update_completion_progress(self, task_launcher) -> None:
         """ Flag the task run that the assignments' generator has finished """
-        if task_launcher.get_assignment_thread_flag():
+        if task_launcher.get_assignments_are_all_created():
             self.assignments_generator_done = True
```
JackUrb: Why are these separate functions? You can define a python function that takes in keyword arguments (like `status` and `task_launcher` but defaults these to `None`) and chooses behavior based on which arguments are provided. Alternately, you can make `update_completion_progress` rely on `update_assignment_generator_done`.

Further, `update_assignment_generator_done`'s function signature at the moment doesn't actually represent what you're doing with it.

saraEbrahim: fixed it now. Thanks!

PR #223, opened 2020-08-05
sqlite connection as context manager

bottler:
For each sqlite3 database, we have one connection per thread. There are a few places where we had transactions which don't call `rollback()` on every error. If an error escapes them, the database will still be locked to that thread, breaking every other thread's ability to access the database. I now use the connection itself as a context manager around every transaction, so this can't happen.

I've simplified grant_qualification using an extension to sql syntax which needs sqlite version 3.24.0 (2018-06-04). This is old enough to be included in any build of python 3.8.


JackUrb:
Thanks for these fixes @bottler!

PR #226, opened 2020-08-12
Revert to allow local_database to work on 3.6

JackUrb:
The changes in #223, while nice, do make this incompatible with 3.6 and 3.7, which most of our users are on at the moment. I've left a note on the possible upgrade, but reverted the usage for now.


jxmsML:
thanks for the quick fix!

PR #237, opened 2020-09-01
front_end_error_logging

saraEbrahim:
Hello all,

I'm running into an issue, here is the detailed explanation:

For the front end error message:
- I've added a `ErrorBoundary` component inside `static_react_task/webapp/app.jsx` temporarily (once I correct the behavior of the issue I have I'll move it to `mephisto-task/utils`
- I was trying to render a nice looking message box or popup once an error occurs.
- In the test case Pratik suggested, the error does show up in the console but the `componentDidCatch` inside `ErrorBoundary` component is not called, I've tried to `console.log()` some messages inside that method with nothing showing up in the console. Am I missing something or it's real that `ErrorBoundary` could be not triggered in that test case as stated [here](https://reactjs.org/docs/error-boundaries.html#introducing-error-boundaries) and I just need to try a different test case?

For propagating the error from the front to the backend:
- I added `handleFatalError` same as `handleSubmit` in `useMephistoTask` with just calling `handleSubmit` with a different data dictionary.
- Then I added a property to the `ErrorBoundary` called `handleError` which is triggered in `componentDidCatch` (which is not called properly as I explained in the first part)
- was what I added functioning but not running due to the issue of `componentDidCatch` not being triggered or do I miss something important here?


JackUrb:
One issue is that react doesn't appear to be able to catch errors that happen in handlers, only in lifecycle functions: https://github.com/facebook/react/issues/11409.

These kinds of errors will end up being type 2 and 3 instead.

For testing this, you should throw your error during the `render` method.

JackUrb (commented):
Error logging flow in general looks good. Nice job following the existing conventions for constants and post methods! Next up is going to be polishing and dealing with the global case in a way that allows users to decide if something major broke. 

It might be worth some kind of heuristic on global errors, like setting a timeout when they occur, and then seeing if the user is still interacting with the app and making progress (which would imply that the error was actually minor). Any thoughts on what a good practice would be for this @pringshia?

examples/static_react_task/webapp/src/app.jsx
```diff
@@ -11,19 +11,76 @@ import ReactDOM from "react-dom";
 import { BaseFrontend, LoadingScreen } from "./components/core_components.jsx";
 import { useMephistoTask } from "mephisto-task";
 
+class ErrorBoundary extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { error: null, errorInfo: null };
+  }
```
pringshia, line 18: I believe you can just put `state = { error: null, errorInfo: null }` outside of the constructor. this should let you avoid having to implement the constructor altogether and make this code a little more condensed. see more [here](https://www.robinwieruch.de/react-state-without-constructor) and [here](https://github.com/the-road-to-learn-react/react-alternative-class-component-syntax)

packages/mephisto-task/src/utils.js
```diff
@@ -132,6 +132,16 @@ export function postCompleteTask(agent_id, complete_data) {
...
 }
 
+export function postErrorLog(agent_id, complete_data) {
+  return postData("/log_error", {
+    USED_AGENT_ID: agent_id,
+    final_data: complete_data,
+  })
+    .then(function (data) {
+      console.log("Error log sent to server");
+    });
```
pringshia, line 142: we can remove these debugging console logs once you're ready to merge

packages/mephisto-task/src/index.js
```diff
@@ -68,6 +69,14 @@ const useMephistoTask = function () {
     [state.agentId]
   );
 
+   const handleFatalError = React.useCallback(
+   (data) => {
+       console.log('inside handleFatalError ...')
+       postErrorLog(state.agentId, data)
+//       handleSubmit(data)
```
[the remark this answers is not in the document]
saraEbrahim: How can I categorize the fatal errors from non fatal ones?

examples/static_react_task/webapp/src/app.jsx
```diff
@@ -11,19 +11,76 @@ import ReactDOM from "react-dom";
...
+
+  componentDidCatch(error, errorInfo) {
+    // Catch errors in any components below and re-render with error message
+      this.setState({
+      error: error,
+      errorInfo: errorInfo
+    })
+    // alert Mephisto worker of a component error
+    alert("error from a component" + error)
```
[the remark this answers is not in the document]
saraEbrahim: Yes. It was an indication for me that the error was raised from `ErrorBoundary`

mephisto/server/architects/router/deploy/server.js
```diff
@@ -476,6 +479,20 @@ app.post('/submit_task', upload.any(), function(req, res) {
   }
 });
 
+app.post('/log_error', function(req, res) {
+  var provider_data = req.body;
+  let agent_id = provider_data.USED_AGENT_ID;
+  delete provider_data.USED_AGENT_ID;
```
pringshia, line 485: tip: one way you can condense the functionality of the above lines is to do the following:

`const { USED_AGENT_ID: agent_id, ...provider_data} = req.body`

this will end up doing the same thing as the above but in a single line. what it does is destructures `req.body` storing `USED_AGENT_ID` in `agent_id` and then capturing the remaining properties in `provider_data` using the spread operator.

examples/static_react_task/webapp/src/components/core_components.jsx
```diff
@@ -62,7 +66,9 @@ function SimpleFrontend({ taskData, isOnboarding, onSubmit }) {
             <div className="control">
               <button
                 className="button is-success is-large"
-                onClick={() => onSubmit({ rating: "good" })}
+                // Test for type 2 errors
+                onClick={()=> {throw new Error("test error event_handler");}}
+//                 onClick={() => onSubmit({ rating: "good" })}
```
pringshia, line 71: just a reminder to revert this when we're ready to merge

pringshia: to echo Jack, these errors in the callback may actually be caught as type 3 errors, instead of 2

examples/static_react_task/webapp/src/app.jsx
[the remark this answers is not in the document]
saraEbrahim: I actually don't see a difference between type 2 and type 3 (global) errors. I had to place the `window.eventListener('error'..)` definition after the `useMephistoTask` as it uses `handleFatalError`. And even placing an error at the beginning of `MainApp` wouldn't fire the `window.eventListener('error', ..)` nor `window.onerror` callback as the error is place before defining any of them. So I don't know how to produce a global error that might happen right after defining the `MainApp` unless `useMephistoTask` should handle that type inside it, is that right?

examples/static_react_task/webapp/src/app.jsx
```diff
@@ -11,19 +11,76 @@ import ReactDOM from "react-dom";
...
+  }
+  event.error.hasBeenCaught = true
+  if (confirm("Do you want to report the error?")) {
+          console.log("You pressed OK!");
+          handleFatalError({errorMsg: event.error.message, error: event.error.stack})
+        } else {
+          console.log("You pressed Cancel!");
+        }
+        return true;
+})
```
pringshia, line 80: once this seems to be working correctly, let's move it into `wrap_crowd_source.js` so all tasks automatically inherit this functionality

examples/static_react_task/webapp/src/app.jsx
```diff
@@ -11,19 +11,76 @@ import ReactDOM from "react-dom";
 import { BaseFrontend, LoadingScreen } from "./components/core_components.jsx";
 import { useMephistoTask } from "mephisto-task";
 
+class ErrorBoundary extends React.Component {
```
pringshia, line 14: great first step in implementing this. the next step would be to move this component into `mephisto-task` and import it into this project from there. this will allow us to have a reusable mephisto component we can use.

pringshia:
As Jack mentioned `componentDidCatch` is not being triggered because the error is being thrown from an onClick event handler, which gets run in a different execution context. What you'd want to do instead if throw an error from within the render method (that is not triggered via an event handler).

packages/mephisto-task/src/index.js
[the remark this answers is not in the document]
pringshia: as a temporary step you can trigger a `prompt()` and allow the user to choose? i don't like this because it can be a jarring experience for users, especially in the case of false positives, but it can be a good stopgap for now

examples/static_react_task/webapp/src/app.jsx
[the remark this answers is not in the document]
pringshia: Hmm, I wonder why this is happening, as `<script src="wrap_crowd_source.js"></script>` is defined in the `<head` tag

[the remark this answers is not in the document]
pringshia: seems like I was able to get it to work:

## console

Screen Shot 2020-09-23 at 12 53 21 PM

## examples/parlai_chat_task_demo/custom_simple/main.js

Screen Shot 2020-09-23 at 12 59 54 PM

## wrap_crowd_source.js

Screen Shot 2020-09-23 at 1 00 03 PM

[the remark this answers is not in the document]
pringshia: Screen Shot 2020-09-23 at 1 04 38 PM

packages/mephisto-task/src/index.js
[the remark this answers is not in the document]
saraEbrahim: I already did this for type 2 errors. Thansk

examples/static_react_task/webapp/src/app.jsx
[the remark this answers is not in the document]
saraEbrahim: Yes I am able to see them in the console but do you get them back in the backend? there should be a logging INFO for the error sent back to the backend

packages/mephisto-task/src/index.js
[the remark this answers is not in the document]
saraEbrahim: Do you mean do the `confirm()` inside `componentDidCatch()`? because I already do `confirm()` in the `window.onerror` event.

saraEbrahim:
I've polished the debugging logs and script, and re-located the error handling components (`window.onerror` event and `ErrorBoundary` react component) to be inside `useMephistoTask` and `utils.js` respectively. 

I tried to add `window.onerror` to `wrap_crowd_source` as you said @pringshia but I need `handleFatalError` from `useMephistoTask` and also I didn't understand how this can be used for other providers since `wrap_crowd_source` is per provider. Did you mean to make it specific for every provider?

examples/static_react_task/webapp/src/components/core_components.jsx
saraEbrahim: the event handlers errors which can't be caught by React's `ErrorBoundary` are caught by the `window.onerror` event handler and those which can be caught by both the `window.onerror` event handler and `ErrorBoundary` are caught twice but this should be the case for `mode=development` and once the `mode=production` it should be caught only once in the `ErrorBoundary` [according to StackOverFlow]

pringshia:
Are you still planning on adding a const to the top of wrap_crowd_source.js to flag between the two different behaviors we discussed? (i.e. auto-submit on error vs. prompt to submit on error).

mephisto/providers/mock/wrap_crowd_source.js
```diff
@@ -15,6 +15,7 @@ for both to be able to register them with the backend database.
 Returning None for the assignment_id means that the task is being
 previewed by the given worker.
 \------------------------------------------*/
+auto_submit = false
```
pringshia, line 18: looks good, I would make this ALL_CAPS as such and declare it as a `const`

mephisto/providers/mock/wrap_crowd_source.js
```diff
@@ -53,3 +54,22 @@ function handleSubmitToProvider(task_data) {
...
+  }
+  event.error.hasBeenCaught = true
+  if (!auto_submit) {
+      if (confirm("Do you want to report the error?")) {
+          prompt('send the following error to the email address: '+
+          '[email address]', JSON.stringify(event.error.message))
+            }
+  }
+  else {
+    console.log("sending to email address: ####")
```
pringshia, line 72: we can wire up this portion to the `/log_error` endpoint you created?

mephisto/providers/mock/wrap_crowd_source.js
```diff
@@ -53,3 +54,22 @@ function handleSubmitToProvider(task_data) {
...
+window.addEventListener('error', function (event) {
+
+    if (event.error.hasBeenCaught !== undefined){
+    return false
+  }
+  event.error.hasBeenCaught = true
+  if (!auto_submit) {
+      if (confirm("Do you want to report the error?")) {
+          prompt('send the following error to the email address: '+
+          '[email address]', JSON.stringify(event.error.message))
```
JackUrb, line 68: Sorry a bit out of context here - why are we asking for an email address?

saraEbrahim: Pratik told me there should be a choice whether for the user to auto submit the error (wiring up `/log_error`) or show them an email and a copy-enabled version of the error message to send that error by themselves to an email address.

JackUrb: Understood - generally we may have a case where we want the submit to be optional but not explicitly state the contact info (for a case where people may not want to expose their actual email, but instead get contacted through the site). So really there should be three options:
1. Auto-submit. Display the message "An error has occurred. This has been logged to the requester automatically."
2. Don't auto-submit. Display the message. "We think an error has occurred. If the task is broken, please send this context to the requester.
3. Don't auto-submit. Same message as the above, but with an email instead of "to the requester"

mephisto/providers/mock/wrap_crowd_source.js
```diff
@@ -53,3 +54,22 @@ function handleSubmitToProvider(task_data) {
     alert("The task has been submitted! Data: " + JSON.stringify(task_data))
     return true;
 }
+
+// Adding event listener instead of using window.onerror prevents the error to be caught twice
+window.addEventListener('error', function (event) {
```
JackUrb, line 59: @pringshia is there a reason this listener is defined in `wrap_crowd_source.js` and not as a part of `mephisto-task`?

saraEbrahim: the first place I put it inside `mephisto-task` but Pratik said it's better to be inside each provider separately but I didn't find it convenience because we will need to put it in each Provider's `wrap_crowd_source.js`, @pringshia  could you elaborate why is this better?

pringshia: @JackUrb - The reason for putting this in `wrap_crowd_source.js` as opposed to `mephisto-task` is so that this global catch-all error handling can also apply to:

- Errors in static HTML tasks
- Errors that may occur at the React library level, before the application code has had a chance to run (maybe related due to a build linking / webpack error?)

JackUrb:
Changes in here are all included in #343, so I'm closing this version.

PR #248, opened 2020-09-16
Added yaml to the requirements file.

mojtaba-komeili:
While installing Mephisto and running `mephisto check`, I noticed that *yaml* was not installed. I added it to the requirements file.


JackUrb:
Thanks for the fix!

PR #251, opened 2020-09-17
Parlai live task disconnects

JackUrb:
# Overview
Sometimes disconnect events weren't picked up from the router, especially if the only incoming messages so far had been heartbeats. Adding an outgoing ping from the server every once in a while forces the disconnect to occur when the system tries to use the socket.


saraEbrahim (approved):

saraEbrahim:
Thanks so much for the fix!

Issue #256, opened 2020-09-21
bug when run parlai_chat_task_demo on mturk sandbox


WilliamsToTo:
We use the latest mephisto.
This is the commend we use `python parlai_test_script.py conf=onboarding_example mephisto/architect=heroku mephisto/provider=mturk_sandbox`.  We use this commend because the commend in [quickstart](https://github.com/facebookresearch/mephisto/blob/master/docs/quickstart.md) reports error.

Then we get the following error:

>[09-21 16:32:15] p26949 {operator.py:254} ERROR - Encountered error while launching run, shutting down
Traceback (most recent call last):
  File "/home/tao/Mephisto/mephisto/core/operator.py", line 242, in validate_and_run_config_or_die
    provider.setup_resources_for_task_run(task_run, run_config, task_url)
  File "/home/tao/Mephisto/mephisto/providers/mturk/mturk_provider.py", line 107, in setup_resources_for_task_run
    for qualification in task_args.get("qualifications", []):
NameError: name 'task_args' is not defined
[2020-09-21 16:32:15,742][mephisto.core.operator][ERROR] - Encountered error while launching run, shutting down
Traceback (most recent call last):
  File "/home/tao/Mephisto/mephisto/core/operator.py", line 242, in validate_and_run_config_or_die
    provider.setup_resources_for_task_run(task_run, run_config, task_url)
  File "/home/tao/Mephisto/mephisto/providers/mturk/mturk_provider.py", line 107, in setup_resources_for_task_run
    for qualification in task_args.get("qualifications", []):
NameError: name 'task_args' is not defined
Heroku: Deleting server: tao-parlai-chat-example-6-cc11
Destroying tao-parlai-chat-example-6-cc11 (including all add-ons)... done
[09-21 16:32:23] p26949 {operator.py:352} ERROR - Ran into error while launching run: 
Traceback (most recent call last):
  File "/home/tao/Mephisto/mephisto/core/operator.py", line 349, in validate_and_run_config
    run_config=run_config, shared_state=shared_state
  File "/home/tao/Mephisto/mephisto/core/operator.py", line 263, in validate_and_run_config_or_die
    raise e
  File "/home/tao/Mephisto/mephisto/core/operator.py", line 242, in validate_and_run_config_or_die
    provider.setup_resources_for_task_run(task_run, run_config, task_url)
  File "/home/tao/Mephisto/mephisto/providers/mturk/mturk_provider.py", line 107, in setup_resources_for_task_run
[...]

PR #250, opened 2020-09-17
Some additional hydra suggestions

JackUrb:
# Overview
Just a few general fixes following Omry's advice in #246 
- Renaming the path for structured configs from `mephisto.<abstraction>` to `mephisto/<abstraction>` to fit norms
- fixing multiline yaml by using `>`
- Updating all mephisto configs to directly use `mephisto.provider.requester_name` instead of instantiating a requester config for just that.
- Some additional string fixes that were breaking parsing (oops).

PR #257, opened 2020-09-21
Hydra MTurk fixes, plus doc update

JackUrb:
# Overview
A small bug was flagged in #256 that was breaking deploys to MTurk sandbox (as well as an unexpected change following applying Omry's advice in #250). This updates the Doc describing how to work things in the quickstart, and fixes the underlying bug in the mturk provider.

# Testing
Screen Shot 2020-09-21 at 9 30 59 AM
`pytest` runs tests fine.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -12,7 +12,7 @@ $ pip install -e .
 
 # Verify that mephisto is installed correctly, and handle any required config:
 $ mephisto check
-Please enter the full path to a location to store Mephisto run data. By default this would be at '/private/home/jju/mephisto/data'. This dir should NOT be on a distributed file store. Press enter to use the default:  
+Please enter the full path to a location to store Mephisto run data. By default this would be at '/private/home/jju/mephisto/data'. This dir should NOT be on a distributed file store. Press enter to use the default:
 Mephisto seems to be set up correctly.
 ```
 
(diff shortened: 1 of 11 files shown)


docs/quickstart.md
```diff
@@ -83,8 +83,9 @@ $ python static_test_script.py mephisto.architect=heroku mephisto.provider.reque
 # Note: my_mturk_user_sandbox is what we used to name the requester
 # when we registered the mturk account in the previous step.
 ```
-The arguments `mephisto.provider.requester_name=my_mturk_user_sandbox` and `mephisto.architect=heroku` will tell the the script to use the mturk sandbox provider and the heroku architect (as opposed to the mock provider and local architect).
+The arguments `mephisto.provider.requester_name=my_mturk_user_sandbox` and `mephisto/architect=heroku` will tell the the script to use the mturk sandbox provider and the heroku architect (as opposed to the mock provider and local architect). Notice that if we're setting a full abstraction (like the architect) we reference it with `mephisto/abstraction=val`, however when we're setting an argument, we use `mephisto.abstraction.argument=val`. This tells hydra whether we're providing an argument value or the name of a configuration to load.
```
EricMichaelSmith, line 86: Hmm - line 81 should now be changed to say `mephisto/architect=heroku`, right?

EricMichaelSmith (approved):
Seems sensible! Stamping, but others might have other concerns

docs/quickstart.md
EricMichaelSmith: Also, there's still a reference to `mephisto.architect=heroku` in https://github.com/facebookresearch/Mephisto/blob/master/docs/hydra_migration.md

JackUrb: Thanks for the flags! Updated all of them now

PR #258, opened 2020-09-21
Making it possible to further control task pairing

JackUrb:
# Overview
ParlAI MTurk used to have the ability to individually check to see if a worker was eligible to work on a particular task. This was batched into a number of options collectively called `eligibility_function`. 

This PR intends to restore that functionality to Mephisto, using the `SharedTaskState` to register a function called `worker_can_do_unit`.

# Implementation
- Adds `worker_can_do_unit` to `SharedTaskState`
- `get_valid_units_for_worker` now calls `worker_can_do_unit` to filter out the units before returning a list.
- Default value for `worker_can_do_unit` always returns `True`

# Testing
Replacing the default return value with `False` makes it impossible to launch any tasks.
As is, it's possible to run the demo tasks to completion.

### DIFF
--- a/mephisto/data_model/blueprint.py
+++ b/mephisto/data_model/blueprint.py
@@ -17,6 +17,7 @@ from typing import (
     Union,
     Iterable,
     AsyncIterator,
+    Callable,
     TYPE_CHECKING,
 )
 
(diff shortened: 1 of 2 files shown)


pringshia (approved):
nice and clean, good naming 👍

PR #261, opened 2020-09-23
Add prettier precommit hook

pringshia:
Title

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,3 +12,7 @@ repos:
     hooks:
       - id: black
         language_version: python3.6
+  - repo: https://github.com/prettier/prettier
+    rev: stable
+    hooks:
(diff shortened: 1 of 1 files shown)


examples/parlai_chat_task_demo/conf/custom_simple.yaml
```diff
@@ -6,8 +6,8 @@ mephisto:
     task_name: parlai-chat-example
     task_title: "Test ParlAI Simply Built Chat Task"
     task_description: >
-      This is a simple chat between two people 
-      used to demonstrate the functionalities around using Mephisto
```
pringshia, line 10: This was a good bug that was caught. The trailing whitespace here was causing double spaces to be added to the string.

JackUrb (approved):
Mephisto is getting prettier every day.

PR #262, opened 2020-09-23
Remove old deps

JackUrb:
# Overview
Generally the title. We had some old deps that people were copy-pasting that didn't need to be there, so I've removed them. Tested by launching each of these tasks and ensuring they still loaded and interacted.


pringshia (approved):
Perfect, thanks!

PR #265, opened 2020-09-24
Corrected arguments for MTurk Requester

vaibhavad:
Correcting [arguments](https://github.com/facebookresearch/Mephisto/blob/c7b3e3988e7a0d31db543df93c5e68f660752367/mephisto/providers/mturk/mturk_requester.py#L35) for MTurk Requester


JackUrb (approved):
Good catch, thanks!

PR #266, opened 2020-09-24
Update quickstart.md

vaibhavad:
Registering for mturk_sandbox requester was a bit unclear, so I added an explicit command. Also added the starting logs of running static_test_script.py, and updated description of examining results.


JackUrb (approved):
Thanks for taking the time to update these docs to make them easier to use for future users!

PR #270, opened 2020-09-29
Introduce `mephisto review` CLI utility

pringshia:
How it works:

### 1. Create a sample `data.csv`:
```
This is good text, row1
This is bad text, row2
```
### 2. Use `create-react-app` to quickly create a Review UI
```npx create-react-app review-ui```

Replace `App.js` with the following:

```jsx
import React, { useState, useEffect } from 'react';
import './App.css';

function App() {

  const [data, setData] = useState(null)
  const [counter, setCounter] = useState(0)

  useEffect(() => {
    fetch("/data")
      .then(res => res.json())
      .then((data) => setData(data));
  }, [counter])

  const handleApprove = React.useCallback(() => {
    fetch("/next", { method: "POST", body: JSON.stringify({ "result": "approved" }) })
      .then(() => setCounter(counter + 1))
  }, [counter, setCounter])
  const handleReject = React.useCallback(() => {
    fetch("/next", { method: "POST", body: JSON.stringify({ "result": "rejected" }) })
      .then(() => setCounter(counter + 1))
  }, [counter, setCounter])

  return (
    <div className="App">
      {!data
        ? <h1>Loading...</h1>
        : data.finished === true
          ? <h1>Done reviewing!</h1>
          : <header className="App-header">
            <pre>
              {JSON.stringify(data.data, 2)}
            </pre>
            <button onClick={handleApprove}>Approve</button>
            <button onClick={handleReject}>Reject</button>
            </header>
      }
    </div>
  );
}

export default App;
```
And build it:

`npm run build`

### Run `mephisto review`

```bash
$ cat data.csv | mephisto review ~/review-test/build > results.csv
```

### DIFF
--- a/mephisto/client/cli.py
+++ b/mephisto/client/cli.py
@@ -24,6 +24,15 @@ def web():
     app.run(debug=False)
 
 
+@cli.command("review")
+@click.argument("build_dir", type=click.Path(exists=True))
+# @click.argument("index_path", type=click.Path(exists=True))
+def review(build_dir):
+    from mephisto.client.review_server import run
+
+    run(build_dir)
+
+
 @cli.command("check")
 def check():
     """Checks that mephisto is setup correctly"""
--- a/mephisto/client/review_server.py
+++ b/mephisto/client/review_server.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+from flask import Flask, Blueprint, send_file, jsonify
+from mephisto.client.config import Config
+from mephisto.client.api import api
+import os
+import atexit
+import signal
+import csv
+import sys
+import time
+import threading
+
+# app = Flask(__name__, static_url_path="/static", static_folder="../webapp/build/static")
+# app.config.from_object(Config)
+
+
+def run(build_dir):
+    global index_file, app
+    global ready_for_next, current_data, finished, index_file
+    app = Flask(
+        __name__,
+        root_path=os.getcwd(),
+        static_url_path="/static",
+        static_folder=build_dir + "/static",
+    )
+
+    def consume_data():
+        global ready_for_next, current_data, finished
+        data_source = csv.reader(iter(sys.stdin.readline, ""))
+
+        finished = False
+        for row in data_source:
+            ready_for_next = threading.Event()
+            current_data = row
+            sys.stdout.write("{!r}\n".format(row))
+            sys.stdout.flush()
+            ready_for_next.wait()
+        finished = True
+
+    @app.route("/data")
+    def data():
+        global current_data, finished
+        return jsonify(
+            {"finished": finished, "data": current_data if not finished else None}
+        )
+
+    @app.route("/next")
+    def next():
+        global current_data, ready_for_next, finished
+        ready_for_next.set()
+        return jsonify(finished)
+
+    @app.route("/")
+    def index():
+        global index_file
+        return send_file(build_dir + "/index.html")
+
+    @app.after_request
+    def after_request(response):
+        response.headers.add("Access-Control-Allow-Origin", "*")
+        response.headers.add(
+            "Access-Control-Allow-Headers", "Content-Type,Authorization"
+        )
+        response.headers.add(
+            "Access-Control-Allow-Methods", "GET,PUT,POST,DELETE,OPTIONS"
+        )
+        response.headers.add("Cache-Control", "no-store")
+        return response
+
+    term_handler = signal.getsignal(signal.SIGINT)
+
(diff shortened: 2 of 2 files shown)


mephisto/client/review_server.py
```diff
@@ -0,0 +1,103 @@
...
+import threading
+
+
+if False:  # sys.stdout.isatty():
+    pass
+else:
+    # disable all logging out to stdout
+    import logging
+
+    log = logging.getLogger("werkzeug")
```
JackUrb, line 23: What logger is this?

mephisto/client/review_server.py
```diff
@@ -0,0 +1,103 @@
...
+        data_source = csv.reader(iter(sys.stdin.readline, ""))
+
+        finished = False
+        counter = 0
+        for row in data_source:
+            ready_for_next = threading.Event()
+            current_data = row
+            counter += 1
+            ready_for_next.wait()
+        finished = True
```
JackUrb, line 51: In future iterations it'll be useful to specify data in/out types, but for now this is a great start for that.

JackUrb (approved):
This is exciting! I'm liking the vision of tooling to _create_ review tools. Hooks for `useMephistoReview` to plug directly into the `MephistoDB` when no specific file is specified (or some other way to support a default flow) would be really cool, but I'm getting ahead of myself there haha.

mephisto/client/review_server.py
pringshia: Yup! This is how I'm hoping to be inclusive towards use-cases like ego4d.

100% agreed that it would be nice to support more than just csv. I plan to add support for more I/O readers as this evolves. There are also some existing handy utils that allow these transformations, so you can pipe your datasource through a converter util & convert to csv as a stopgap measure. But for now I think this should be good to get us started, with maybe JSON just around the corner.

mephisto/client/review_server.py
pringshia: This is the suggested way to disable stdout logging for flask: https://stackoverflow.com/a/18379764

Since this util outputs the review data on stdout, I needed std out to be clean.
The name comes from an underlying lib that flask uses: https://werkzeug.palletsprojects.com/en/1.0.x/

pringshia:
> Hooks for `useMephistoReview` to plug directly into the `MephistoDB` when no specific file is specified (or some other way to support a default flow) would be really cool, but I'm getting ahead of myself there haha.

I have an internal quip where I already started brainstorming this, will link it to you :)

mephisto/client/review_server.py
JackUrb: It could make sense to require the review file name as an argument instead? I'm unsure anyone would specifically need the stdout setting.

pringshia: Good idea! I think that the file name will definitely be used a lot more. If anything I think we can move the cutesy stdout setting behind a flag instead of default behavior.

pringshia:
Re-requesting review. Full list of changes since last time are in the "Upcoming in the PR" section.

Future work items have had issues created for them to allow tracking.

There's now a `-o` command that allows you to output directly to a file. Using the `--stdin` flag is what triggers the previously default behavior since it isn't expected to be common usage.

mephisto/client/cli.py
```diff
@@ -27,11 +27,21 @@ def web():
...
 @click.argument("review_app_dir", type=click.Path(exists=True))
 @click.option("-p", "--port", type=(int), default=5000)
-def review(review_app_dir, port):
+@click.option("-o", "--output", type=(str), default="")
+@click.option("--stdout", "output_method", flag_value="stdout")
+@click.option("--file", "output_method", flag_value="file", default=True)
+@click.option("--csv-headers/--no-csv-headers", default=False)
+@click.option("-d", "--debug", type=(bool), default=False)
+def review(review_app_dir, port, output, output_method, csv_headers, debug):
     """Transform data from stdin to stdout via a webapp. Good for review."""
```
JackUrb, line 36: Nit description not fully up to date anymore.

mephisto/client/review_server.py
```diff
@@ -14,22 +14,20 @@
...
     global index_file, app
     global ready_for_next, current_data, finished, index_file
     global counter
+
+    if not debug or output == "":
+        # disable noisy logging of flask
+        import logging
+
+        log = logging.getLogger("werkzeug")
+        log.disabled = True
```
JackUrb, line 27: nit: naming can make this code more self-documenting. `flask_log` makes the implicit knowledge that the flask logger's name is `werkzueg` than just calling it `log`.

JackUrb (approved):
LGTM! Improvements over last review are nice, only minor nits.

mephisto/client/review_server.py
pringshia: Good suggestion! Much more clear than this stack overflow copy pasta

mephisto/client/cli.py
pringshia: good catch, fixed

PR #275, opened 2020-10-07
`bootstrap-chat` docs update

pringshia:
Finished up the `bootstrap-chat` docs.

Have one minor note below regarding the `IDLE` state for `INPUT_MODE` which doesn't seem to be ever set anywhere. Should we remove? Or does this represent a future feature we will be adding regarding idleness/timeouts?

### DIFF
--- a/packages/bootstrap-chat/README.md
+++ b/packages/bootstrap-chat/README.md
@@ -65,8 +65,7 @@ function MyChatApp() {
 
 Here we replace the rendering of the chat message with the `renderMessage` prop. This is one of the provided optional render overrides, for the others see below.
 
-However, note that we also import the `<ChatMessage />` component and use that as part of the rendering process. We don't have to do this portion, and could easily just create our own isolated implementation, e.g. something like `<div>{message.text + "!!!!!"}</div>`. However this example illustrates how we can opt out of default behavior, while still having available to us primitives and core components that can get us back some of that default functionality at our discretion. In this case, we
-ll still have the look and feel of the original chat message component, however with a custom message property.
+Note that we also import the `<ChatMessage />` component to use as part of the render method. We don't have to do this portion, and could easily just create our own isolated implementation, e.g. something like `<div>{message.text + "!!!!!"}</div>`. However this example illustrates how we can opt out of default behavior, while still having available to us primitives and core components that can get us back some of that functionality at our discretion. In this case, we'll still have the look and feel of the original chat message component, however with a custom message property.
 
 
 ## The `<ChatApp />` component
--- a/packages/bootstrap-chat/README.md
+++ b/packages/bootstrap-chat/README.md
@@ -101,13 +101,14 @@ Note: If `renderResponse()` is also implemented, this render prop will be ignore
 
 This render prop is for cases where you'd like to minimally alter the response text box. For larger customizations, we recommend implementing `renderReponse()` instead.
 
-```js
-{
-    onMessageSend,
-    inputMode,
-    active, // helper boolean, essentially is true when inputMode is INPUT_MODE.READY_FOR_INPUT
-}
-```
+**Arguments:**
+
+**`onMessageSend`**: 
+
+**`inputMode`**: 
+
+**`active`**: A helper boolean, essentially is true when `inputMode` is `INPUT_MODE.READY_FOR_INPUT`
+
 
 ### `renderResponse({ onMessageSend, inputMode, mephistoContext, appContext })`
 A render prop that if implemented, overrides the way the entire response panel is rendered. You will be responsible for handling all aspects of the panel in this case.
--- a/packages/bootstrap-chat/README.md
+++ b/packages/bootstrap-chat/README.md
@@ -115,7 +115,6 @@ This render prop is for cases where you'd like to minimally alter the response t
 - `INACTIVE`: Occurs when something goes wrong, usually if the agent experiences a disconnect, timeout, the task is returned, or has expired.
 - `DONE`: The task is complete.
 - `READY_FOR_INPUT`: The current agent is ready and is ready to send over data.
-- `IDLE`: (TODO: This doesn't seem to be set anywhere. Should we remove?)
 
 **`active`**: A helper boolean, essentially is true when `inputMode` is `INPUT_MODE.READY_FOR_INPUT`
 
--- a/packages/bootstrap-chat/README.md
+++ b/packages/bootstrap-chat/README.md
@@ -24,4 +24,4 @@ npm link bootstrap-chat
 
 ```
 
-## Usage (`useMephistoTask`)
+## Usage
--- a/packages/bootstrap-chat/README.md
+++ b/packages/bootstrap-chat/README.md
@@ -133,7 +133,7 @@ All render props will receive the `appContext` argument. This context is general
 
 **`taskContext`**: 
 
-**`appSettings`**: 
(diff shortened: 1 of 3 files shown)


packages/bootstrap-chat/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+### `onMessagesChange(messages)`
+
+A callback that will get called any time a new message is received. The callback will receive as an argument all messages received thus far.
+
+### The `mephistoContext` argument
+
+All render props will receive the `mephistoContext` argument.
+
+This object is basically the return value of calling the `useMephistoLiveTask` hooks. Further documentation on these properties can be found in the documentation for the `mephisto-task` package under the `useMephistoLiveTask` section.
```
JackUrb, line 126: Nit: worth linking this.

packages/bootstrap-chat/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+### The `appContext` argument
+
+All render props will receive the `appContext` argument. This context is generally used to handle app-specific information, as opposed to Mephisto-specific information.
```
JackUrb, line 130: Nit - isn't `appContext` defined by `bootstrap-chat`, such that you could note all of the default arguments. (Also, is it possible to add arguments to appContext?)

packages/bootstrap-chat/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+
+### `renderMessage({ message, idx, mephistoContext, appContext })`
+A render prop that if implemented, overides the way a chat message is rendered.
+
+**Arguments:**
+
+**`message`**: An object with the following schema:
+
+TODO: where does this object come from in the backend? Find to verify the below
```
pringshia, line 81: Want to flag this question - where exactly does the `message` object schema get defined in the backend?

packages/bootstrap-chat/README.md
```diff
@@ -0,0 +1,141 @@
...
+A render prop that if implemented, overrides the way the response text box is rendered. The default implementation of `ChatApp` is rendering a `<TextResponse />` component here.
+
+Note: If `renderResponse()` is also implemented, this render prop will be ignored.
+
+This render prop is for cases where you'd like to minimally alter the response text box. For larger customizations, we recommend implementing `renderReponse()` instead.
+
+```js
+{
+    onMessageSend,
+    inputMode,
```
pringshia, line 107: TODO: list out all the input modes

packages/bootstrap-chat/README.md
pringshia: yep, we can definitely do this. I'll add it down below. Add the moment it's not possible to add more arguments, will also make a note of that

packages/bootstrap-chat/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+**Arguments:**
+
+**`taskContext`**: 
+
+**`appSettings`**: You can use this to have access to various app-wide settings. By default it is used for volume control.
+
+**`setAppSettings`**: 
+
+**`onTaskComplete`**:
```
pringshia, line 140: TODO: double check why we needed this method

packages/bootstrap-chat/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+**Arguments:**
+
+**`message`**: An object with the following schema:
+
+TODO: where does this object come from in the backend? Find to verify the below
+
+```
+{
+    message_id,
```
pringshia, line 85: TODO: I also see an id property used in certain places, but also message_id.... is there a difference between the two?

pringshia (commented):
some WIP comments & TODOs

packages/bootstrap-chat/README.md
JackUrb: This actually comes all the way from the ParlAI definition of the [`Message`](https://github.com/facebookresearch/ParlAI/blob/bca1dd153845c265f9ce53fe0ff0ab51666c878d/parlai/core/message.py#L14) class, but I'm realizing now that that no longer explicitly sets certain fields. We should probably standardize some keys that we expect here though:

- `'id'`: actually the speaker id for the message, but ParlAI has called this the id
- `'message_id'`: We add this when a message is passed to ensure it's always unique, no funny business.
- `'text'`: This contains the actual message text, and the message should not be displayed if this is undefined or the empty string
- `'task_data'`: This is the dict that is used to update the `taskContext` variable, and is later accessible from the message for specific messages.

packages/bootstrap-chat/README.md
JackUrb: Addressed in another comment.

packages/bootstrap-chat/README.md
JackUrb: The response pane is responsible for displaying the done button, which submits the task calling this method.

packages/bootstrap-chat/README.md
pringshia: Thanks! Added this info.

packages/bootstrap-chat/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+**`onMessageSend`**: A callback provided for you to invoke in your implementation. Invoking this callback will send the passed in message back to the server. Example usage: `onMessageSend({ text: textValue, task_data: {} })`
+
+**`inputMode`**: Represents an enumeration for the status of the chat task. It is set to one of the following:
+
+- `WAITING`: The current agent is ready but is waiting for a response from the server, e.g. waiting for an incoming chat message.
+- `INACTIVE`: Occurs when something goes wrong, usually if the agent experiences a disconnect, timeout, the task is returned, or has expired.
+- `DONE`: The task is complete.
+- `READY_FOR_INPUT`: The current agent is ready and is ready to send over data.
+- `IDLE`: (TODO: This doesn't seem to be set anywhere. Should we remove?)
```
pringshia, line 118: So we have an `IDLE` state for the `INPUT_MODE` enum, however I don't see usage anywhere. The default implementation doesn't ever set the state to `IDLE` for any case. Thoughts on whether to keep this enum or delete?

packages/bootstrap-chat/README.md
```diff
@@ -0,0 +1,157 @@
...
+            renderMessage={({ message }) => <ChatMessage message={message.text + "!!!!!"} />}
+        />);
+}
+
+```
+
+Here we replace the rendering of the chat message with the `renderMessage` prop. This is one of the provided optional render overrides, for the others see below.
+
+However, note that we also import the `<ChatMessage />` component and use that as part of the rendering process. We don't have to do this portion, and could easily just create our own isolated implementation, e.g. something like `<div>{message.text + "!!!!!"}</div>`. However this example illustrates how we can opt out of default behavior, while still having available to us primitives and core components that can get us back some of that default functionality at our discretion. In this case, we
+ll still have the look and feel of the original chat message component, however with a custom message property.
```
JackUrb, line 69: nit: newline breaks "we'll". Also may be worthwhile to drop one of the however's in the above block for flow.

packages/bootstrap-chat/README.md
JackUrb: Yeah we can kill `IDLE`, I don't think we need it anymore.

JackUrb (approved):
Great docs - really solid overview for how `bootstrap-chat` is intended to be used, and important documentation of the heaviest-lifting parts of the package. 

Reading through though, it may be also good to document _some_ of the more-likely-to-be-used underlying components, such as `ChatMessage`, perhaps in a `bootstrap-chat/src/README.md`.

PR #279, opened 2020-10-12
Add `mephisto review --json` flag. Default reading in is csv

pringshia:
Adding a `--json` flag will parse the input stream as a list of newline separated json objects. The default option is `--csv`.

---

Sample usage:

**data.json**
```
{"value": 1}
{"value": 2}
{"value": 3}
{"value": 4}
```

```cat data.json | mephisto review my-review-ui/ --json --stdout```


JackUrb (approved):
lgtm

PR #283, opened 2020-10-21
Swapping order of operator abstraction initialization

JackUrb:
As other elements of setup (the Architect primarly) may rely on options set by a blueprint, it's important that we initialize the blueprint before the other abstractions.

This was surfaced in a strange internal case where we couldn't launch a server without the task details, but realistically this ordering should be more correct overall.


pringshia (approved):
Makes sense to me 👍

PR #286, opened 2020-10-21
Javascript Dependency update script

JackUrb:
Creates a simple bash script that automatically goes through every folder with a `package.json`, runs `npm install; npm audit fix; yarn install` to generate new lock files with safe versions, then removes the excess `package-lock.json` and `yarn.lock` files that weren't there to begin with.

Not quite dependabot with updated lists of how likely these changes are to be breaking, but it at least works :).


pringshia:
love love love this! #betterengineering!

pringshia (approved):

pringshia:
Merging in because the alert is annoying lol

PR #288, opened 2020-10-22
Add an index docs guide

pringshia:
Adds an index README to the docs page as a helpful jumping off point to some of the other docs we have.


JackUrb (approved):
Seems like a good start to me!

PR #289, opened 2020-10-22
Cleanup script for MTurk

JackUrb:
Adds a simple script for being able to clean up HITs from MTurk by checking the available requesters, being able to see what HITs are available to remove, and then removing them.

# Testing

```
>>>> python cleanup.py
You have the following requesters available for mturk and mturk sandbox:
['XXXXXXXXX', 'XXXXXXXXX_sandbox']
Enter the name of the requester to clear HITs from: 
>> XXXXXXXXX
The requester XXXXXXXXX has 6 outstanding HIT types, with 0 suspected active or broken HITs.
This may include tasks that are still in-flight, but also tasks that have already expired but have not been disposed of yet.
Would you like to cleanup by (t)itle, or just clean up (a)ll? 
>> t
HIT TITLE: ----------------------removed--------------
HIT COUNT: 114
Should we cleanup this hit type? (y)es for yes, anything else for no: 
>> y
HIT TITLE: ----------------------removed--------------
HIT COUNT: 118
Should we cleanup this hit type? (y)es for yes, anything else for no: 
>> y
HIT TITLE: ----------------------removed--------------
HIT COUNT: 110
Should we cleanup this hit type? (y)es for yes, anything else for no: 
>> y
HIT TITLE: ----------------------removed--------------
HIT COUNT: 30
Should we cleanup this hit type? (y)es for yes, anything else for no: 
>> y
HIT TITLE: ----------------------removed--------------
HIT COUNT: 104
Should we cleanup this hit type? (y)es for yes, anything else for no: 
>> y
HIT TITLE: Which Conversational Partner is Better?
HIT COUNT: 40
Should we cleanup this hit type? (y)es for yes, anything else for no: 
>> y
Disposing 516 HITs.
Disposed!



>>>> python cleanup.py
You have the following requesters available for mturk and mturk sandbox:
['XXXXXXX', 'XXXXXXX_sandbox']
Enter the name of the requester to clear HITs from:
>> XXXXXXX_sandbox
The requester XXXXXXX_sandbox has 13 outstanding HIT types, with 209 suspected active or broken HITs.
This may include tasks that are still in-flight, but also tasks that have already expired but have not been disposed of yet.
Would you like to cleanup by (t)itle, or just clean up (a)ll?
>> a
Disposing 215 HITs.
After disposing, 209 could not be disposed.
These may not have been reviewed yet, or are being actively worked on.
They have been expired though, so please try to dispose later.
```

### DIFF
--- a/mephisto/scripts/mturk/cleanup.py
+++ b/mephisto/scripts/mturk/cleanup.py
@@ -1,0 +1,89 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+Utility script that finds, expires, and disposes HITs that may not
+have been taking down during a run that exited improperly.
+"""
+from mephisto.providers.mturk.mturk_utils import (
+    get_outstanding_hits,
+    expire_and_dispose_hits,
+)
+from mephisto.core.local_database import LocalMephistoDB
+
+db = LocalMephistoDB()
+
+all_requesters = db.find_requesters(provider_type="mturk")
+all_requesters += db.find_requesters(provider_type="mturk_sandbox")
+
+print("You have the following requesters available for mturk and mturk sandbox:\n>> ")
+r_names = [r.requester_name for r in all_requesters]
+print(sorted(r_names))
+
+use_name = input("Enter the name of the requester to clear HITs from: ")
+while use_name not in r_names:
+    use_name = input(
+        f"Sorry, {use_name} is not in the requester list. "
+        f"The following are valid: {r_names}\n"
+        f"Select one:\n>> "
+    )
+
+requester = db.find_requesters(requester_name=use_name)[0]
+client = requester._get_client(requester._requester_name)
+
+outstanding_hit_types = get_outstanding_hits(client)
+num_hit_types = len(outstanding_hit_types.keys())
+sum_hits = sum([len(outstanding_hit_types[x]) for x in outstanding_hit_types.keys()])
+
+all_hits = []
+for hit_type in outstanding_hit_types.keys():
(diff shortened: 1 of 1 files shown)


bottler:
This is very useful. 

I observed it's very easy to hit rate limits on mturk when trying to clear up dozens/hundreds of hits. In my code I sleep a second between each delete, which works but might be overkill.

mephisto/scripts/mturk/cleanup.py
```diff
@@ -0,0 +1,89 @@
...
+use_name = input("Enter the name of the requester to clear HITs from:\n>> ")
+while use_name not in r_names:
+    use_name = input(
+        f"Sorry, {use_name} is not in the requester list. "
+        f"The following are valid: {r_names}\n"
+        f"Select one:\n>> "
+    )
+
+requester = db.find_requesters(requester_name=use_name)[0]
+client = requester._get_client(requester._requester_name)
```
pringshia, line 35: just curious, what does the the `_` at the start of `_get_client` signify? does this signify a private method?

pringshia (approved):
Looks like Christmas came early this year... Appreciate all these usability enhancement PRs. Looks good to me.

I wonder if I should add something to the `mephisto` CLI command that can be used to run script files from this directory, similar to how `npm` can run scripts defined in `package.json`. So then someone could just do `mephisto run cleanup` from anywhere in the project structure.

mephisto/scripts/mturk/cleanup.py
JackUrb: Yeah, `_get_client` really shouldn't be being used directly in user code, as it hits on the low-level API.

JackUrb:
Agreed that it would be reasonable to start mapping from the CLI to some of the scripts present here, though I'd still prefer they have the given subdivisions (like `mephisto run mturk cleanup` or `mephisto run mturk/cleanup`).

PR #290, opened 2020-10-23
Unit expiration timer with tqdm

JackUrb:
Generally the shutdown experience needs some work, and one big component of that is feedback. This PR adds in `tqdm` so that users can see the expirations occurring after they have shut the system down, possibly preventing them from repeatedly exiting.

# Testing
Ran a job with 20 tasks on sandbox, then expired:
```
[2020-10-23 13:06:51,715][mephisto.core.operator][INFO] - expiring units
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 20/20 [00:07<00:00,  2.51it/s]
```

### DIFF
--- a/mephisto/core/task_launcher.py
+++ b/mephisto/core/task_launcher.py
@@ -18,6 +18,7 @@ from mephisto.data_model.assignment import (
 )
 
 from typing import Dict, Optional, List, Any, TYPE_CHECKING, Iterator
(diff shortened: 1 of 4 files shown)


pringshia:
Awesome usability enhancement, and love how simple it was to add in.

pringshia (approved):

PR #291, opened 2020-10-26
Remove old references

EricMichaelSmith:
Remove references to an old ParlAI MTurk task that will be removed as of https://github.com/facebookresearch/ParlAI/pull/3225


JackUrb (approved):
Great flag - though these three files should actually be just removed entirely! They're pretty old documentation that no longer applies.

EricMichaelSmith:
> Great flag - though these three files should actually be just removed entirely! They're pretty old documentation that no longer applies.

Okay cool, I remember that these are analogous to other ones that you mentioned could be removed. I've removed these 3 READMEs and am requesting re-review (just in case)

PR #296, opened 2020-10-29
AbstractTestSupervisor

EricMichaelSmith:
Abstract away the core functionality of the TestSupervisor so that it can be used for ParlAI end-to-end tests of Blueprints' inputs and outputs; move the abstract base class to `mephisto/data_model/test/utils.py`


JackUrb (commented):
I'm unsure making this abstraction is necessary - more in-depth comment on the ParlAI PR

PR #295, opened 2020-10-29
Refactor a lot of paths to re-organize structure

JackUrb:
As noted in #285, updating the paths. The goal here is to better follow the overall architecture design document. Most extra lines come from newly committed lock files. We're moving to this format so that the system as a whole is more understandable as we look to finalize components of the design. Nothing has been deleted in this PR, just moved, so all existing code should still work. If you're running on this tag though and see warnings, that means you'll need to make some changes before updating to v0.3.

# How to switch over?
If you're reading this, it's likely because you're about to do a bunch of find replaces in order to ensure that your code is now up to the new structure. Following this list of find-replaces will likely help. Here's a list of the changes, in order of likely relevance:
- `mephisto.core.local_database` is now `mephisto.abstractions.databases.local_database`
- `mephisto.core.data_browser` is now `mephisto.tools.data_browser`
- the rest of `mephisto.core.` is now at `mephisto.operations.`
- `mephisto.utils.scripts` is now `mephisto.tools.scripts`
- `mephisto.data_model.database` is now `mephisto.abstractions.database`
- `mephisto.server.blueprints.` is now `mephisto.abstractions.blueprints.`
- `mephisto.server.architects.` is now `mephisto.abstractions.architects.`
- `mephisto.providers.` is now `mephisto.abstractions.providers.`
- `mephisto.data_model.assignment.Unit` has been broken out into `mephisto.data_model.unit`
- `mephisto.data_model.task.TaskRun` has been broken out into `mephisto.data_model.task_run`
- `mephisto.data_model.assignment_state` is now `mephisto.data_model.constants.assignment_state`
- `mephisto.server.channels.channel` is now `mephisto.abstractions.channel`
- `mephisto.server.channels.websocket_channel` is now `mephisto.abstractions.architects.channels.websocket_channel`
- `mephisto.data_model.architect` is now `mephisto.abstractions.architect`
- `mephisto.data_model.crowd_provider` is now `mephisto.abstractions.crowd_provider`
- `mephisto.data_model.blueprint` is now `mephisto.abstractions.blueprint`

### TODO
- [x] move the files
- [x] run the tests before updating callsites
- [x] run the tests after updating callsites
- [x] document how to transition
- [x] rebase on master

PR #297, opened 2020-10-29
Mass delete redirection files

JackUrb:
Following creating the new paths in #295 and having them redirect, this PR removes the old locations, leaving the clean new Mephisto structure underneath. It also updates a few inline locations that were still using the wrong import location for `Unit`.

All tests still pass, and this should be good to go for `v0.3.0`.

TODO
- [x] remove the files
- [x] Run the tests, assert still passing
- [x] Merge master
- [x] READMEs in the new dirs (handled in #316)


pringshia (approved):
Didn't look through all the files but if tests pass 👍

PR #302, opened 2020-11-02
Tweak adding the world module to the path

EricMichaelSmith:
It appears that, if the worlds module is at `/dummy/path/here/demo_worlds.py`, the code will append `/dummy/path/here/demo_worlds` to the Python path. This PR changes this code to instead add `/dummy/path/here`, which will allow `demo_worlds.py` to be detected by the interpreter


JackUrb (approved):
LGTM. This was always a little hacky, so I'm not surprised it needed a touch-up. Thanks for the fix!

PR #309, opened 2020-11-06
Small ParlAI blueprint improvements

JackUrb:
Adds two improvements to the `ParlAIChatBlueprint`:

1. For builds using the simple directory build process (relying on `bootstrap-chat`), these directories can now contain a `src/` directory that `main.js` refers to for subcomponents.
2. Now that `shared_state` is standard, we can pass the actual world module from the run script in this to the blueprint rather than needing to pass the path, which makes this code compatible with compiled/bundled/packaged python scripts like we use internally.

These are used in the `TurnAnnotationsBlueprint` being created for ParlAI.


EricMichaelSmith (approved):
Cool, seems reasonable!

pringshia:
Looks good! So will `main.js` have to access via `"./components/Custom"` ? I see a `src/components` folder in the diff. Should any documentation be added/updated as part of this PR?

pringshia (approved):

PR #310, opened 2020-11-10
Small fix to redundant call in Operator.

JackUrb:
Due to the way that blueprints are initialized in a `TaskRun`, the current Mephisto code was initializing a blueprint, then immediately discarding it and building another. This redundancy can be expensive if blueprint initialization contains a lot of setup.

It also adds an `id` field to the ParlAI wrapper for a Mephisto agent, as the `id` field is sometimes referenced from ParlAI worlds rather than `agent_id`.


JackUrb:
Pulling this in as tests still pass, changes are minor, and this is blocking a ParlAI PR.

PR #311, opened 2020-11-11
Resolve double message regression

JackUrb:
Resolves a regression of #107 that was re-introduced in the swap over to `mephisto-task`. Ensures all messages have a message id, and then filters out duplicates on the `mephisto-task` level. I'm still unsure why the double message happens in the first place (likely the router's first connection, or a lack of ack or something), but this will prevent it moving forward.


pringshia (approved):
LGTM

PR #316, opened 2020-11-18
Full extended docs for every core directory in Mephisto

JackUrb:
This PR aims to include documentation for all of the new mid-level directories that properly separate the components of the Mephisto architecture. It hits all of the top level directories, and every sub directory, up until the subdivisions of the `abstractions` directory (doesn't include specific docs for abstraction implementations yet). 

It also moves some stuff from `data_model/test` to `abstractions/test`, as I noticed while working on their docs they were testing what are now abstractions and made sense to move. 

Tests pass locally.


pringshia (approved):
Looks great! Great to have docs in a good place with the new directory structure. Just flagged some minor typos.

PR #317, opened 2020-11-30
Bringing back argument parsing functionality.

JackUrb:
Some merging madness dropped these utilities out of argparse parser and I needed to bring them back. They replace the functionality that argparse parser had for determining relevant arguments of Mephisto abstractions, but now implemented using the data classes.


pringshia (approved):
LGTM

PR #319, opened 2020-12-01
Change how the ParlAI chat world is loaded throughout

EricMichaelSmith:
Apply the path fix in https://github.com/facebookresearch/Mephisto/pull/302 to two other places in the code


JackUrb (approved):
Nice, though the next improvement is likely going to be the ability to directly pass the module itself over via `SharedTaskState`, as was partially introduced [here](https://github.com/facebookresearch/Mephisto/pull/309/files#diff-0c59e331655a245a9d5d06fa6a2fff4a27914744fb802633b2b1fa0f2537acb4R61).

I note that the `ParlAIChatTaskRunner` is not actually prepared to load in a shared-state provided world module yet, so this doesn't need to happen immediately.

PR #321, opened 2020-12-01
Updating `mephisto web` to load again.

JackUrb:
Part of the hydra and path migration broke the `mephisto web` cli method for launching the webserver, and some of the server components themselves. This fix at least gets the server running again, though more (very) minor work may need to be done to get it up to speed.


pringshia (approved):
thanks for the fix

PR #326, opened 2020-12-04
ParlAIBlueprint - Adding initializationData to make_world

JackUrb:
# Overview
Right now it's somewhat nebulous how a ParlAI task world would be able to access the `InitializationData` that is provided for the assignment that world is being built for. As it's likely that this information is useful to the world, this PR adds the ability to pass the argument to the `make_world` function.

Not all world modules will be expecting the new `initialization_data` argument, so it's wrapped in a `try-catch`.

# Testing
Ran a quick demo world that was able to receive this argument, seemed to work fine both with and without the extra argument.


jxmsML (approved):
Thanks a lot for debugging this!

PR #328, opened 2020-12-10
Multimodal chat

JackUrb:
# Summary
This PR adds a multimodal chat view that can put two agents into the same shared environment. In this case, both agents are the same body in the environment, such that one can control and the other observes. Many thanks to @pringshia for the 3d rendering component.

# Testing
Runs so far! Video shared internally, will put an external test video up soon

# Still TODO
- [ ] Gotta finish up the README
- [ ] Need to pick a different resource to use
- [ ] Need to land a version of bootstrap-chat that only makes the `ding` sound on incoming text messages, not whenever act is requested.
- [ ] Maybe a fix for moving when WASD is pressed from the actor typing a message to send

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,7 @@ examples/**/build/*
 **/_generated/*
 **/outputs/*
 .coverage
-
+examples/multimodal_chat/resources/*
 
 # PyCharm
 .idea
--- a/examples/multimodal_chat/README.md
+++ b/examples/multimodal_chat/README.md
@@ -1,0 +1,2 @@
+# Multimodal Chat
+README TODO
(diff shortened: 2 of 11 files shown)


.gitignore
```diff
@@ -14,7 +14,7 @@ examples/**/build/*
 **/_generated/*
 **/outputs/*
 .coverage
-
+examples/multimodal_chat/resources/*
```
pringshia, line 17: maybe we should put this in a .gitignore inside the `multimodal_chat` directory to have it closer to the source? i'm open to either way. The thinking is that if examples move around, get renamed, or get removed in the future, the links in the global .gitignore might go stale, but this is less likely to occur for the .gitignore within the example proj directory

pringshia (approved):

pringshia:
I didn't realize that the ding sound plays on every message received, which could definitely be super noisy in this case!

PR #336, opened 2020-12-16
Some improvements for ParlAI chat blueprint and bootstrap-chat

JackUrb:
Title. Included improvements/fixes:

1. Adding a style file to the auto-build script for simple ParlAI frontend builds. This standardizes the method for updating style for an app. @mojtaba-komeili 
2. Fixes specifying the world module via `SharedParlAITaskState.world_module`. Code written with this style is more easily used internally.
3. Makes it possible to set default settings for the `ChatApp`. I wanted to at least be able to set volume, but as we may expand some of the settings in the future I thought I'd just make a way to initialize the whole thing. @pringshia is this a good way to do this? Or was there an alternate method for doing this already present somehow?

I included updates to the READMEs for all of these.

### DIFF
--- a/packages/bootstrap-chat/README.md
+++ b/packages/bootstrap-chat/README.md
@@ -131,7 +131,7 @@ The arguments for this render prop are the same as for `renderTextResponse`, wit
 
 A callback that will get called any time a new message is received. The callback will receive as an argument all messages received thus far.
 
-### `propAppSettings`
+### `defaultAppSettings`
 
 This optional prop lets you provide some initial app settings for the `ChatApp`, setting some default values for `appContext.appSettings`.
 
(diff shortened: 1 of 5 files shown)


packages/bootstrap-chat/README.md
```diff
@@ -131,6 +131,10 @@ The arguments for this render prop are the same as for `renderTextResponse`, wit
 
 A callback that will get called any time a new message is received. The callback will receive as an argument all messages received thus far.
 
+### `propAppSettings`
```
pringshia, line 134: the approach looks good to me, i would just rename the prop to something like `defaultAppSettings`, using `prop-` in a prop name isn't very conventional

pringshia (approved):
LGTM, just a naming suggestion. Thanks for the docs update!

PR #339, opened 2020-12-16
Adding a post task end hook

JackUrb:
# Overview
This PR introduces a new `SharedTaskState` hook that can now be used to automatically run any arbitrary script after a task has completed. This can be used to run quick post-task validation, or to update local state based on completion of a task.

# Implementation
The implementation is rather simple, just providing the function signature (with a default that is a no-op), and then calling it at the end of every run `Unit` or `Assignment`.

# Testing
I'm using this live in a task with the following function:
```
    def validate_final_unit(unit):
        try:
            u_data = mephisto_data_browser.get_data_from_unit(unit)
        except AssertionError:
            return  # unit was not in a final state, return or timeout
        ratings = u_data['data']['outputs']
        if (...some_validation_on_ratings_fails...):
            worker = unit.get_assigned_agent().get_worker()
            worker.grant_qualification(BLOCKING_QUALIFICATION, 1)
```

# Discussion
I don't think the name `validate_final_unit` properly describes what this hook is for, but I don't really know what better name to give it.

### DIFF
--- a/mephisto/abstractions/blueprint.py
+++ b/mephisto/abstractions/blueprint.py
@@ -77,7 +77,7 @@ class SharedTaskState:
     worker_can_do_unit: Callable[["Worker", "Unit"], bool] = field(
         default_factory=lambda: (lambda worker, unit: True)
     )
-    validate_final_unit: Callable[["Unit"], None] = field(
+    on_unit_submitted: Callable[["Unit"], None] = field(
         default_factory=lambda: (lambda unit: None)
     )
 
@@ -208,7 +208,7 @@ class TaskRunner(ABC):
 
             traceback.print_exc()
             self.cleanup_unit(unit)
-        self.shared_state.validate_final_unit(unit)
+        self.shared_state.on_unit_submitted(unit)
         del self.running_units[unit.db_id]
         return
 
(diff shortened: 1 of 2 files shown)


pringshia (commented):
I agree that the name is a bit confusing.  So if I understand correctly, when a task run is complete, this hook gets executed once per each Unit in case the user wants to do some sort of processing on the unit?

The final unit part makes it sound almost like it only runs on the last, final unit, instead of running "finally" on all units.

JackUrb:
So it runs when a unit is completed, every time a unit is completed, throughout the run. This is useful as for instance someone can write a script that easily determines if the given data was junk, and then ban a worker before they accept any more tasks.

pringshia:
Got it, in the docs I read "when the TaskRunner ends" so got confused. For lifecycle hook methods I like names like `on_unit_received`, `on_unit_submitted`, `validate_unit_on_completion`

pringshia (approved):
Don't want to block this on the name change, so will go ahead and approve, the functionality looks ok. Happy with whatever you decide to go with.

JackUrb:
Totally forgot to follow through on this - great suggestions for naming choices, will try to follow this pattern for any other hooks we add.

PR #340, opened 2020-12-16
Ensuring proper shutdown with limited unit launch

JackUrb:
# Overview
A small bug in the ordering between when `shutdown` and `expire_units` are called by an `Operator` (units are expired after shutdown), tasks launched using a unit limit would hang indefinitely when shutting down early. This fixes that detail by ensuring that `shutdown` will trigger closing of the unit launching thread.


pringshia:
Nice bug fix

pringshia (approved):

PR #341, opened 2020-12-16
Improving run speed of Mephisto IO bottlenecks

JackUrb:
# Overview
At the moment, some of the largest bottlenecks for a Mephisto run happen during database and file read/write interactions. This PR attempts to make some forward progress along these issues, first by selecting one level of speed over safety on our SQlite databases, and by deferring loads of data from file until they're absolutely necessary.

# Implementation
1. ~~Set `PRAGMA synchronous = 1` across all of our SQLite databases. SQLite's default synchronicity setting is `SAFE`, which checks the database file a few times during the write, and exists to be resilient to critical failure (application error, system exit, power outage) during writing operations across multiple processes. Mephisto doesn't require _quite_ this level of safety, and we can tolerate the change to a normal operation mode for a speed boost.~~
2. Defer the loading of an `Agent`'s `AgentState` until the property is actually accessed. Some situations load up an `Agent` just to check the `Agent`'s status, or static attributes. We don't need to load up the saved local `AgentState` from disk just to handle these operations. 

# Notes
For some reason it seems like making the deferred state loading change is having a bad interaction with static tasks, wherein some tasks lose their state and have it replaced with the empty state, or the state containing only the inputs, even though the task is successfully submitted and we can retrieve the data from MTurk.

### DIFF
--- a/mephisto/abstractions/databases/local_database.py
+++ b/mephisto/abstractions/databases/local_database.py
@@ -269,7 +269,6 @@ class LocalMephistoDB(MephistoDB):
         with self.table_access_condition:
             conn = self._get_connection()
             conn.execute("PRAGMA foreign_keys = 1")
-            conn.execute("PRAGMA synchronous = 1")
             with conn:
                 c = conn.cursor()
                 c.execute(CREATE_PROJECTS_TABLE)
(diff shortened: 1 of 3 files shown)


JackUrb:
Marking ready for review, as profiling suggests this small change alone can result in a reduction of runtime cost by ~30%. @mojtaba-komeili let me know how a run with this goes.

mojtaba-komeili (approved):

PR #343, opened 2020-12-17
UI Error Logging

pringshia:
_#237 went a little stale (it was from before The Great Refactor), so re-did it with all of @saraEbrahim's awesome contributions as well as some additional features here._

---

This PR adds features to make for a better experience when a UI task encounters uncaught errors.

This better experience is targeted for both crowd workers and Mephisto task developers.

- Crowd workers: will get a better path to resolution when an error does occur, as well as some sort of visual feedback that something went wrong.
- Task creators: will have less angry crowd workers, will hopefully be pinged earlier when an error does occur, and in the case of non-static-HTML tasks, e.g. using `mephisto-task`, will have a way to auto-log errors.

For now, these features are opt-in only via a config change. In the future they may be shipped enabled by default.

## Example 1: Opt into error handling behavior for a static HTML task

In `examples/simple_static_task/server_files/demo_task.html`, we add the following lines:

Screen Shot 2020-12-17 at 5 07 45 PM

By setting the global var `_MEPHISTO_CONFIG_.ADD_ERROR_HANDLING` to `true`, we opt into the behavior added by this PR.
We then force an error for test purposes.

Screen Shot 2020-12-17 at 5 00 16 PM

On hitting OK:

Screen Shot 2020-12-17 at 5 17 21 PM

---

## Example 2: Opt into error handling behavior for a React task

Within, `examples/static_react_task/webapp/src/components/core_components.jsx:`, we add a forced error as shown in the last line of `SimpleFrontend` as shown here:

Screen Shot 2020-12-17 at 5 45 07 PM
Screen Shot 2020-12-17 at 5 43 56 PM

We also opt into error handling via the global config var:

Screen Shot 2020-12-17 at 5 47 24 PM

And we now get error handling for React tasks as well:

Screen Shot 2020-12-17 at 5 45 07 PM

---

## Example 3: `ErrorBoundary` and `handleFatalError`

You can also wrap your entire app with a provided `<ErrorBoundary />` component, eg.:

```jsx
import { ErrorBoundary } from "mephisto-task";
...
return (
  <ErrorBoundary>
    <MyApp />
  </ErrorBoundary>
);
```

This now gives you an overlay instead of a blank screen on errors:

Screen Shot 2020-12-17 at 5 45 16 PM

Additionally, you can also wire up your own auto-error logging by using a new method exposed by the `useMephistoTask` hook:

```jsx
import { ErrorBoundary } from "mephisto-task";
...
const { handleFatalError, /* ... */ } = useMephistoTask();
...
return (
  <ErrorBoundary handleError={handleFatalError}>
    <MyApp />
  </ErrorBoundary>
);
```

`handleFatalError` can be used with `<ErrorBoundary />'s `handleError` prop or used in any custom logic code you wish - for example, in handling errors for AJAX requests which live outside of the scope of React Error Boundaries.

Error packets can be seen in the console logs:

Screen Shot 2020-12-17 at 5 46 19 PM


JackUrb:
A lot of this I think is great and ready to go, though I'm still a little concerned by the placement of the extra handling in `wrap_crowd_source.js`. If it needs to exist outside of the react that's fine, but then it would be nice to have some documentation about it (where we document crowd providers) and we should update all of the `wrap_crowd_source.js` files to have some amount of this inside.

JackUrb (approved):

pringshia:
@JackUrb - I'll add documentation for this at `mephisto/abstractions/providers/README.md`

I'll also add some documentation for how to enable this feature in it's own README (will copy over most of what's in this PR). I'm open to improving on this feature, but this may be good for a beta and we can iterate as we get feedback. For now this is also opt-in only (will have directions documented, so we can phase this in).

PR #335, opened 2020-12-16
Parlai chat initialization data

jxmsML:
Add

PR #345, opened 2020-12-23
add initialization data during onboarding world

jxmsML:
Add initialization data during parlai chat `make_onboarding_world`, similar to https://github.com/facebookresearch/Mephisto/pull/335 and https://github.com/facebookresearch/Mephisto/pull/326

This could be useful, for example to get onboarding data within `MTurkOnboardWorld` without having race condition.


JackUrb (approved):
LGTM!

Issue #327, opened 2020-12-07
ParlAI Chat Demo - Something wrong with worker pairing and agent status updates


vaibhavad:
Hi,

I have been doing some pilot studies for a crowdsourcing task that pairs 2 MTurk workers for a conversation. I noticed some unusual behavior in some assignments. Specifically, there are cases where the worker starts the HIT and gets the partner timeout message within a few seconds. I'll list down the steps to replicate the task setup and provide the logs below.

I forked Mephisto ([vaibhavad/Mephisto](https://github.com/vaibhavad/Mephisto)) (just a few hours ago) and made some minor changes:
1) Changed the logging (vaibhavad/Mephisto@f5158ed0ce0fd91a215cd13b6482ef21e6c510d9) in `supervisor.py`, `operator.py`, and `blueprint.py` so that issue-specific information is logged.
2) Statically linked packages (vaibhavad/Mephisto@a18ac6e9967fd016be3dafb740744fcee3e37fb6) as somehow node modules we not working on my system (#325)
3) Changed task configuration to 10 conversations (vaibhavad/Mephisto@c00c4b488146f247697580c2f0086373623ff02a), and using `custom_prebuilt.yaml` so that `bundle.js` is picked from `webapp`.

Running the task  - 
I used the following steps
```
git clone https://github.com/vaibhavad/Mephisto.git
pip install parlai
cd Mephisto
pip install -e .
mephisto register mturk_sandbox name=my_mturk_user_sandbox access_key_id=<ACCESS_KEY> secret_access_key=<SECRET_KEY>
cd Mephisto/packages/mephisto-task
npm install; npm run dev
cd Mephisto/packages/bootstrap-chat
npm install; npm run dev
cd Mephisto/examples/parlai_chat_task_demo/webapp
npm install; npm run dev
cd ..
python parlai_test_script.py mephisto/architect=heroku mephisto.provider.requester_name=my_mturk_user_sandbox
```

I tested the system using three different MTurk Sandbox accounts, so 3 workers are registered. I kept returning the task midway frequently and starting new ones from different accounts to replicate the scenario I observed in production. Here are some observation (with reference to logs)
[mephisto_logs.txt](https://github.com/facebookresearch/Mephisto/files/5652330/mephisto_logs.txt)

1. (Line 79) Task available on Worker Sandbox
[...]

PR #347, opened 2020-12-30
Free unit when disconnected agent is in waiting

JackUrb:
# Overview
There was a not-so-great interaction between `Agent`s in `STATUS_WAITING` and disconnecting that allowed tasks to start with agents that had already disconnected. This PR resolves the issue.

As a fun bonus, this PR also resolves the issue where units that were freed (due to an agent disconnecting before a task started) would never be acceptable again due to a lock file not being properly cleared.

# Detailed rundown + fix
Much of the expected state updates for an `Agent` on disconnect were set up to handle the process of clearing and expiring all `Unit`s of an `Assignment` that `Agent`'s `Unit` was a part of. As this happened during a live `Assignment`, the `Agent` was expected to take an act, and we could catch the disconnect in the `TaskRunner`. When an `Agent` is in `STATUS_WAITING` however, it is not managed by a `TaskRunner` and as such if the worker disconnects, the agent is still attached to the `Unit`, but from the perspective of the `Supervisor` that unit has a valid agent attached and is ready for pairing. 

Considering that a `Unit` with an `Agent` in `STATUS_WAITING` hasn't done any work just yet, we are allowed to return that `Unit` to the pool for another `Worker` to claim. This means that it's appropriate in this special case to ask the `Agent` to just clear it's association with a `Unit` if it is in `STATUS_WAITING` during either a disconnect or return event.

# Testing
Can no longer replicate the issue in #327, so long as I wait long enough for the disconnect event to be registered by the `Supervisor` (max of 4 seconds).


pringshia (approved):
Awesome work on tracking down this race condition, and much credit to @vaibhavad with his stellar repro steps in #327.

vaibhavad:
Great fix @JackUrb!

PR #348, opened 2020-12-31
Consistently clean shutdown

JackUrb:
# Overview
Mephisto shutdown has been a little... finicky up until now. Sometimes when an assignment thread was still running, the shutdown would halt with no messaging, leading to some amount of frustration, likely a number of repeated Ctrl-C's, and more. Due to this, a wrong Ctrl-C could lead things to become in some incomplete states, like HITs remaining live, or a heroku server not shutdown (or a local thread running as a phantom in the background). This PR seeks to resolve the issue by improving the manual shutdown process and blocking off important shutdown parts.

This also includes a new `force_shutdown` method that will run through the motions that a human could on shutting down, but for use in tests to (mostly) ensure that all resources are cleaned between executions. (@EricMichaelSmith).

The end result is a system wherein live tasks will always be able to be shut down with one Ctrl-C (two for emergencies, potentially cutting off someone who is in-task), and test tasks can shutdown cleanly with one call.

# Implementation
On the safety side, I've added a signal wrapper around the expiration segment, to prevent people from trying to interrupt this stage. People having HITs up that are not expired is the most common issue in Mephisto, so this extra step should help prevent this from surfacing.

On the completion side, I found that incomplete Unit and Assignment threads were the biggest thing causing the system to hang during an incomplete shutdown (as the threads would suspend resources while waiting to close), so additional tracking has been added to the `TaskRunner` component to track this stuff down and end it. Much of that implementation surrounds adding a new shutdown-based error to Agents to ensure that they release their threads.

### DIFF
--- a/mephisto/abstractions/blueprint.py
+++ b/mephisto/abstractions/blueprint.py
@@ -18,6 +18,7 @@ from typing import (
     Iterable,
     AsyncIterator,
     Callable,
+    Tuple,
     TYPE_CHECKING,
 )
 
@@ -28,6 +29,7 @@ from mephisto.data_model.exceptions import (
     AgentReturnedError,
     AgentDisconnectedError,
     AgentTimeoutError,
+    AgentShutdownError,
 )
 from mephisto.data_model.constants.assignment_state import AssignmentState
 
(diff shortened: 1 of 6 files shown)


EricMichaelSmith (approved):
Seems reasonable to me `force_shutdown()` is pretty useful. Thanks!

PR #357, opened 2021-01-11
Changed typo in import for review module

sghmk12:
There was a typo in the imports for mephisto/client/cli.py from this previous pull request: https://github.com/facebookresearch/Mephisto/pull/295.
It appeared as an import error when running 'mephisto review' with a react-app. 
Previously it was written as 'mephisto.client.review_server' and it is now corrected to 'mephisto.client.review.review_server'.


pringshia (approved):
Thanks for the fix! LGTM

PR #360, opened 2021-01-12
Updated pre-commit-config to reference the latest version of the prettier library

sghmk12:
Prettier has migrated from 'https://github.com/prettier/prettier' to 'https://github.com/prettier/pre-commit' and the version attribute has changed in format.

Change is referenced in this issue on the Prettier repo: https://github.com/prettier/prettier/issues/9459

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,8 +15,8 @@ repos:
     hooks:
       - id: black
         language_version: python3.6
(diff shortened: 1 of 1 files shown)


pringshia (approved):
Thanks for the update here.

For future reference, this fix was based on this issue: https://github.com/prettier/prettier/issues/9459

PR #361, opened 2021-01-15
Integrating mephisto review with Mephisto DB

sghmk12:
Prior to this change users could only use ```mephisto review``` via standard input of a ```CSV``` or ```JSON``` file.
This allows users to alternatively provide no standard input and instead source the task information from their ```mephistoDB```.
Users can do this by adding a ```--db``` flag to their ```mephisto review``` command followed by the task name within the db which they wish to review.

Such a command can look like this:
```
mephisto review my-review/build -o results.csv --db task_name
```

### DIFF
--- a/mephisto/client/review/review_server.py
+++ b/mephisto/client/review/review_server.py
@@ -14,7 +14,16 @@ import time
 import threading
 
 
-def run(build_dir, port, output, csv_headers, json=False, debug=False):
+def run(
+    build_dir,
+    port,
+    output,
+    csv_headers,
+    json=False,
+    database=False,
+    database_task_name=None,
+    debug=False,
+):
     global index_file, app
     global ready_for_next, current_data, finished, index_file
     global counter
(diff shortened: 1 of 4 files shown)


mephisto/client/review/review_server.py
```diff
@@ -41,16 +42,39 @@ def json_reader(f):
         for jsonline in f:
             yield json.loads(jsonline)
 
+    def mephistoDBReader():
+        if database is None:
+            return []
```
pringshia, line 47: Perhaps we should print an error message or something here instead to denote that no task_name in the database was specified

pringshia: this also prob never gets run because Line 71 below only gets run if Line 70 is true and database is not None

pringshia: How about if no task_name is specified, we print out an error message, as well as a list of the current task_names that are stored in the db (perhaps sorted by desc updated date)? I think this should be a huge UX win

mephisto/client/review/review_server.py
```diff
@@ -41,16 +42,39 @@ def json_reader(f):
...
+
+        db = LocalMephistoDB()
+        mephisto_data_browser = MephistoDataBrowser(db=db)
+        units = mephisto_data_browser.get_units_for_task_name(database)
+
+        def format_data_for_review(data):
+            contents = data["data"]
+
+            inputs = contents["inputs"]
+            inputs_string = f"Character: {inputs['character_name']}, Description: {inputs['character_description']}"
```
pringshia, line 61: Where are `'character_name'` and `'character_description'` coming from here? Is this specific nto the `static_example` task? Not all tasks will have this data schema, we should generalize

mephisto/client/review/review_server.py
sghmk12: This is actually intentional as a safety mechanism for future use of the method however I can remove it as I realize it is a little messy. There is also no way to distinguish between the db flag being given no argument or the db flag not being called so I didn't include an error message

pringshia:
This looks good to me. Please update the PR description to have the latest usage (incl. task name w/ flag)

pringshia (approved):

PR #362, opened 2021-01-19
DB can get an object from own id

bottler:
I change some `find_` functions to add the id of the object as a parameter. This enables, say, easily getting a Unit object corresponding to unit 123.

I find this useful because I am implementing program logic in a separate process from the one which is actually running mephisto. So I might want to find all the completed agent ids in one call, and then accept certain ones in a separate call, but I haven't got an easy way to actually store the Agent objects between the calls. So I might as well recreate them from their ids in the second call.

### DIFF
--- a/mephisto/abstractions/database.py
+++ b/mephisto/abstractions/database.py
@@ -217,6 +217,7 @@ class MephistoDB(ABC):
     def find_task_runs(
         self,
         task_id: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


JackUrb:
If you already have the agent's id and the db, wouldn't you just reconstruct it with `Agent(db, id)`?

bottler:
You're right. I was confused and going round the long way.

PR #363, opened 2021-01-19
suggest -i in heroku login

bottler:
There has been a recent change to the heroku CLI which means that it recommends using a browser to log in. Some mephisto users are working on text-only connections to remote machines making this annoying. Adding `-i` on the heroku command line makes it behave the old way, so let's suggest it.


JackUrb (approved):
Great change 👍 . Will save many people some future headaches

PR #366, opened 2021-01-21
Removing improper agent state transitions

JackUrb:
# Overview
Thanks to @jxmsML, I was able to identify a few state transitions that were occurring that shouldn't have been happening. Each were problematic, and in total they meant that human-human chat tasks would freeze up after the first occurrence of two people leaving a chat at roughly the same time (within 2 mins of each other, the average time for MTurk to mark a return). 

This fix addresses the two following issues:
- The routing server would sometimes have stale information, and this information was getting sent back to the main mephisto server and putting the wrong state up. 
  - Now this is updated to only check for DISCONNECT changes, as this is the only status that the server will ever have that Mephisto needs to read, and anything else causes Mephisto to send a status update to sync the server.
- When a task was returned, the `MTurkUnit` that marked this return would move directly from `ASSIGNED` to `LAUNCHED`, as this is the MTurk HIT status that we use to note a return event in the first place.
  - We now defer the local transition from `ASSIGNED` back to `LAUNCHED`, as the `Blueprint` will decide shortly after the `RETURNED` status is set whether the `Unit` needs to be expired. If it does, the `Unit` will end up in an `EXPIRED` status, otherwise the `Blueprint` will call `clear_assigned_agent` to move the `Unit` from `ASSIGNED` back to `LAUNCHED`

## Testing
I was able to reproduce these state issues before this set of changes, but can no longer recreate them on this branch.

### DIFF
--- a/mephisto/data_model/unit.py
+++ b/mephisto/data_model/unit.py
@@ -58,9 +58,6 @@ class Unit(ABC):
         self.requester_id = row["requester_id"]
         self.worker_id = row["worker_id"]
 
-        # Flag for task runner to set on units used in assignments
-        self._is_concurrent: bool = False
-
         # Deferred loading of related entities
         self.__task: Optional["Task"] = None
         self.__task_run: Optional["TaskRun"] = None
--- a/mephisto/data_model/unit.py
+++ b/mephisto/data_model/unit.py
@@ -58,6 +58,9 @@ class Unit(ABC):
         self.requester_id = row["requester_id"]
         self.worker_id = row["worker_id"]
 
+        # Flag for task runner to set on units used in assignments
+        self._is_concurrent: bool = False
+
         # Deferred loading of related entities
         self.__task: Optional["Task"] = None
         self.__task_run: Optional["TaskRun"] = None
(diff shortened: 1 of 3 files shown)


mephisto/data_model/unit.py
```diff
@@ -181,7 +181,7 @@ def get_requester(self) -> "Requester":
 
     def clear_assigned_agent(self) -> None:
         """Clear the agent that is assigned to this unit"""
-        logger.debug(f"Clearing assigned agent {self.__agent} from {self}")
+        logger.debug(f"Clearing assigned agent {self.agent} from {self}")
```
jxmsML, line 184: nit: `MTurkUnit` does not have `agent`

jxmsML (approved):
LGTM! Thanks for fixing this!!! Only one nit to be fixed.

mephisto/data_model/unit.py
JackUrb: Ack! I fixed this locally and didn't push - wanted this to be `self.agent_id` because we actually stopped using `self.__agent`

PR #367, opened 2021-01-22
Improved review-ui template with CSS and HTML to look more visually engaging to users

sghmk12:
This minor redesign improves the following aspects of the review template to be more visually engaging to users:

- Uses bigger and easier to press color-coded buttons when reviewing:
    regular_review_text

- Review data that won't fit on screen becomes scrollable:
    scrolling_review_text

- Error text is now color-coded and 'Loading' and 'Done' notices are centered:
    loading_error

    done

### DIFF
--- a/packages/cra-template-mephisto-review/template/src/App.css
+++ b/packages/cra-template-mephisto-review/template/src/App.css
@@ -21,6 +21,10 @@
   overflow: auto;
 }
 
+.error {
+  color: #e3242b;
+}
+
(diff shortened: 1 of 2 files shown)


pringshia:
Nice work! Much more pleasing to the eye and I like the scroll for larger payloads.
Thanks for taking the initiative to improve part of the codebase here.

pringshia (approved):

PR #371, opened 2021-01-23
Endpoint for users to review all data at once in mephisto review

sghmk12:
This feature allows users to pull all data points from a review source at once to their front-end via our ```mephisto review``` server with a new endpoint with the URL ```/all_data```.

For example if the data source consisted the following ```data.csv``` file:
```
This is good text, row1
This is bad text, row2
```
The new endpoint will return the following response when queried:
response

To access this endpoint users must include a ```--all``` flag in their ```mephisto review``` command.
This can be in conjunction with the ```--db``` flag such as:
```
mephisto review my-review/build -o results.csv --db html-static-task-example --all
```
or in conjunction with standard input such as:
```
cat data.csv | mephisto review my-review/build -o results.csv --all
```

If the ```--all``` flag is not set the endpoint will return an error.
Conversely, If the ```--all``` flag is set and other endpoints for reviewing individual tasks are accessed they will also return error as a safety measure

Users can now query the endpoint ```/all_data``` for a list of all data points related to the task that they are currently reviewing

Users can also add parameters to the query such that they get a paginated result using the params ```page``` and ```results_per_page```

Such a query can look like this: ```/all_data?page=2&results_per_page=3```

If users do not specify a value for results per page the server will default to 10 results per page and if neither parameter is provided the server returns all results.

### DIFF
--- a/mephisto/client/review/review_server.py
+++ b/mephisto/client/review/review_server.py
@@ -85,6 +85,23 @@ def run(
             ready_for_next.wait()
         finished = True
 
+    def consume_all_data():
+
+        if json:
+            data_source = json_reader(iter(sys.stdin.readline, ""))
+        else:
+            data_source = csv.reader(iter(sys.stdin.readline, ""))
+
+        if csv_headers:
+            next(data_source)
+
+        data_point_list = []
+        index = 0
+        for row in data_source:
+            data_point_list[index] = row
+            index += 1
+        return data_point_list
+
     @app.route("/data_for_current_task")
     def data():
         global current_data, finished
@@ -97,6 +114,15 @@ def run(
 
         return jsonify(
             {"finished": finished, "data": current_data if not finished else None}
+        )
+
+    @app.route("/all_data_for_current_task")
+    def data():
+        data_point_list = consume_all_data()
+        return jsonify(
+            {"data_points": data_point_list, "length": data_point_list.length}
+            if data_point_list is not None
+            else {error: "No data points for current task"}
         )
 
     @app.route("/submit_current_task", methods=["GET", "POST"])
--- a/mephisto/client/review/review_server.py
+++ b/mephisto/client/review/review_server.py
@@ -89,19 +89,14 @@ def run(
             ready_for_next.wait()
         finished = True
 
-    def consume_all_data(page, results_per_page):
-        limit = (
-            results_per_page
-            if (
-                results_per_page is not None
-                and isinstance(results_per_page, int)
-                and results_per_page > 0
-            )
-            else RESULTS_PER_PAGE_DEFAULT
-        )
-        paginated = (
-            True if page is not None and isinstance(page, int) and page > 0 else False
-        )
+    # returns all data or page of all data given a limit where pages are 1 indexed
+    def consume_all_data(page, limit=RESULTS_PER_PAGE_DEFAULT):
+        paginated = type(page) is int
+        if paginated:
+            assert page > 0, "Page number should be a positive 1 indexed integer."
+            assert (
+                type(limit) is int and limit > 0
+            ), "results_per_page should be a positive integer"
 
         first_index = (page - 1) * limit if paginated else 0
         data_point_list = []
@@ -160,7 +155,7 @@ def run(
             {"finished": finished, "data": current_data if not finished else None}
         )
 
-    @app.route("/all_data_for_current_task")
+    @app.route("/all_data")
     def all_task_data():
         if not all_data and database_task_name is None:
             return jsonify(
@@ -170,7 +165,11 @@ def run(
             )
         page = request.args.get("page", default=None, type=int)
         results_per_page = request.args.get("results_per_page", default=None, type=int)
-        data_point_list = consume_all_data(page, results_per_page)
+        try:
+            data_point_list = consume_all_data(page, results_per_page)
+        except AssertionError as ae:
+            print(f"Error: {ae.args[0]}")
+            return jsonify({"error": ae.args[0]})
         return jsonify({"data": data_point_list, "length": len(data_point_list)})
 
     @app.route("/submit_current_task", methods=["GET", "POST"])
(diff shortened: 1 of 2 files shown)


mephisto/client/review/review_server.py
```diff
@@ -99,10 +160,28 @@ def data():
             {"finished": finished, "data": current_data if not finished else None}
         )
 
+    @app.route("/all_data_for_current_task")
```
pringshia, line 163: I would just call this endpoint `/all_data`. The reason the other endpoint is called `/data_for_current_tasks` is bc it only works for a single task. having both all and current task in the endpoint name is sort of an oxymoron/contradictory.

mephisto/client/review/review_server.py
```diff
@@ -85,10 +89,67 @@ def consume_data():
...
+    def consume_all_data(page, results_per_page):
+        limit = (
+            results_per_page
+            if (
+                results_per_page is not None
+                and isinstance(results_per_page, int)
+                and results_per_page > 0
+            )
+            else RESULTS_PER_PAGE_DEFAULT
+        )
```
pringshia, line 101: it might be cleaner to replace these Lines 92-101 here with this, using a default parameter value & assertion, e.g.:

```python
def consume_all_data(page, results_per_page = RESULTS_PER_PAGE_DEFAULT):
    assert type(results_per_page) is int, "results_per_page should be an int"
```

mephisto/client/review/review_server.py
```diff
@@ -85,10 +89,67 @@ def consume_data():
...
+        paginated = (
+            True if page is not None and isinstance(page, int) and page > 0 else False
```
pringshia, line 103: you don't need to do a None check if you're also doing isinstance, you can just do the isinstance, since it will also cover the None clause (a value of None will always make isinstance false, i.e. `isinstance(None, int) == false`)

that is to say you could just shorted all of the above with:
```python
paginated = type(page) is int and page > 0
```

mephisto/client/review/review_server.py
```diff
@@ -85,10 +89,67 @@ def consume_data():
             ready_for_next.wait()
         finished = True
 
+    def consume_all_data(page, results_per_page):
```
pringshia, line 92: is page 0-indexed or 1-indexed? a comment for this method may be useful

mephisto/client/review/review_server.py
```diff
@@ -85,10 +89,62 @@ def consume_data():
             ready_for_next.wait()
         finished = True
 
+    # returns all data or page of all data given a limit where pages are 1 indexed
```
pringshia, line 92: You may want to write this comment as a [Python docstring](https://www.programiz.com/python-programming/docstrings) instead, such that IDEs can pick up and surface documentation for it.

pringshia:
Overall this looks good! One minor comment about using a docstring. Aside from that, approved on my end.

@JackUrb since this PR is quite Python-heavy do you mind taking a quick scan? I tried my best on any Python feedback I could give.

@sghmk12 for this one since it's Python-heavy, let's also wait for an approval from Jack before merging in

pringshia (approved):

JackUrb (changes requested):
Sorry about the late review! Some comments inline, but the basic overview is that, while this method will _work_, on larger review loads it's going to be slow. In short, calls to the `get_units_for_task_name` and `get_data_from_unit` methods are going to be a bottleneck, and should optimally be minimized.

The first major change is to treat the `MephistoDataBrowser` data in the same way as local data, populating the `all_data_list` on the first initialization.

Then, in order to pull the strongest benefits from pulling from the DB directly, you can implement one of two ways to refresh the `all_data_list` from the database, first defining a `fill_from_db` method or similar, and then calling it either periodically from a thread, or after enough time has passed since the last call during an `all_task_data` request.

Sorry about the late context here, but great work otherwise!

mephisto/client/review/review_server.py
```diff
@@ -137,8 +215,23 @@ def after_request(response):
         response.headers.add("Cache-Control", "no-store")
         return response
 
-    thread = threading.Thread(target=consume_data)
-    thread.start()
+    if all_data and database_task_name is None:
```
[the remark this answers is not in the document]
sghmk12: Thanks for the input, definitely much needed advice. Just implemented these features. Please let me know if more is needed or if there are any python errors in my latest commits!

mephisto/client/review/review_server.py
```diff
@@ -137,8 +210,26 @@ def after_request(response):
...
+            data_source = json_reader(iter(sys.stdin.readline, ""))
+        else:
+            data_source = csv.reader(iter(sys.stdin.readline, ""))
+            if csv_headers:
+                next(data_source)
+
+        all_data_list = []
+
+        for row in data_source:
+            all_data_list.append(row)
```
pringshia, line 227: maybe I'm missing something, but the above 3 lines here could just be `all_data_list = list(data_source)` no?

sghmk12: You are correct, I am still getting accustomed to advanced python syntax haha. Will change

mephisto/client/review/review_server.py
```diff
@@ -102,39 +103,30 @@ def consume_all_data(page, limit=RESULTS_PER_PAGE_DEFAULT):
...
-                        break
-                    data_point_list.append(data)
-            else:
-                for row in data_source:
-                    data_point_list.append(row)
+            # If differnce in time since the last update to the data list is over 5 minutes, update list again
+            # This can only be done for usage with mephistoDB as standard input is exhausted when originally creating the list
+            global datalist_update_time
+            now = datetime.now()
+            if (now - datalist_update_time).total_seconds() / 60.0 > 5:
```
JackUrb, line 110: This (5) should be moved to a constant near the top of the file for easy editing. Generally it's good to avoid magic numbers (could also move both by making the constant based on a second timeout).

JackUrb (approved):
Great changes - sorry for the delay! Just one nit but otherwise this is very much good to go.

mephisto/client/review/review_server.py
pringshia: It would also be good to have a boolean constant flag to toggle this functionality on/off altogether

sghmk12: Added both for my next PR. My bad I didn't have this in the current version.

Issue #372, opened 2021-01-23
DoneButton disappear after episode done


jxmsML:
I've received some emails saying that turkers are unable to click Done with this HIT button in parlai chat. cmd: 
``` python ~jingxu23/ParlAI/parlai_internal/mturk/tasks/reverse_persona/parlai_chat_task_demo/parlai_test_script.py ```
The DoneButton would flash for just a couple of seconds right after `episode_done() = True` and then disappear.
It appears that it might be related to this pr https://github.com/facebookresearch/Mephisto/pull/366, where

- when world episode_done, and `_mark_agent_done(agent_info)` is called https://github.com/facebookresearch/Mephisto/blob/master/mephisto/operations/supervisor.py#L773 (the frontend would mark agent status as 'complete' with nonempty `done_text`, thus the `DoneButton` is displayed), but before the local db_status is mark as completed: https://github.com/facebookresearch/Mephisto/blob/master/mephisto/operations/supervisor.py#L374,  agent local db status is still 'in task'.
- and there is a potential mismatch between remote status ('completed') and local db_status('in task'): https://github.com/facebookresearch/Mephisto/pull/366/files#diff-8682adcf9f382b8b56166423e545fbea94a9487afc4ad18d67fb5c2a7cc7a18aR802,
which after the pr is landed,  the remote status would be reset to the same as db_status ('in task') again during `_handle_updated_agent_status` , which leads to the case where in frontend `agentState.task_done` changes from `true` to `false` and thus `DoneButton` disappears.

Above is only my hypothesis, not sure if it's the actual cause. My temporary way of fixing it is modifying https://github.com/facebookresearch/Mephisto/blob/master/mephisto/operations/supervisor.py#L801-L806 to below:
```
            if status != db_status:
                if status == AgentState.STATUS_COMPLETED:
                    continue
                if status != AgentState.STATUS_DISCONNECT:
                    # Stale or reconnect, send a status update
                    self._send_status_update(self.agents[agent_id])
                    continue  # Only DISCONNECT can be marked remotely, rest are mismatch
                agent.update_status(status)
```
which would render the DoneButton again which I can also click on. 
Not sure if it's correct tho.

JackUrb:
[...]

PR #373, opened 2021-01-25
Donot send status update if frontend agent status is complete

jxmsML:
fixed https://github.com/facebookresearch/Mephisto/issues/372 (see problem description and rational for this fix)


JackUrb (approved):
Thanks for figuring this out and getting a fix together so soon!

PR #374, opened 2021-01-27
Add a Dockerfile to encapsulate Mephisto in a container + add matrix version testing workflow

pringshia:
## Overview

Adds a `Dockerfile` and a `.dockerignore` file to the repo.

The default base image being used is [nikolaik/python-nodejs](https://hub.docker.com/r/nikolaik/python-nodejs), with the tags for Python 3.9 and Node.JS 15.

One advantage of having this Dockerfile and a containerized setup is that it should allow us to quickly alter the python and node versions and check if Mephisto works across different python/node version numbers. We can conceivably add a matrix of these tests into a periodic testing workflow. [**I've set up a workflow already here**](https://github.com/facebookresearch/Mephisto/actions/runs/514175332). (I think we can have a better assertion at the end, or maybe `mephisto check` as it is is fine).

Another advantage I imagine is that folks can use the Dockerfile to create repro situations for bug reports and tests.

### Setup the image
Note: Using the `.dockerignore` file, I was able to exclude unneeded files from the build context to minimize the the size to around 12.9MB for faster builds. Mainly involved ignore `node_modules`, `.git` folder, and cache folders.

```console
$ docker build -t mephisto .
...
Successfully tagged mephisto:latest
```

### Run Mephisto from within the container

```console
$ docker run mephisto
Mephisto seems to be set up correctly.
```

```console
$ docker run mephisto bash -c 'mephisto requesters'

  requester_id  provider_type    requester_name    registered
--------------  ---------------  ----------------  ------------
             1  mock             MOCK_REQUESTER    False
```

Once #375 is resolved, we can also run tasks directly out of the container, bound to a port on the local machine:

```console
$ docker run -p 3000:3000 mephisto bash -c 'cd /mephisto/examples/simple_static_task && python static_test_script.py'
...
Listening on 3000
...
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+**/node_modules
+**/.git
+**/.mypy_cache
+**/build/*
+data/*
+**/*.mp4
--- a/.github/workflows/docker-testing-matrix.yml
+++ b/.github/workflows/docker-testing-matrix.yml
@@ -14,7 +14,7 @@ on:
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
   # This workflow contains a single job called "build"
-  mephisto-check:
+  check:
     # The type of runner that the job will run on
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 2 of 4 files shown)


JackUrb (approved):
This is awesome! Love getting more consistency and coverage in our testing.

PR #383, opened 2021-01-28
New CLI command: config & some doc updates

pringshia:
# Usage

`mephisto config [IDENTIFIER] [VALUE]`

### Print out the current full config

```bash
$ mephisto config
/Users/tikir/.mephisto/config.yml:

core:
  main_data_directory: /Users/tikir/mephisto/data
```

### Update a key
```bash
$ mephisto config core.main_data_directory /Users/tikir/mephisto/new_data_folder
core.main_data_directory succesfully updated to: /Users/tikir/mephisto/new_data_folder
```

### Read a key
```bash
$ mephisto config core.main_data_directory
/Users/tikir/mephisto/new_data_folder
```

### Validation
```bash
$ mephisto config main_data_directory
Usage: mephisto config [OPTIONS] [IDENTIFIER] [VALUE]

Error: Invalid value: Identifier must be of format: <section>.<key>
You passed in: main_data_directory
```

---

Also updated the README to include setting the data directory config via this CLI tool (I think this makes the quickstart tutorial flow a bit better). Moved the Docker setup info to a separate docker readme where I'll add more Docker recipes in the future.

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -1,0 +1,15 @@
+### Docker installation (new!)
+```bash
+# Build the docker image and tag with name 'mephisto'
+$ docker build -t mephisto . 
+
+# By default, the container just runs `mephisto check`:
+$ docker run mephisto
+Mephisto seems to be set up correctly.
+
+# You can also bind ports and pass in shell commands, e.g. to run
+# a task directly from the container
+$ docker run -p 3000:3000 mephisto bash -c 'cd mephisto/examples/simple_static_task && python static_test_script.py
+
+```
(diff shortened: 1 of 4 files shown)


pringshia:
Hmm, that's correct... there's no easy way to handle those cases. Maybe we just provide this as a helper for string values, while still treating modifying the .yml file directly as the golden case.

I looked up the docs for `git config` as that is somewhat related for inspiration. It seems like they offer some basic support, and call these sorts of cases for lists, called "multivars". I'm not sure I see anything for structured objects.

https://docs.oracle.com/cd/E88353_01/html/E37839/git-config-1.html

PR #384, opened 2021-01-28
[WIP] Print balance (or other arbitrary information) at the start of mTurk runs

pringshia:
- Adds an abstract method `print_banner_on_start(self, task_run)` to the `CrowdProvider` class.
- Implements it for the `MTurkProvider` (and via inheritance also `MTurkSandboxProvider`) to print out balance information at the start of a run.

```console
$ python static_test_script.py mephisto/architect=heroku mephisto.provider.requester_name=mturk
...
[2021-01-28 00:22:43,290][sh.command][INFO] - <Command '/bin/rm -rf /Users/tikir/mephisto/data/data/runs/NO_PROJECT/361/build/heroku_server_html-static-task-example_361', pid 40240>: process completed
[2021-01-28 00:22:43,323][botocore.credentials][INFO] - Found credentials in shared credentials file: ~/.aws/credentials
============================
Balance: 32644.93
============================

[2021-01-28 00:22:43,762][mephisto.operations.supervisor][INFO] - Sending alive
[2021-01-28 00:22:43,763][mephisto.operations.supervisor][INFO] - Sending alive
```

### DIFF
--- a/mephisto/abstractions/crowd_provider.py
+++ b/mephisto/abstractions/crowd_provider.py
@@ -123,6 +123,10 @@ class CrowdProvider(ABC):
         raise NotImplementedError()
 
     @abstractmethod
+    def print_banner_on_start(self) -> str:
+        return ""
+
+    @abstractmethod
     def cleanup_resources_from_task_run(
         self, task_run: "TaskRun", server_url: str
     ) -> None:
--- a/mephisto/operations/operator.py
+++ b/mephisto/operations/operator.py
@@ -245,6 +245,7 @@ class Operator:
             provider.setup_resources_for_task_run(
                 task_run, run_config, shared_state, task_url
             )
+            print(provider.print_banner_on_start(task_run))
 
             initialization_data_array = blueprint.get_initialization_data()
 
--- a/mephisto/abstractions/providers/mock/mock_provider.py
+++ b/mephisto/abstractions/providers/mock/mock_provider.py
@@ -73,6 +73,9 @@ class MockProvider(CrowdProvider):
         """Mocks don't do any initialization"""
         return None
 
+    def print_banner_on_start(self):
+        return f"============================\nStarting mock provider\n============================\n"
+
     def cleanup_resources_from_task_run(
         self, task_run: "TaskRun", server_url: str
     ) -> None:
--- a/mephisto/abstractions/providers/mturk/mturk_provider.py
+++ b/mephisto/abstractions/providers/mturk/mturk_provider.py
@@ -20,6 +20,7 @@ from mephisto.abstractions.providers.mturk.mturk_utils import (
     setup_sns_topic,
     delete_sns_topic,
     delete_qualification,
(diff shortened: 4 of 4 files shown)


mephisto/operations/operator.py
```diff
@@ -245,6 +245,7 @@ def validate_and_run_config_or_die(
             provider.setup_resources_for_task_run(
                 task_run, run_config, shared_state, task_url
             )
+            print(provider.print_banner_on_start(task_run))
```
pringshia, line 248: should this use the logger instead?

mephisto/abstractions/crowd_provider.py
```diff
@@ -122,6 +122,10 @@ def setup_resources_for_task_run(
         """
         raise NotImplementedError()
 
+    @abstractmethod
+    def print_banner_on_start(self) -> str:
+        return ""
```
JackUrb, line 127: Nit: if this isn't printing itself, it shouldn't have a name in the format `print_*`. `get_start_banner()` or similar could work.

That being said, I think it actually makes more sense for this method to do the printing, as it can also be responsible for confirm steps, like:
```
You are about to launch 50,000 tasks, for an estimated total of $1,000,000, are you sure you want to do that...?
```

Unfortunately, if we're adding confirm steps, it's also possible that we want to skip the confirmations entirely, so that's something to account for.

mephisto/operations/operator.py
JackUrb, line 248: At this point it's likely a little too late for a person to really use this kind of configuration information to make a decision (or potentially change the launch). Realistically you'd probably want to be displaying this just around the `assert_task_args` section.

JackUrb (changes requested):
I really like taking steps to print arbitrary but useful run information at the start of a job, but it's important to note (what I think is) one of the most important parts of printing it: being able to review (and potentially stop a job) when the data comes out. Once the architect is up, you're only moments away from things being live, and this would be too late.

Some thoughts on the design: in order to get the info up sooner in the run, you can rely directly on `run_config` and use an abstract class method, rather than just an abstract method. This would have the signature:
```
@classmethod
@abstractmethod
def print_banner_on_start(self, run_config: DictConfig) -> None:
```

Of course, encapsulation comes with some issues, as in order to print the correct information that someone may want, information may actually need to come from multiple sources. Thankfully the only case I can think of for this _for now_ is budgeting related: In order to know how much something costs, you need both `Blueprint` and `CrowdProvider` information. Only the `Blueprint` knows how many units it expects to launch, and unfortunately this value is not easily accessed just from the `run_config`. This would lead `Blueprint`s to have to expose something like an abstract classmethod of `get_expected_unit_count()` or something? I'm not fully sure yet. 

Perhaps more of a discussion is needed before there's a clear right path here though.

pringshia:
This is great feedback, thanks @JackUrb. I most likely won't have the bandwidth to work on this, so this is up for grabs for whoever wants it.

PR #385, opened 2021-02-02
Fixing race condition for MTurk qualification creation

JackUrb:
# Overview
A race condition exists if creating a new qualification for the first time live, when no worker has been assigned the qualification before. This became evident in @mojtaba-komeili's run, wherein a qualification is assigned the moment a worker accepts a task for the first time. Multiple workers accepting the task all at once would lead Mephisto to think it needed to create multiple qualifications. This was never surfaced before, as most qualification flows would grant at the end of onboarding or the task, and that made it much less likely the first few workers to get the qualification would get it at the exact same time.

# Implementation
If the insert fails due to a uniqueness break, we now treat it as a no-op, and warn the user if there's any strong differences between the inserted qual and the one that's actually stored. As we expect that Mephisto users will be assigning by the given `qualification_name` and not using the mturk qualification ids directly (in that we don't even expose these without asking for them), it's likely that this is the desired behavior under all circumstances.

# Testing
Added a test for idempotency, as well as other things for qualification mappings.

### DIFF
--- a/mephisto/abstractions/providers/mturk/mturk_datastore.py
+++ b/mephisto/abstractions/providers/mturk/mturk_datastore.py
@@ -258,7 +258,9 @@ class MTurkDatastore:
     ) -> None:
         """
         Create a mapping between mephisto qualification name and mturk
-        qualification details in the local datastore
+        qualification details in the local datastore.
+
+        Repeat entries with the same `qualification_name` will be idempotent
         """
         try:
             with self.table_access_condition, self._get_connection() as conn:
@@ -280,6 +282,7 @@ class MTurkDatastore:
                 return None
         except sqlite3.IntegrityError as e:
             if is_unique_failure(e):
+                # Ignore attempt to add another mapping for an existing key
                 qual = self.get_qualification_mapping(qualification_name)
                 logger.debug(
                     f"Multiple mturk mapping creations for qualification {qualification_name}. "
--- a/mephisto/abstractions/providers/mturk/mturk_datastore.py
+++ b/mephisto/abstractions/providers/mturk/mturk_datastore.py
@@ -14,8 +14,13 @@ from datetime import datetime
 
 from botocore.exceptions import ClientError
 from botocore.exceptions import ProfileNotFound
+from mephisto.abstractions.databases.local_database import is_unique_failure
 
 from typing import Dict, Any, Optional
+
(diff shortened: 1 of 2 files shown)


mephisto/abstractions/providers/mturk/mturk_datastore.py
```diff
@@ -255,23 +260,47 @@ def create_qualification_mapping(
...
+                    ) VALUES (?, ?, ?, ?);""",
+                    (
+                        qualification_name,
+                        requester_id,
+                        mturk_qualification_name,
+                        mturk_qualification_id,
+                    ),
+                )
+                return None
+        except sqlite3.IntegrityError as e:
```
mojtaba-komeili, line 281: nit: looks good but may add a quick comment that this is to address the race condition

test/abstractions/providers/mturk/test_mturk.py
```diff
@@ -60,6 +60,68 @@ def test_create_and_find_worker(self) -> None:
         failed_worker = MTurkWorker.get_from_mturk_worker_id(db, "FAKE_ID")
         self.assertIsNone(failed_worker, f"Found worker {failed_worker} from a fake id")
 
+    def test_mturk_qual_create_mapping(self) -> None:
```
mojtaba-komeili, line 63: Does this create enough number of request for race condition to be really tested?

mojtaba-komeili (approved):
Looks good and thanks for fixing it so quickly.

test/abstractions/providers/mturk/test_mturk.py
JackUrb: All that needs to happen to test is that `create_qualification_mapping` is called twice with the same `qualification_name`

mojtaba-komeili: Yeah, but if the first one get resolve fast enough, then it doesn't happen.

JackUrb: Right, the creation call doesn't do any checking, so the race occurs the moment that the system decides it will need to create a qualification. Here to be specific:
https://github.com/facebookresearch/Mephisto/blob/ce88a722729f0c6c974dcf188d9ecca82ad0e296/mephisto/abstractions/providers/mturk/mturk_worker.py#L90-L96

PR #389, opened 2021-02-04
Fix conditional hook execution in mephisto-review-hook

pringshia:
`useEffect` always gets called now, but short-circuits in mock case. This is to prevent a variable amount of hooks from being executed across renders, which React does not like.

https://reactjs.org/docs/hooks-rules.html


JackUrb (approved):
Always happy to upgrade to best practices, as eventually those I find become breaking.

PR #390, opened 2021-02-04
Try making python requirement less specific

pringshia:
Trying to fix: https://github.com/facebookresearch/Mephisto/runs/1834054533

Using solution from: https://github.com/pre-commit/pre-commit/issues/1375

Testing: Checks for this branch pass


JackUrb (approved):
Sorry for the late review here! Only failure is due to a flakey test, so this looks good to me

PR #391, opened 2021-02-04
Overhaul of review template and review hook to support all data review endpoint.

sghmk12:
Mephisto review template is now redesigned using BlueprintJS to support reviewing all data with a grid mode.
The mephisto review hook is now also updated to support this change in the mephisto review template.
Both the hook and the template are now updated to version ```2.0.0```

Reviewing data using the all flag such as the following command:
```
mephisto review new-review/build -o results.csv --db html-static-task-example --all
```
will produce the following UI at location ```/``` with all data organized in pages:
All_data_screen
The data in the cards is scrollable.

Clicking on a card within the grid will route users to location ```/:id``` and allow users to review the item:
All_data_item_review

Reviewing data in the standard way without the all flag with a command such as:
```
mephisto review new-review/build -o results.csv --db html-static-task-example
```
will produce the following UI where the user is continually routed from location ```/``` to location ```/:id``` based on the item currently being reviewed:
Single_review

The ID property in both review modes merely references the index of the data point in the list of data given to mephisto review.
Reviews sent from the review template now contain a property called ```id``` to identify which item was reviewed for usage of mephisto review in all mode, as items can be reviewed out of order.

The original mephistoReviewHook can be accessed by importing ```useMephistoReviewLegacy``` from ```mepshitoReviewHook```

The review server has been given one new endpoint named ```/data/<id>```.
Sending a GET request to this endpoint returns the data currently being reviewed in standard mode or a single piece of data from the list of all data when in all mode.
A response to a GET request could look like:
```
{
    "data": {
        "data": "my review content",
        "id": 0
    }
    "mode": "ALL"
}
```
Sending a POST request to this endpoint with review data now allows the user to review a piece of data.
The json body of a POST request to ```/data/<id>``` could look like:
```
{
    "result": "approved,
    "id": 0
}
```
To which a response to the POST request could look like:
```
{
    "result": "SUCCESS",
    "finished": false
    "mode": "ALL"
}
```

The endpoint ```/all_data``` from the review server has been renamed to ```/data```.
This endpoint now returns the id of the task being currently reviewed if mephisto review is in standard mode and behaves normally if mephisto review is in all mode.
A response from requesting ```/data``` in all mode now looks like:
```
{
    "data": [
        {
            "data": "my review content",
            "id": 0
        },
        {
            "data": "more review content",
            "id": 1
        }
    ],
    "mode": "ALL",
    "total_pages": 1
}
```
where total_pages is the total number of pages that the app can request at the given number of results per page in the request.
A response from requesting ```/data``` in standard (one-by-one) mode now looks like:
```
{
    "data": {
        "data" : "my review content"
        "id": 0
    },
    "mode": "OBO",
}
```

Both ```/data``` and ```/data/<id>``` also return a mode state representing whether the app is in standard or all mode. Standard mode is referenced as ```one-by-one``` or ```OBO``` mode in the server.


packages/mephisto-review-hook/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mephisto-review-hook",
-  "version": "1.0.3",
+  "version": "2.0.0",
```
pringshia, line 3: I think it's better to go with 1.1.0 here instead of 2.0.0

packages/cra-template-mephisto-review/template/src/GridRenderer.jsx
```diff
@@ -0,0 +1,59 @@
...
+              <Pagination
+                totalPages={totalPages}
+                page={page}
+                setPage={setPage}
+              />
+            </>
+          ) : (
+            <h5 className="grid-renderer-header error">
+              Sorry, no data available. Please provide Mephisto Review with some
+              data
```
pringshia, line 50: It would be good to make this message more helpful and provide users instructions on how to provide the data.

packages/cra-template-mephisto-review/template/src/GridView.jsx
```diff
@@ -0,0 +1,30 @@
...
+
+  return data && data.length > 0 ? (
+    <div className="grid-container">
+      {data.map((item) => (
+        <Card
+          key={item.id}
+          interactive={true}
+          elevation={Elevation.TWO}
+          className="grid-item"
+          onClick={() => history.push(`/${item.id}`)}
```
pringshia, line 16: any specific reason for using `history.push` here as opposed to wrapping the Card in a `<Link />` tag? Just curious

packages/cra-template-mephisto-review/template/src/components/Pagination.jsx
```diff
@@ -0,0 +1,81 @@
+import React from "react";
+import { Button, Icon, Intent } from "@blueprintjs/core";
+import "./components.css";
+
+function Pagination({ page = 1, totalPages = 1, setPage = () => {} }) {
+  const MAX_CELLS = 7;
+  const CELL_MID_LEN = ~~(MAX_CELLS / 2);
```
pringshia, line 7: pretty cool! didn't know `~~` was a substitute for Math.floor for positive numbers

packages/cra-template-mephisto-review/template/src/components/Pagination.jsx
```diff
@@ -0,0 +1,81 @@
...
+      for (let i = 1; i < MAX_CELLS - 5; i++) {
+        pages[CELL_MID_LEN + i] = { number: page + i };
+        pages[CELL_MID_LEN - i] = { number: page - i };
+      }
+    }
+  } else {
+    for (let i = 0; i < totalPages; i++) {
+      pages[i] = { number: i + 1, ellipsis: false };
+    }
+  }
```
pringshia, line 41: Might be good to add a comment to explain what this code does and what it's intentions are, maybe with some sample outputs, I'm think this logic could also be encapsulated in a utility method

packages/mephisto-review-hook/package.json
sghmk12: I used 2.0.0 to indicate a breaking change. Will 1.1 have the same effect?

packages/cra-template-mephisto-review/template/src/GridRenderer.jsx
sghmk12: Will do!

packages/cra-template-mephisto-review/template/src/GridView.jsx
sghmk12: I forgot that was an available component, I will change that

packages/cra-template-mephisto-review/template/src/components/Pagination.jsx
sghmk12: Saw it on a github thread while writing this haha!

packages/cra-template-mephisto-review/template/src/components/Pagination.jsx
sghmk12: Will definitely try that out!

packages/cra-template-mephisto-review/template/src/components/Pagination.jsx
JackUrb: Little things like this are cool, but should probably be commented 😉

mephisto/client/review/review_server.py
```diff
@@ -147,36 +159,89 @@ def data():
...
+                    {
+                        "finished": finished,
+                        "data": current_data if not finished else None,
+                        "mode": MODE,
+                    }
+                )
+        else:
+            review = request.get_json(force=True)
+            if output == "":
+                sys.stdout.write("{}\n".format(review))
```
JackUrb, line 197: Is there a specific reason why you're directly writing with sys rather than print or some other kind of logging.

JackUrb (commented):
Python contents here look good to me. One thought is that the usage of `:id` should probably be documented clearly, as there are _a lot_ of ids in Mephisto, and when associated with data I jump to the actual `agent_id` that the data comes from, even when this doesn't correspond (even when using the Database directly here!).

As this file extends it would also be good to get function level doc-blocks for what these endpoints are trying to do.

mephisto/client/review/review_server.py
sghmk12: I just used the method originally implemented in this app to write out reviews. I can change the method for both this instance and other instances if there is a more efficient way.

packages/cra-template-mephisto-review/template/src/components/Pagination.jsx
```diff
@@ -0,0 +1,81 @@
...
+
+  pages.forEach((p) => {
+    if (p.number === page) p.active = true;
+  });
+
+  const leftArrow = page > 1;
+  const rightArrow = page < totalPages;
+
+  return (
+    <div className="pt-button-group pagination">
```
pringshia, line 51: I think you want bp3-button-group here instead of pt-*. The buttons in your screenshot still have their border radius between them, which they shouldn't by using button group (only at the outer edges)

packages/mephisto-review-hook/src/index.js
```diff
@@ -1,54 +1,64 @@
...
   useEffect(() => {
-    if (isMock) {
-      return;
-    }
-
-    fetch("/data_for_current_task")
-      .catch((err) => {
-        setError({ type: "DATA_RETRIEVAL", error: err });
-      })
+    if (isMock) return;
```
pringshia, line 19: So hooks can't have early returns. See #389 for how I fixed this. You'll want to incorporate those changes here, otherwise it will revert the behavior.

packages/cra-template-mephisto-review/template/src/components/Pagination.jsx
pringshia: agreed, most folks wouldn't be able to decipher this. I'd either comment or revert to the more familiar Math.floor

packages/cra-template-mephisto-review/template/src/GridView.jsx
```diff
@@ -0,0 +1,30 @@
...
+        >
+          <div className="grid-item-text">
+            <pre>{item.data}</pre>
```
pringshia, line 19: Looking at the screenshot you provided, I see that the whole payload is wrapped in a string " ". Can you look into ways to avoid those quotes? e.g. I think the JSON itself should be embedded, not the string representation of the JSON

packages/mephisto-review-hook/package.json
pringshia: hmm, ok let's keep 2.0.0

mephisto/client/review/review_server.py
JackUrb: @pringshia was there a reason you were using `sys.stdout` particularly? Otherwise we can likely just use `print`.

packages/mephisto-review-hook/src/index.js
pringshia: whoops, i misread this - my bad!

packages/cra-template-mephisto-review/template/src/ItemRenderer.jsx
```diff
@@ -0,0 +1,96 @@
...
+
+  const reviewError = (error) => {
+    AppToaster.show({ message: `ERROR: ${error}` });
+  };
+
+  return (
+    <main className="Item-Renderer">
+      {mode === "ALL" ? (
+        <Button
+          onClick={() => history.push("/")}
```
pringshia, line 36: I think this can also be wrapped in a Link tag. Generally, using Link tags are better for browser accessibility. You also get to see the destination URL when hovering over the tag

packages/mephisto-review-hook/src/index.js
pringshia, line 19: did you run the pre-commit command on all files?

packages/cra-template-mephisto-review/template/src/ItemRenderer.jsx
```diff
@@ -0,0 +1,96 @@
...
+      {error && <h5 className="error">Error: {JSON.stringify(error)}</h5>}
+      {isLoading ? (
+        <h1>Loading...</h1>
+      ) : isFinished ? (
+        <h1>Done reviewing! You can close this app now</h1>
+      ) : (
+        <>
+          <H2>Please review the following data:</H2>
+          <Card className="item" elevation={Elevation.TWO}>
+            <pre>{(data && data.data) || data}</pre>
```
pringshia, line 54: you don't need `|| data` anymore

packages/mephisto-review-hook/src/index.js
sghmk12: Yep, it doesn't seem to have a quarrel with this, however, I will wrap it around brackets for consistency

packages/cra-template-mephisto-review/template/src/ItemRenderer.jsx
sghmk12: Tested changing it to data && data.data and it breaks when the only valid object is data. However I am considering changing the endpoint on review_server.py to always return a data.data and data.id in the data portion of the response. Thoughts?

pringshia: ah you're right, thanks for testing this. && returns the first falsey value, so if data.data is null, it would evaluate to null

if we were to use || instead, then the first truthy value would be returned, which is also not quite what we want because data would always be true.

If data is guaranteed to always be defined, then we could do `data.data || data`, but idk if that assumption is true.

changing the endpoint could also be good. it would def simplify the usage a bit here

mephisto/client/review/review_server.py
pringshia: no particular reason @JackUrb aside from the fact that since originally this was demoed as a unix-esque CLI use-case of piping back out to shell, i was trying to be symmetrical to reading in from `sys.stdin`. After doing some research, I see that there's no other reason between either usage, except that `print` adds a newline where as `sys.stdout.write` doesn't. Since we're adding on a new line at the end anyway, it seems fine to use print to me.

sghmk12: Sounds good, I'll make the changes

packages/cra-template-mephisto-review/template/src/ItemRenderer.jsx
sghmk12: I opted to change the endpoints for now such that data.data is guaranteed for most responses to improve consistency throughout the app. For now that means there's no use for the OR clause however maybe we'll change that in the future.

mephisto/client/review/review_server.py
```diff
@@ -189,8 +189,7 @@ def next_task():
         )
 
         if output == "":
-            sys.stdout.write("{}\n".format(result))
-            sys.stdout.flush()
+            print("{}\n".format(result))
```
pringshia, line 192: You don't need the "\n" anymore since print auto-appends. To keep the functionality the same, let's remove that

mephisto/client/review/review_server.py
```diff
@@ -246,8 +245,7 @@ def task_data_by_id(id):
         else:
             review = request.get_json(force=True)
             if output == "":
-                sys.stdout.write("{}\n".format(review))
-                sys.stdout.flush()
+                print("{}\n".format(review))
```
pringshia, line 248: Same, re: "\n"

sghmk12: Sure!

packages/cra-template-mephisto-review/template/src/components/Pagination.jsx
```diff
@@ -0,0 +1,120 @@
...
+
+  return {
+    pages,
+    leftArrow,
+    rightArrow,
+  };
+}
+
+function Pagination({ page = 1, totalPages = 1, setPage = () => {} }) {
+  const { pages, leftArrow, rightArrow } = getPagination(page, totalPages);
```
pringshia, line 87: minor style note, for booleans I prefer names like `isLeftArrowEnabled`, makes for easier to read code

pringshia (approved):
Woohoo! Nice work getting this together. Excited to start building off of this. It's great that we were able to build this out in a way to also expose the legacy hook for back-compat.

pringshia:
One more thing now that I noticed it, the POST request to `/data/<id>` seems a little redundant in that the body also requires an id? Shouldn't the id just be retrievable from the URL? To eliminate the body altogether we could also just simply do `POST /data/<id>/approve` which I've seen often across various APIs in the wild

sghmk12:
> One more thing now that I noticed it, the POST request to `/data/<id>` seems a little redundant in that the body also requires an id. Shouldn't the id just be retrievable from the URL? To eliminate the body altogether we could also just simply do something like `POST /data/<id>/approve` and `POST /data/<id>/reject` which I've seen often across various APIs in the wild

I was thinking we might want to leave the review open to what the front-end writes in the case that the user wanted custom review formats. The ID from the URL I will definitely change, your point makes a lot more sense.

sghmk12:
> Woohoo! Nice work getting this together. Excited to start building off of this. It's great that we were able to build this out in a way to also expose the legacy hook for back-compat.

Same, super stoked about this!

packages/cra-template-mephisto-review/template/src/components/Pagination.jsx
sghmk12: Added!

pringshia:
Looks good to me! Happy with leaving the payload for future consideration

PR #392, opened 2021-02-10
Mephisto-specific templated variables in HTML Static

JackUrb:
# Overview
In order to give Static HTML templates a bit more functionality that React-based tasks already get from the `mephisto-task` package, I added a few template variables that HTML tasks can use to populate the `provider_worker_id` and the `mephisto_agent_id` into the template, in case they need to log this somewhere or use it in a script.

# Implementation
Created an object mapping from the template name to the actual variable. Added a more extensive doc about the usage and implementation for the `StaticHTMLBlueprint` in general.

# Testing
Added one of these templated variables to the example.

### DIFF
--- a/mephisto/abstractions/blueprints/static_html_task/source/dev/app.jsx
+++ b/mephisto/abstractions/blueprints/static_html_task/source/dev/app.jsx
@@ -36,7 +36,15 @@ function MainApp() {
     handleSubmit,
     isOnboarding,
     agentId,
+    providerWorkerId,
   } = useMephistoTask();
+
+  // We expose these as template variables, even if they aren't
+  // directly asked for
+  const MEPHISTO_KEY_MAP = {
+    mephisto_agent_id: agentId,
+    provider_worker_id: providerWorkerId,
+  };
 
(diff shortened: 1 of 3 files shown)


Rebecca-Qian (approved):
Thanks for adding this!

mephisto/abstractions/blueprints/static_html_task/source/dev/app.jsx
```diff
@@ -126,10 +142,16 @@ function ShowURL({ url, data = null }) {
...
+    <HtmlRenderer
+      html={retrievedHtml}
+      data={data}
+      mephisto_keys={mephisto_keys}
+    />
+  );
 }
 
-function HtmlRenderer({ html, data }) {
+function HtmlRenderer({ html, data, mephisto_keys }) {
```
pringshia, line 154: overall looks good, I'd just recommend using the JS camelCase, e.g. `mephistoKeys`.

also, thought on naming it as  `mephistoVars` instead? i feel that keys is a little data-structure/implementation specific, whereas vars denotes that they are values set via mephisto

PR #395, opened 2021-02-12
Form submit on Enter

mojtaba-komeili:
Trying to solve problem on disconnect when pressing Enter on form responses.


packages/bootstrap-chat/src/FormResponse.jsx
```diff
@@ -98,6 +98,14 @@ class FormResponse extends React.Component {
...
                   });
                 }}
+                onKeyPress={(e) => {
+                  console.log(e);
+                  if (e.key === "Enter") {
+                    tryMessageSend();
+                    e.stopPropagation();
+                    e.nativeEvent.stopImmediatePropagation();
+                  }
+                }}
```
pringshia: So I believe this onKeyPress handler should be added instead to L118 (L126 after this addition).

```jsx
<FormControl
    type="text"
    style={{ fontSize: "16px" }}
```

packages/bootstrap-chat/src/FormResponse.jsx
```diff
@@ -129,6 +129,14 @@ class FormResponse extends React.Component {
...
             />
           </Col>
+          onKeyPress={(e) => {
+            console.log(e);
+            if (e.key === "Enter") {
+              tryMessageSend();
+              e.stopPropagation();
+              e.nativeEvent.stopImmediatePropagation();
+            }
+          }}
```
pringshia, line 139: This prop should be on the FormControl component, right now it's dangling not attached to any component

packages/bootstrap-chat/src/FormResponse.jsx
```diff
@@ -115,6 +115,14 @@ class FormResponse extends React.Component {
             {question}
           </Col>
           <Col sm={5}>
+            onKeyPress={(e) => {
```
pringshia, line 118: This is 1) on the Col component now (it should be a prop on the FormControl component) and 2) it should be within the > tag of the component. Right now onKeyPress is after the ending > tag of the Col component

packages/bootstrap-chat/src/FormResponse.jsx
```diff
@@ -127,6 +127,14 @@ class FormResponse extends React.Component {
                   return { responses: new_res };
                 });
               }}
+              onKeyPress={(e) => {
+                console.log("Pressed");
+                if (e.key === "Enter") {
+                  tryMessageSend();
```
pringshia, line 133: the message is still being sent on Enter with this call, remove

pringshia (approved):
Added e.preventDefault() and tested locally. FormResponse doesn't auto-submit anymore.

PR #397, opened 2021-02-16
[FIX] Review server and Review template only serving and rendering string data

sghmk12:
Up until now due to recent changes the mephisto review template would only render string data.
This was an issue because when using the --json flag, the server would serve raw json data in its responses and the review app would not render this data.
This issue was not present for csv and mephistoDB data which were served as strings.

Now the review template renders raw json data. As an added bonus mephistoDB data is now served as raw JSON instead of as a string of JSON.


pringshia (approved):
lgtm

PR #398, opened 2021-02-17
Flask-Based router, enabled on LocalArchitect

JackUrb:
## Overview
In order to get custom Flask-based Mephisto Routers running up and about, this PR takes the first steps of creating a Mephisto Router Flask Blueprint, which can be used by a Flask App to deploy Mephisto-Routing on the backend.

## Implementation
Almost a direct translation of our existing node-based server. I'm sure that the underlying logic of the `MephistoRouter` can probably be improved, or better encapsulated, but so long as the overall design is correct that can be resolved in the future without the end-user needing to change anything (as they'll still just import the class).

The big details here are making sure that all `MephistoRouter`s share the same state (as every new request can initialize a new router, according to how Flask works), and that `monkey.patch_all()` is called in any `app.py` that relies on the `MephistoRouter` to ensure that the websocket requests and HTTP requests can interleave properly.

This PR also includes some changes to the `build_router.py` file, as well as the `LocalArchitect`, that make it possible to select a deploy to the Flask-based server rather than the node-based one we used to have.

**Note**
Absent from this PR are the changes that will:
- let this work on a `HerokuArchitect`
- let the `build_router.py` file pull in an alternate build (for the case of someone wanting to deploy something that extends the `MephistoRouter`)
- actually allow you to configure the ports for this
- Provide some missing documentation
All of these will be addressed in a follow-up PR once this is merged.

## Testing
Launched all of our example tasks with flask as the deploy server, like:
```
> python parlai_test_script.py mephisto.architect.server_type=flask
```
Screen Shot 2021-02-17 at 5 21 13 PM
Screen Shot 2021-02-17 at 5 21 43 PM

Screen Shot 2021-02-17 at 2 10 52 PM

## Discussion
One important note is that the `mephisto_flask_blueprint.py` file _does not_ rely on the Mephisto module as of now, leading to the case where there is some duplication of constants in this file that are available elsewhere. As of now, I've done this to prevent installing Mephisto from being a requirement on the routing server, as the routing server should be as lightweight as possible. I'm open to changing this to require the routing server to have Mephisto installed as well, but the alternate would be to write a script that can sync these constants from a master file to the other locations, mark it as generated, and create a test that ensures they remain synced. Unsure if that's overkill though, and perhaps we should stay as-is until we find ourselves duplicating _a lot_ of constants.

### DIFF
--- a/mephisto/abstractions/architects/local_architect.py
+++ b/mephisto/abstractions/architects/local_architect.py
@@ -43,6 +43,9 @@ class LocalArchitectArgs(ArchitectArgs):
         default="localhost", metadata={"help": "Addressible location of the server"}
     )
     port: str = field(default="3000", metadata={"help": "Port to launch the server on"})
+    server_type: str = field(
+        default="npm", metadata={"Help": "Type of server to run, node or Flask"}
+    )
 
 
 @register_mephisto_abstraction()
@@ -76,6 +79,7 @@ class LocalArchitect(Architect):
         self.hostname: Optional[str] = args.architect.hostname
         self.port: Optional[str] = args.architect.port
         self.cleanup_called = False
+        self.server_type = args.architect.server_type
 
     def _get_socket_urls(self) -> List[str]:
         """Return the path to the local server socket"""
@@ -129,7 +133,7 @@ class LocalArchitect(Architect):
 
     def prepare(self) -> str:
         """Mark the preparation call"""
-        self.server_dir = build_router(self.build_dir, self.task_run)
+        self.server_dir = build_router(self.build_dir, self.task_run, self.server_type)
         return self.server_dir
 
     def deploy(self) -> str:
@@ -143,11 +147,18 @@ class LocalArchitect(Architect):
 
         return_dir = os.getcwd()
         os.chdir(self.running_dir)
-        self.server_process = subprocess.Popen(
-            ["node", "server.js"],
-            preexec_fn=os.setpgrp,
-            env=dict(os.environ, PORT=f"{self.port}"),
-        )
+        if self.server_type == "npm":
+            self.server_process = subprocess.Popen(
+                ["node", "server.js"],
+                preexec_fn=os.setpgrp,
+                env=dict(os.environ, PORT=f"{self.port}"),
+            )
+        elif self.server_type == "flask":
+            self.server_process = subprocess.Popen(
+                ["python", "app.py"],
+                preexec_fn=os.setpgrp,
+                env=dict(os.environ, PORT=f"{self.port}"),
+            )
         self.server_process_pid = self.server_process.pid
         os.chdir(return_dir)
 
--- a/mephisto/abstractions/architects/local_architect.py
+++ b/mephisto/abstractions/architects/local_architect.py
@@ -44,7 +44,7 @@ class LocalArchitectArgs(ArchitectArgs):
     )
     port: str = field(default="3000", metadata={"help": "Port to launch the server on"})
     server_type: str = field(
-        default="npm", metadata={"Help": "Type of server to run, node or Flask"}
+        default="node", metadata={"Help": "Type of server to run, `node` or `flask`"}
     )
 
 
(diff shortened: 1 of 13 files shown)


mephisto/abstractions/architects/local_architect.py
```diff
@@ -43,6 +43,9 @@ class LocalArchitectArgs(ArchitectArgs):
         default="localhost", metadata={"help": "Addressible location of the server"}
     )
     port: str = field(default="3000", metadata={"help": "Port to launch the server on"})
+    server_type: str = field(
+        default="npm", metadata={"Help": "Type of server to run, node or Flask"}
```
pringshia, line 47: perhaps the default should be `"node"` here, as `npm` solely refers to the package manager

mephisto/abstractions/architects/local_architect.py
```diff
@@ -143,11 +147,18 @@ def deploy(self) -> str:
 
         return_dir = os.getcwd()
         os.chdir(self.running_dir)
-        self.server_process = subprocess.Popen(
-            ["node", "server.js"],
-            preexec_fn=os.setpgrp,
-            env=dict(os.environ, PORT=f"{self.port}"),
-        )
+        if self.server_type == "npm":
```
pringshia, line 150: then would also have to change this to node

mephisto/abstractions/architects/router/build_router.py
```diff
@@ -41,7 +42,7 @@ def install_router_files() -> None:
     Create a new build including the node_modules
     """
     return_dir = os.getcwd()
-    os.chdir(SERVER_SOURCE_ROOT)
+    os.chdir(NPM_SERVER_SOURCE_ROOT)
```
pringshia, line 45: NODE here makes more sense as well

mephisto/abstractions/blueprints/parlai_chat/parlai_chat_task_runner.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, *args, **kwargs):
 if TYPE_CHECKING:
     from mephisto.data_model.task_run import TaskRun
     from mephisto.abstractions.blueprint import AgentState
-    from mephsito.data_model.assignment import Assignment
+    from mephisto.data_model.assignment import Assignment
```
pringshia, line 50: oh wow, this typo is quite common: https://github.com/facebookresearch/Mephisto/search?q=mephsito

pringshia (commented):
Overall this look good.

My main piece of feedback is in quite a few cases, `node` makes more sense than `npm`, as `npm` refers to the package manager itself, not to node which is the server technology... it's akin to calling the flask server the pip server or something.

we could also call it based on the framework, e.g. express, to parallel it with flask, but I think node is more familiar to people

mephisto/abstractions/architects/local_architect.py
Rebecca-Qian: agree!

mephisto/abstractions/architects/router/build_router.py
```diff
@@ -18,7 +18,8 @@
     from mephisto.data_model.task_run import TaskRun
 
 ROUTER_ROOT_DIR = os.path.dirname(router_module.__file__)
-SERVER_SOURCE_ROOT = os.path.join(ROUTER_ROOT_DIR, "deploy")
+NPM_SERVER_SOURCE_ROOT = os.path.join(ROUTER_ROOT_DIR, "deploy")
```
Rebecca-Qian, line 21: Nit: `NODE` instead of NPM here as well for consistency

mephisto/abstractions/architects/router/flask/app.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+from flask import Flask
+
+flask_app = Flask(__name__)
+flask_app.register_blueprint(mephisto_router, url_prefix=r"/")
+
+if __name__ == "__main__":
+    WebSocketServer(
+        ("", 3000),
```
Rebecca-Qian, line 26: Assuming this will be configurable in the future?

Rebecca-Qian (approved):
Thanks for making these changes @JackUrb ! 

Looks good, agree with @pringshia 's nit about using `node` instead of `npm` when referring to the server.

Looking forward to the follow up PR(s).

mephisto/abstractions/blueprints/parlai_chat/parlai_chat_task_runner.py
JackUrb: All in my type checking code, which I haven't run in a while. I should add mypy signal!

pringshia (approved):

Issue #399, opened 2021-02-19
Improve integration with hydra 


ytsheng:
Repro step: `python run_mturk.py` in https://github.com/facebookresearch/dynabench/blob/master/annotators/run_mturk.py with the below change.

ideal setting:
<img width="433" alt="Screen Shot 2021-02-19 at 10 19 04 AM" src="https://user-images.githubusercontent.com/1545487/108559851-f9212e80-729f-11eb-8525-ccb89e0d9948.png">

revised/less ideal setting:

<img width="445" alt="Screen Shot 2021-02-19 at 10 49 34 AM" src="https://user-images.githubusercontent.com/1545487/108559996-2b329080-72a0-11eb-883f-04091eb41fc4.png">
<img width="296" alt="Screen Shot 2021-02-19 at 10 49 49 AM" src="https://user-images.githubusercontent.com/1545487/108560037-3b4a7000-72a0-11eb-93a5-4aad8dd2d373.png">

We should not be using json.dumps in `build_router.py`; Hydra recommends using pickle.

JackUrb:
> We should not be using json.dumps in `build_router.py`; Hydra recommends using pickle.

Are you referring to this line:
https://github.com/facebookresearch/Mephisto/blob/1df548c0827864ef83ac6a1ffb77d2886d7ac93d/mephisto/abstractions/architects/router/build_router.py#L83

We have to have the dump there be `json` compatible as anything in that is being directly packaged up and delivered to the frontend `taskConfig` variable, accessible from the `useMephistoTask` hook. You can trace what's being added by the `get_frontend_args` to see this isn't the source of the issue.

Though I do know where the usage of `json` comes from though: 
https://github.com/facebookresearch/Mephisto/blob/5852aecc7647d4f1655c09de3698396dea0e49dd/mephisto/operations/operator.py#L188
Which we use so that we can later load the same setup from the db:
https://github.com/facebookresearch/Mephisto/blob/1df548c0827864ef83ac6a1ffb77d2886d7ac93d/mephisto/data_model/task_run.py#L53

The key design goal here though is that we want to be able to _store_ the exact arguments that were used _in the database_, for which a pickle isn't quite appropriate. Is there an alternate for being able to create the static yaml representation?


Also it would be useful to see the context on "hydra recommends pickle", there may be more there that's useful.

omry:
Context on pickle recommendation:

It depends on what you are trying to achieve.
[...]

PR #400, opened 2021-02-22
Updating usage to to_yaml to fix some structured config bugs

JackUrb:
# Details
In trying to store configurations into the `MephistoDB` for future use, we had been using `OmegaConf.to_container` which seems to not preserve all of the behaviors that are ideal for nested structured configs (in reconstruction). This PR changes to use `to_yaml`. This change should not effect previously saved runs, as both the results of `to_container` and `to_yaml` are valid parameters for the later `OmegaConf.create()` call.

Following discussion in #399, this also removes the `resolve` portion of writing out the configuration, as downstream tasks may want the interpolation and there's no real risk for us including them. We retain storing the yaml rather than using pickles (and getting type safety) as configuration signatures for specific tasks may change over time, and we want the old ones to be properly loadable/examinable.

@ytsheng please see if this resolves the issue you were having in your setup.

# Testing
Unit testing

### DIFF
--- a/mephisto/operations/operator.py
+++ b/mephisto/operations/operator.py
@@ -185,7 +185,7 @@ class Operator:
         new_run_id = self.db.new_task_run(
             task_id,
             requester_id,
-            json.dumps(OmegaConf.to_container(run_config, resolve=True)),
+            json.dumps(OmegaConf.to_yaml(run_config)),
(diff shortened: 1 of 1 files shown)


pringshia (approved):
seems fine to me 👍

PR #402, opened 2021-02-23
Fixing documentation issue in static react docs

JackUrb:
This guidance is old, and should be updated.


pringshia (approved):
thanks for the update

PR #405, opened 2021-03-01
Adding max_num_concurrent_units hydra config arg

JackUrb:
# Overview
The concurrent `Unit` launch cap has been available in-code for a while, but with no way to turn it on other than manually changing the `TaskLauncher` file. Turns out it's pretty simple to just bubble this argument up, even if this isn't the final implementation. There shouldn't be any future changes to how a user toggles this option, even if it turns out we change the way that it's transmitted behind the scenes.

cc @EricMichaelSmith as I know a lot of ParlAI users are using this manually at the moment.
cc @ytsheng as I know this is used in Dynabench as well.

# Testing
Ran a job limiting to 1 unit at a time, asserted that if I opened a second there was nothing to take. Completed the job normally one at a time
```
python static_test_script.py mephisto.task.max_num_concurrent_units=1
```

### DIFF
--- a/mephisto/data_model/task_config.py
+++ b/mephisto/data_model/task_config.py
@@ -80,6 +80,15 @@ class TaskConfigArgs:
             )
         },
     )
+    max_num_concurrent_units: int = field(
+        default=0,
+        metadata={
+            "help": (
+                "Maximum units that will be released simultaneously, setting a limit "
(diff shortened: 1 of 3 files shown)


EricMichaelSmith (approved):
Yeah this is great - thanks for adding!

PR #411, opened 2021-03-04
Adding config arguments to heroku architect

JackUrb:
# Overview
In order to set configuration arguments for Heroku, there needs to be a way to specify them to be pushed from the CLI. This PR adds an argument for setting these to the structured config, and then parses it and writes it out. Implementation is pretty straightforward.

# Testing:
I haven't tested extensively. @snyxan please try this out for your case!

Ran
```
print(dict(args.architect.heroku_config_args), type(dict(args.architect.heroku_config_args)))
```
With the contents I added to the readme, and this works.


snyxan:
Thanks for this PR!

On high level LGTM. To make it work e2e we need some minor changes. Please see this PR: #413

PR #412, opened 2021-03-05
Redesigned review UI to use navbar for major UI components

sghmk12:
### Summary:
Redesigned the review template UI to use a navigation bar to contain major UI components such that there is more room to display data.

### Changes per route:
- **AllItemView ("/"):**
    - Search bar and "Mephisto Review" logo have been moved into a navigation bar at the top of the screen
    - The itemListRenderer div has increased in size to fill all remaining room on viewport unconsumed by the navbar
    - Pagination only appears for review data with more than one page of items
- **ItemView ("/:id"):**
    - The review helper text as well as the Approve, Reject and Back buttons have been moved into a navigation bar at the top of the screen
    - The itemRenderer div has been increased in size to fill all remaining room on viewport unconsumed by the navbar
    - The itemRenderer div is no longer displayed inside a card by default. Users must create their own back-drops for their review renderer content. This is because the itemRenderer now takes up most of the viewport area

### Extras:
- **File name and location refactoring:**
    - itemRenderer files for the ItemView route are referred to as ItemViewRenderers in their file name. Originally, a single item renderer component could be used as props for both the AllItemView and the ItemView routes. Now there must be a different type of component for each route. This is because the removal of the card back-drop in the itemRenderer div of the ItemView route requires itemRenderers for ItemView to have an additional back-drop design not used in the AllItemView. This new naming scheme differentiates itemRenderers for the ItemView from those for the AllItemView
    - prop names for the ItemView route remain itemRenderer
    - itemRenderer files for the AllItemView route are still referred to as ItemRenderers
    - prop names for the AllItemView route remain itemListRenderer and itemRenderer

### Alternatives:
- **Untested naming scheme alternatives:**
    - Change name of itemRenderer prop for ItemView to itemViewRenderer to reduce confusion
- **Tested naming scheme alternatives:**
    - Remove card style containers for the default itemListRenderer in the AllItemView and use the same itemRenderer component for both ItemView and AllItemView
        - Results: itemRenderer components for ItemView need different CSS styling than those in AllItemView for visual appeal

### Testing:
- Tested using CSV data:
    - Tested in "ALL" mode
        - Tested with multiple pages of Data
        - Tested with search
    - Tested in "OBO" mode
- Tested using JSONL data:
    - Tested in "ALL" mode
        - Tested with multiple pages of Data
        - Tested with search
    - Tested in "OBO" mode

### Visual Preview:

https://user-images.githubusercontent.com/13241633/110155839-33b4bc00-7db4-11eb-9a28-392b8b2f66e0.mov


packages/cra-template-mephisto-review/template/src/AllItemView.jsx
```diff
@@ -1,16 +1,24 @@
...
+  NavbarGroup,
+  NavbarDivider,
+  NavbarHeading,
+  Alignment,
+} from "@blueprintjs/core";
 import { Tooltip2 } from "@blueprintjs/popover2";
-import DefaultItemRenderer from "./components/DefaultItemRenderer";
-import GridView from "./components/GridView";
+import DefaultItemRenderer from "./plugins/DefaultItemRenderer";
+import DefaultItemListRenderer from "./plugins/DefaultItemListRenderer";
```
pringshia, line 15: I really like this idea of a plugins folder

packages/cra-template-mephisto-review/template/src/components/WordCloud.jsx
```diff
@@ -198,7 +198,7 @@ function WordCloud({
 
   //map css styles to spans with corresponding word and populate paragraph with spans
   return (
-    <div className="cloud">
+    <div className="word-cloud">
```
pringshia, line 201: Shouldn't this WordCloud component be in the plugins folder?

packages/cra-template-mephisto-review/template/src/components/css/WordCloud.css
```diff
@@ -1,9 +1,9 @@
-.cloud {
+.word-cloud {
   text-align: center;
   justify-content: center;
 }
 
-.cloud p {
+.word-cloud p {
```
pringshia, line 6: This css file should prob also go in the plugins folder?

packages/cra-template-mephisto-review/template/src/plugins/css/WordCloudItemViewRenderer.css
```diff
@@ -0,0 +1,6 @@
+.word-cloud-item-view {
+  margin: auto 12px;
+  max-width: 75vw;
+  max-height: 75vh;
+  overflow: auto;
+}
```
pringshia, line 6: why do word clouds need two separate css files? any reason all word-cloud specific styles can't just be in one file?

pringshia:
Loving how organized the PR description is! 😍

packages/cra-template-mephisto-review/template/src/components/WordCloud.jsx
sghmk12: This part just renders the div full of spans for the actual word cloud itself (can be resized inside other divs), it plugs into two wrapper components for both the word cloud for the ItemView and the AllItemView

packages/cra-template-mephisto-review/template/src/components/css/WordCloud.css
sghmk12: Same as the above comment, this css file is for the unit word cloud component

packages/cra-template-mephisto-review/template/src/plugins/css/WordCloudItemViewRenderer.css
sghmk12: The itemRenderer for the big ItemView requires different dimensions compared to the miniature one for the AllItemView. This file has a maxWidth of 75vw whereas for the smaller itemView there is no max width necessary as the renderer is placed in the grid view.

pringshia: why can't the max-width just be part of the original css?

then, if the component has space to take up room it does so, otherwise if it's placed in a css grid, it also handles that. trying to understand if we actually need a special case here or we can just have the css be versatile

packages/cra-template-mephisto-review/template/src/components/WordCloud.jsx
pringshia: Could you point me to the two wrapper components?

I was thinking that all wordcloud stuff would go into the plugins folder (preferrably in it's own folder). that way if someone else in the future wants to add a new plugin, they could just follow the precedent and drop a folder into the plugins directory and use it. if we have the files scattered, it makes the plugins less modular and plug-and-play.

we can discuss this architecture further. otherwise this PR looks pretty awesome.

pringshia: ideally, if the code is all in one place, it's more packageable, say in the future if we decide to move it into an npm package. e.g: `npm install mephisto-plugin-wordcloud` or something.

packages/cra-template-mephisto-review/template/src/plugins/css/WordCloudItemViewRenderer.css
sghmk12: We could, however, the other file for the smaller word cloud doesn't need any CSS as all of its bounding is in the grid view. I don't see the application of adding this file to the smaller word cloud. I could make the name of this CSS file more generic for readability, however, I think it makes more sense to match its name to the only jsx file it is used in

packages/cra-template-mephisto-review/template/src/components/WordCloud.jsx
sghmk12: I totally agree, I was testing moving the css, etc. into their own folders for simplicity, but decided to hold off to keep the PR smaller. I think I will restructure the folders for this PR to align to this idea after we clear up the CSS issues.

packages/cra-template-mephisto-review/template/src/plugins/css/WordCloudItemViewRenderer.css
sghmk12: After reading your other comments, I understand better what you're saying in this discussion. In reference to my earlier comment, I was assuming the pure word cloud component ("/components/wordCloud") was staying separate from the two renderers ("/plugins/WordCloudItemRenderer", "/plugins/WordCloudItemViewRenderer") that implement it. Now I see you wish to merge this CSS file ("/plugins/css/WordCloudItemViewRenderer") with that of the pure word cloud ("/components/css/WordCloud"). I am still considering whether that would be the best choice for the WordCloudItemViewRenderer component as the css class above only applies to one div inside the component and no other elements in any other components.

packages/cra-template-mephisto-review/template/src/components/WordCloud.jsx
sghmk12: The wrapper components are in /plugin/WordCloudItemRenderer and /plugins/WordCloudItemViewRenderer

sghmk12:
@pringshia  just pushed new changes to simplify entire project file structure as we discussed and updated PR description accordingly.

PR #413, opened 2021-03-07
Specify heroku app name while setting config args, allow customized app name

snyxan:
# Overview

@JackUrb 

1. While setting heroku config args, it's required to specify the app name or it will fail. So added '-a APP_NAME' option to heroku cli command.
2. Setting heroku app name is different than setting other config vars. So something like 'heroku config:set APP_NAME=XXX' won't work. So I added `heroku_app_name` field to `HerokuArchitectArgs` and use that field to set `self.__heroku_app_name`. In this way, the heroku app will be created using the name we provided.

This PR is a complement to #411.

# Testing

Tested e2e and it passed.


JackUrb (approved):
Thanks for working this out - I actually did something almost exactly the same for the heroku app name arg but forgot to push!

Issue #417, opened 2021-03-16
[Bug report] heroku server not created by mephisto will be deleted when interruption happens


snyxan:
## Type of Issue

Select the type of issue:
- [x] Bug report (to report a bug)
- [ ] Feature request (to request an additional feature)
- [ ] Tracker (I am just using this as a tracker)
- [ ] Refactor request
- [ ] Documentation Ask

## Description

When heroku app already exists in user's account, mephisto will notify that the app name already exists and terminates properly. However, if an interruption(like a KeyboardInterrupt trigger by ctrl+c) happens before code getting here: https://github.com/facebookresearch/Mephisto/blob/master/mephisto/abstractions/architects/heroku_architect.py#L368. It will trigger cleaning up which deletes the heroku app. This is detrimental because we are not supposed to delete app not created by us.

This is probably part of a more general issue here: https://github.com/facebookresearch/Mephisto/blob/master/mephisto/abstractions/architects/mock_architect.py#L302.
Like all state(app created, deployed, deleted) should be tracked in parent class?

PR #418, opened 2021-03-16
Added creation state track in heroku architect

snyxan:
# Description

This PR aims at resolving this issue: #417 .

What it does is to add a creation state tracker to heroku architect. This tracker is only set to be `True` when mephisto actually creates this heroku app. When the heroku architect is shut down, it will delete the heroku app only if creation state is True.

## Type of change

Please check the options that are relevant.

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] Proposes a change (non-breaking change that isn't necessarily a bug)
- [ ] Refactor
- [ ] New feature (non-breaking change that adds a new functionality)
- [ ] Breaking change (fix or feature that would break some existing functionality downstream)
- [ ] This is a unit test
- [ ] Documentation only change

## Type of requested review

- [x] I want a thorough review of the implementation.
- [x] I want a high level review. 
- [x] I want a deep design review.

# Before and After

**Before**

When an interrupt happens, mephisto will delete the `test-mephisto` app even it didn't create it.

Screen Shot 2021-03-15 at 5 12 01 PM

**After**

When an interrupt happens, mephisto won't delete the `test-mephisto` app because it didn't create it.

Screen Shot 2021-03-15 at 5 13 06 PM

# Testing

Tested and passed.

### DIFF
--- a/mephisto/abstractions/architects/heroku_architect.py
+++ b/mephisto/abstractions/architects/heroku_architect.py
@@ -491,5 +491,5 @@ class HerokuArchitect(Architect):
         Shut down the server launched by this Architect, as stored
         in the db.
         """
-        if self.created: # only delete the server if it's created by us
+        if self.created:  # only delete the server if it's created by us
             self.__delete_heroku_server()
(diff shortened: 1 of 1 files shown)


JackUrb (approved):
Overall I like this design. I think we should probably be pausing after the `subprocess.CalledProcessError` to confirm with the user that they still want to launch, but there's a bigger question of actually collecting the correct error type and displaying it there that needs to be answered.

As such I think this is good to go, thanks for the fix!

PR #420, opened 2021-03-19
[BETA] Annotation UI Toolkit - Foundational Skeleton

pringshia:
# Overview

This PR introduces the foundational skeleton for a new annotation UI toolkit.
Additionally, this PR implements a proof-of-concept task to illustrate functionality for the Ego4D VQA use-case.

### Notable Features:
- Concise **task specification** for end-users in the spirit of **self-service**
- **Decoupled** layer architecture in the spirit of **code reuse**
- **Async message passing** between layers in the spirit of **flexibility and customizability**

---

## Task Specification

Here's what an example task looks like.

```jsx
function Layers() {
  return (
    <>
      <Layer
        name="Video"
        icon="video"
        component={VideoPlayer}
        alwaysOn={true}
      />
      <Layer name="Query 1">
        <Layer
          name="Item Crop"
          icon="widget"
          component={BBoxFrame}
          noPointerEvents={true}
        />
        <Layer name="Response Track" icon="path-search" />
        <Layer name="Query Frame" icon="help" />
      </Layer>
      <Layer name="Query 2">
        <Layer name="Item Crop" icon="widget" />
        <Layer name="Response Track" icon="path-search" />
        <Layer name="Query Frame" icon="help" />
      </Layer>
      <Layer name="Query 3">
        <Layer name="Item Crop" icon="widget" />
        <Layer name="Response Track" icon="path-search" />
        <Layer name="Query Frame" icon="help" />
      </Layer>
    </>
  );
}
```

The goal was to make task specifications composable and grokkable. By looking at the above, you can easily reason about what the task may be. We encourage code usability, as seen by using predefined layers for showing a `BBoxFrame` and a `VideoPlayer`.

The above the specification then produces something like this within the UI:

Screen Shot 2021-03-19 at 4 26 05 PM

## Decoupled Layer Architecture

By using global state, we allow layers to communicate with each other easily, irrespective of where they may be located within the component tree.

Of note is a custom hook called `useStore()`. It can be used from anywhere within the app &mdash; especially amongst layers, as such:

```jsx
const { set, get } = useStore()

console.log(get('debug.testValue')); // undefined
set('debug.testValue', 5);
console.log(get('debug.testValue')) // 5
```

This gets especially interesting with layers. All layers self-register themselves when described in the Task Spec above, and then have dominion over a scoped portion of the global state. You can query the state *for* arbitrary layers *from* arbitrary layers. You can even use this feature to pass messages to layers (show in the below section).

For example, the Video layer described above is addressable as such:
```jsx
get(["layers", "Video", "data", "playedSeconds"]) // note that we can also use an array path as opposed to dot-notation. This basically uses lodash's `get` method under-the-hood.
```

Given this, all layers can easily "subscribe" to the state of the VideoPlayer's position. The BBoxFrame, for example, subscribes to this value to know whether or not to show a frame-specific Bounding Box. 

## [Experimental] Message passing

Also by having global state, layers can easily "invoke" methods on each other. This is sort of like a remote procedural call.

For example, to instruct the VideoPlayer to seek to a certain time, one can easily do the following:

```jsx

  const { sendRequest } = useStore()
  sendRequest("Video", { type: "seek", payload: 4.0 });

  // which is syntactic sugar for....


  const { invoke } = useStore()
  invoke("layers.Video.data.requests", prev => [
    ...prev,
    { type: "seek", payload: value },
  ]);
```

The `invoke` helper allows you to immutably update a portion of the state, think of a `get` and a `set` combined. Here we push to the Video layer's request queue an action to seek to a different point in the video.

The Video layer, then consumes this as such:
```jsx
  const requestQueue = get("layers.Video.data.requests");
  useEffect(() => {
    if (!requestQueue || requestQueue.length === 0) return;
    requestQueue.forEach((request) => {
      process(request);
    });
    set(path("requests"), []);
  }, [requestQueue]);

  const process = useCallback(
    (req) => {
      if (req.type === "seek") {
        vidRef.current.seekTo(req.payload, "seconds");
      }
    },
    [vidRef.current]
  );
```

TODO: In the future, the above will be condensed via helpers to make process queues a more "batteries-included" experience.

One can imagine a hook that replaces all of the above with just:

```jsx
const seekTo = React.useCallback(time => vidRef.current.seekTo(time, "seconds"), [vidRef.current]); 
useProcessQueue('Video', {
    "seek": seekTo
}) 
```

## Demo

With layer support, global state, and message passing, we can create some pretty complex task annotation and configuration workflows pretty easily.

Here is a working example at the moment of a simple case, that just shows a bounding box on a layer for a video playing in a different layer. As this task evolves, more of the benefits of the above three features will be evident.

https://user-images.githubusercontent.com/425059/111843995-8d4ed780-88d8-11eb-9ea4-fdb8c7a406a9.mp4

## Future Work

- [ ] Helpers for task process queue
- [ ] include the ability to provide `onLoad()` information as part of the task specification so that initialData can be gathered
- [ ] Continue to further generalize BBox and VideoPlayer (lift the current hardcoded data in BBox to global state)
- [ ] Create better abstractions for specifying Layer actions


packages/annotation-ui/src/panels/ContextPanel.js
```diff
@@ -0,0 +1,40 @@
...
+  const { set, get, invoke } = useContext(Context);
+
+  const videoLayerData = get(["layers", "Video", "data"]);
+  const videoLoaded = !!videoLayerData;
+
+  return videoLoaded ? (
+    <div>
+      Current video time (seconds): {videoLayerData.playedSeconds}
+      <div style={{ padding: 20 }}>
+        <Slider
```
s0mya, line 15: Could this be defined in the layer and rendered here? I'm imagining a scenario where each layer could should additional info such as tracks, ticks, markers etc so that it any interesting sections of the video could be easily discovered via the UI.

pringshia: Ah yes, I agree. I'll add a TODO comment in the code here, will update to extract this out in the future.

I can imagine this ContextPanel serving two purposes: it could be either provide context at a Task level or at a Layer level (or... we just leave it up to the task designer an have them decide - so I guess that's a 3rd option).

Regardless, I'm trying to condense as much of the custom code into one file, currently `Task.js`, so I'd like to move this code there.

packages/annotation-ui/src/panels/LayersPanel.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import { Layers } from "../Task";
+import DebugPanel from "./DebugPanel";
+
+export default function ContextPanel() {
```
pringshia, line 5: Note to self: Rename to LayerPanels

s0mya:
Hey @pringshia this is looking great. I like how layers make it really flexible to create any kind of annotation and review interface. I don't have feedback just yet, mainly questions around what you had in mind to support some different use cases. 

1. I didn't quite understand what "Query 1", "Query 2" was meant to be, Is that a high-level annotation data. e.g. Hands, Faces? 
2. Similar question for "Response Track" and "Query Frame". What are they mean to be? 
3. Could the BBoxFrame be usable in multiple scenarios? i.e. could we support multiple bounding box layer 1 for hands, 1 for faces etc. Just the data is different, but the component is the same. Maybe I missed something, I but I couldn't see how you could pass this into the component. 
4. It'd be useful for us to walk through, but not actually build, some of droidlet and Tactile requirements. I think they both will need a pose visualizer / editor. What would they need to do?

pringshia:
> 1. I didn't quite understand what "Query 1", "Query 2" was meant to be, Is that a high-level annotation data. e.g. Hands, Faces?

These are use-case specific layers for Ego4D's VQA task. They will be extracted out of the base toolkit in the future.

> 2. Similar question for "Response Track" and "Query Frame". What are they mean to be?

These are also use-case specific layers. More information for these can be found in the Review UI Specifications section of [this doc](https://docs.google.com/document/d/1qWmV6Gf6s15E-jZy9Hwc5U686OZbbdZQOeBl2mrDUzU/edit#heading=h.t7uuzcv32byn) (under the Visual Queries & Answer task).

> 3. Could the BBoxFrame be usable in multiple scenarios? i.e. could we support multiple bounding box layer 1 for hands, 1 for faces etc. Just the data is different, but the component is the same. Maybe I missed something, I but I couldn't see how you could pass this data into the component.

As it is now, the BBoxFrame component contains both general logic and specific logic. My goal is to make this component very low-logic (more general). There are still some open questions for myself on how best to architect this and I'm open to suggestions. One possibility is that the BBox Layer itself should be something a user would create, while importing in the BBoxFrame which is general. Then they can pass data down to the BBoxFrame.

> 4. It'd be useful for us to walk through, but not actually build, some of droidlet and Tactile team requirements. I think they both will need a pose visualizer / editor. What would they need to do?

This is a good idea. It may be useful to build two tasks alongside each other. That way as we continue to build one, we can attempt to recreate another. This will help keep reusability and developer experience top-of-mind.

packages/annotation-ui/src/Task.js
```diff
@@ -30,7 +30,20 @@ function VQALayers() {
         <Layer
           displayName="Item Crop"
           icon="widget"
-          component={(props) => <BBoxFrame {...props} />}
+          component={(props) => (
```
s0mya, line 33: Nice! Hopefully this makes it easier to reuse the Bounding box component.

s0mya (approved):
LGTM!

sghmk12:
Looks good to me as well

sghmk12 (approved):

PR #421, opened 2021-03-24
Fix onboarding race condition

JackUrb:
## Overview
A race condition in the `Supervisor` was allowing some agents to be evaluated for the success of their onboarding before a response was fully processed. This condition occurs more frequently either under high stress, when the post-onboarding code is intensive, or when the data being saved after onboarding is large.

## Details
Two separate threads were handling the process of running an onboarding process from start to finish (within the `TaskRunner`) and receiving an onboarding submission message, validating onboarding, and then launching a full task. This change now brings the validation of onboarding and launch of a full task to happen only after the onboarding task run is fully completed.

The race condition present is actually the root cause of #215, and this fix lets us simplify the logic introduced to manage the split outcomes back then.

## Testing
Ran the static run with onboarding, had a worker complete onboarding, fail onboarding, and disconnect in onboarding. All three were handled as expected.

### DIFF
--- a/mephisto/operations/supervisor.py
+++ b/mephisto/operations/supervisor.py
@@ -266,15 +266,8 @@ class Supervisor:
             "request_id"
         ]
         del packet.data["request_id"]
-        assert isinstance(
-            agent, OnboardingAgent
-        ), "Only onboarding agents should submit onboarding"
         agent.pending_actions.append(packet)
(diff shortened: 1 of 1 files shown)


ytsheng (approved):
Confirmed on my end that this works. Thank you for fixing!

PR #423, opened 2021-03-28
[BUG FIX] Fixes bug when searching mephisto review with multiple pages of data

sghmk12:
### Bug
Small bug fix. Previously, when using mephisto review app with multiple pages of data, if a search query was made on the nth page of all data (n is greater than 1), the query would be made for the nth page of filtered data and for some search queries, would not return any data as there were less than n pages of filtered data.

### Fix
Now app resets the page parameter of a data request upon a new search query if the page parameter is not already equal to 1.

### DIFF
--- a/packages/cra-template-mephisto-review/template/src/AllItemView.jsx
+++ b/packages/cra-template-mephisto-review/template/src/AllItemView.jsx
@@ -36,7 +36,7 @@ function AllItemView({
   } = useMephistoReview({ page, resultsPerPage, filters });
 
   const setFiltersAndResetPage = (filtersStr) => {
-    if (page != null && page != 1) setPage(1);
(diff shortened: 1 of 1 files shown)


packages/cra-template-mephisto-review/template/src/AllItemView.jsx
```diff
@@ -35,14 +35,19 @@ function AllItemView({
     totalPages,
   } = useMephistoReview({ page, resultsPerPage, filters });
 
+  const setFiltersAndResetPage = (filtersStr) => {
+    if (page != null && page != 1) setPage(1);
```
pringshia, line 39: I'd recommend `!==`

pringshia (approved):

PR #427, opened 2021-04-01
annotation-toolkit: new packages, better developer workflow, docs

pringshia:
# Overview

Ok this PR is big.

It enables the following PR, which is in a sense the other half of the PR: https://github.com/fairinternal/mephisto-internal/pull/10.

https://user-images.githubusercontent.com/425059/113443935-e5013e80-93c0-11eb-89bf-f39603487616.mp4

It introduces a couple new libraries (`annotation-toolkit` and `global-context-store`), a ton of new features, bug fixes, and enhancements.

Major changes summarized below.

---

### Feature: Release of the `annotation-toolkit` and `global-context-store` library

Creating your own app using the `annotation-toolkit` is as easy as:

```jsx
import Store from "global-context-store";
import { AppShell } from "annotation-toolkit";

// ...
return (
  <Store>
    <AppShell layers={/* ... */} />
  </Store>
)
```

- Initial docs for `global-context-store` at: **https://www.npmjs.com/package/global-context-store**
- Initial docs for `annotation-toolkit` at: **https://www.npmjs.com/package/annotation-toolkit**

You can reference https://github.com/fairinternal/mephisto-internal/pull/10 for more information on usage. I'll be expanding on the docs in the coming few months as other high-pri items clear off of my plate.

`global-context-store` is actually a pretty versatile little library that can be used for many different projects. 

---

### Enhancement: Add `hostname` property for the `useMephistoReview` hook

A big pain point when developing review tasks in the past was the review server is launched via the `$ mephisto review` CLI command which expects a build output to host. This makes it difficult to develop in real-time using hot-reloading, since the CLI command expects a static build output.

We can work around this by running BOTH the CLI command (to launch a review server) as well as something like create-react-app's `npm run start` command together.

In the past, this wasn't possible since the UI hardcoded the path for the review server, using the same hostname as itself. However, this is now modifiable as such:

```
useMephistoReview({ hostname: "http://localhost:5000" })
```

The default value for `hostname` is `""` which reverts to the existing behavior of using the same hostname as the UI.

---

### Enhancement: Easier local development with Lerna

As we create and develop on more JS packages, [Lerna](https://github.com/lerna/lerna) is now setup to make local development a LOT easier. It especially helps to avoid this dreaded warning: https://reactjs.org/warnings/invalid-hook-call-warning.html

Here's how to use it
```bash
# Install lerna in the root of the Mephisto repo:
npm install

# also from the root, run:
lerna bootstrap --hoist="{react,react-dom}"
# this will link all packages that depend on each other, to each other, while also hoisting react and 
# react-dom to ensure that all packages uses the same version of react.

# You can then link the common react version hoisted at the root, to your client project located somewhere else on disk.
# Again, from the root:
npm link /other-project/located-somewhere-else/node_modules/react
# This ensure that all projects use the same version of React.

# then in your client project, you can do something like this:
npm link annotation-toolkit global-context-store

# NOTE: It's important to specify the 2 links together! otherwise i'm noticing behavior where
# the second link wipes out the first (npm v7.6.0 & node v15.8.0)
```


packages/cra-template-mephisto-review/template/src/components/AllItemView.jsx
```diff
@@ -10,10 +10,10 @@ import {
   NavbarHeading,
   Alignment,
 } from "@blueprintjs/core";
-import { Tooltip2 } from "@blueprintjs/popover2";
-import { DefaultItemRenderer } from "./plugins/DefaultItemRenderer";
-import { DefaultItemListRenderer } from "./plugins/DefaultItemListRenderer";
-import { Pagination } from "./components/Pagination";
+import { Tooltip } from "@blueprintjs/core";
```
pringshia, line 13: I think we can just use Tooltip here instead of Tooltip2 and avoid the additional dependency. In the future we'll migrate to @blueprintjs/core v4 anyway.

packages/cra-template-mephisto-review/template/src/index.js
```diff
@@ -1,37 +1,26 @@
...
   <React.StrictMode>
     <Router>
       <Switch>
         <Route path="/:id">
-          {/*
-          This route allows users to review data for individual items.
-          Add custom renderers for item data by adding an 'itemRenderer' property.
-          For more information see the 'Customization' section of the README.md file.
-          */}
+          {/* For more information see the 'Customization' section of the README.md file. */}
```
pringshia, line 19: Thought the above comments didn't add much so removed

packages/cra-template-mephisto-review/template/src/index.js
```diff
@@ -1,37 +1,26 @@
...
         <Route path="/">
-          {/*
-          This route shows all data items being reviewed by page.
-          If you do not wish to use pagination, set the 'pagination' property to 'false' (default is true).
-          The default number of results per page is 9, however can be adjusted by setting the 'resultsPerPage' property (must be an integer).
-          Add custom renderers for the layout of all items by passing an 'itemListRenderer' property.
-          Add custom itemRenderers for individual items by adding an 'itemRenderer' property.
-          For more information see the 'Customization' section of the README.md file.
-          */}
+          {/* For more information see the 'Customization' section of the README.md file. */}
```
pringshia, line 23: Thought the above comments didn't add much that wasn't already obvious from reading the code, so removed

packages/global-context-store/src/Store.js
```diff
@@ -44,38 +34,50 @@ const Reducer = (state, action) => {
...
+    },
+    [dispatch]
+  );
 
-  const sendRequest = (layer, action) => {
-    if (!get(["layers", layer, "data", "requests"])) {
-      set(["layers", layer, "data", "requests"], []);
-    }
-    invoke(["layers", layer, "data", "requests"], (prev) => [...prev, action]);
-  };
```
pringshia, line 74: This was very `annotation-toolkit` specific so I moved it into a helper, to keep `global-context-store` a generic utility.

packages/global-context-store/src/Store.js
```diff
@@ -44,38 +34,50 @@ const Reducer = (state, action) => {
...
+      }
+      invoke(path, (prev) => [...prev, value]);
+    },
+    [get, set, invoke]
+  );
 
   return (
     <Context.Provider
-      value={{ state, dispatch, set, get, invoke, unset, sendRequest }}
+      value={{ state, dispatch, set, get, invoke, unset, push }}
```
pringshia, line 80: Generalized what `sendRequest` was, which was basically an array push. Clients can use it as they see fit, which is what `annotation-toolkit` does.

packages/mephisto-review-hook/src/index.js
```diff
@@ -21,8 +22,8 @@ function useMephistoReview({
       return;
     }
     const DATA_URL = taskId
-      ? `/data/${taskId}`
-      : `/data?${page ? "page=" + page : ""}${
+      ? `${hostname}/data/${taskId}`
+      : `${hostname}/data?${page ? "page=" + page : ""}${
```
pringshia, line 26: allows you to specify where the review server exists. good for local development.

packages/cra-template-mephisto-review/template/src/Index.css
```diff
@@ -67,7 +67,7 @@ code {
   align-items: center;
   text-align: center;
   font-size: calc(10px + 1vmin);
-  margin-top: 50px;
+  padding-top: 50px;
```
pringshia, line 70: Helps prevent a nasty margin-collapsing bug I was running into when integrating `annotation-toolkit` into the review template. Basically the page always had a weird whitespace at the bottom and a forced scrollbar because of it, instead of nicely taking up 100% of the screen.

packages/annotation-toolkit/src/panels/ContentPanel.js
```diff
@@ -0,0 +1,147 @@
...
+          ? [...acc.actions, component.actions]
+          : acc.actions;
+        const actionPaths = component.actions
+          ? [...acc.actionPaths, path.join(" / ")]
+          : acc.actionPaths;
+        return { actions, path, actionPaths };
+      },
+      { actions: [], path: [], actionPaths: [] }
+    );
+  }
```
pringshia, line 74: This method gathers all the actions for the currently active layer as well as all of its parents up the tree.

packages/annotation-toolkit/src/layers/VideoPlayer.js
```diff
@@ -0,0 +1,128 @@
...
+    [vidRef.current]
+  );
+
+  React.useEffect(() => {
+    /* Determine video dimensions */
+    if (!videoLoaded) return;
+    if (!vidRef.current.getInternalPlayer()) return;
+    const videoWidth = vidRef.current.getInternalPlayer().videoWidth;
+    const videoHeight = vidRef.current.getInternalPlayer().videoHeight;
+    set(path("detectedSize"), [videoWidth, videoHeight]);
```
s0mya, line 59: Does it make sense to extract variables as constants? Help avoid bugs due to typos.

packages/annotation-toolkit/src/layers/VideoPlayer.js
```diff
@@ -0,0 +1,128 @@
+import React, { useContext, useEffect, useCallback, useRef } from "react";
+import ReactPlayer from "react-player";
+import { useStore } from "global-context-store";
+import { Spinner } from "@blueprintjs/core";
+
+export default function VideoPlayer({
```
s0mya, line 6: Overall this looks great! I was curious how to communicate to the users of this component the API it provides. i.e. global data variables that are set ("playedSeconds", "duration", etc) and the actions that are supported (seek, ff, screenshot)

packages/annotation-toolkit/src/layers/VideoPlayer.js
```diff
@@ -0,0 +1,128 @@
...
+          set(path("duration"), duration);
+        }}
+        onSeek={() => {}}
+      />
+      <canvas
+        style={{ visibility: "hidden" }}
+        height={height}
+        width={width}
+        ref={canvasRef}
+      ></canvas>
```
s0mya, line 125: Oh Interesting, I was surprised to see a canvas in a video player, but then looked up to see it's being used for screenshots, which makes sense.

s0mya (approved):
Looks great!

packages/annotation-toolkit/src/layers/VideoPlayer.js
pringshia: This is being tracked with #433 will likely add in the future

packages/annotation-toolkit/src/layers/VideoPlayer.js
pringshia: Will add in a follow-up. Being tracked with #434

PR #428, opened 2021-04-01
[FIX] Added responsive behaviour to mephisto review components

sghmk12:
### Summary
Previously buttons and inputs on the navigation bar of the mephisto review UI would move out of place on smaller screens. This small fix hides unnecessary texts on smaller screens such that this does not happen.

### Video Preview

https://user-images.githubusercontent.com/13241633/113361872-fdb71900-931a-11eb-9c65-1c48b38d6aed.mov


pringshia (approved):
Awesome, thanks

PR #432, opened 2021-04-12
Creation of the SingletonMephistoDB

JackUrb:
## Overview
In order to increase the speed of Mephisto, this PR trades off active memory (and eventual memory leaks) for speed. All data model accesses will refer to cached objects in the `SingletonMephistoDB`, such that object initialization (which accounts for a _surprising_ amount of Mephisto's runtime, more on this later) is minimized. This database is only particularly useful in a running mode with an `Operator`, and direct database changes are pushed to the database, but the actual data model objects will all be singletons during use. Thus, when Mephisto later needs to "load all the units" to get their status, they're all already loaded. 

### Usage
You can use this feature with the `mephisto.database._database_type=singleton` hydra arg. At the moment this is to be considered a beta feature, and is subject to some change.

## Implementation
1. Add the `optimized_load` and `cache_result` method to the `MephistoDB` class. These methods provide an anchor for which a `MephistoDB` setup can produce faster results based on a local (or external) cache than a regular `__new__` and `__init__` call that python would generally do.
2. Creates the `MephistoDBBackedMeta` and `MephistoDBBackedABCMeta` metaclasses. These metaclasses will ensure that during initialization (like calling `Agent(db, "800")`) will first check the `MephistoDB`'s `optimized_load` method before doing a default call, and will store results via `cache_result`. We assume we only need to manually call `cache_result` if `optimized_load` didn't return a value, as the underlying `MephistoDB` could have cached locally otherwise.
3. Update all of the Mephisto data model objects (conveniently all with the same init signature of `(db: MephistoDB, db_id: str)`) to extend these metaclasses. The `ABC` version extends the `ABCMeta` metaclass, this way the data model classes that used to extend `ABC` can still maintain their abstract nature. 
4. Create the `SingletonMephistoDB`. This class is a wrapper around the `LocalMephistoDB`, but it utilizes the `optimized_load` and `cache_result` methods. When loading from the DB, now we check to see if there's a cached version first and return that if it exists. All of these cache entries are stored with `WeakValueDictionary`s, this way we can do _some_ garbage collection where possible.
5. `weakref` backwards edges in `AgentState` classes. This allows the `SingletonMephistoDB` to clean up agents when no longer referenced outside of circular reference cases.
6. Adds `mephisto.database._database_type` as a hydra argument to be able to configure to use `singleton`. Default is `local`.

Note that step 5 here doesn't actually cover nearly all of the possible `weakref`-relevant cases: more experimentation can be done to update the vertical path of the datamodel to `weakref` the upwards references that currently exist, like `<agent>._unit`.
## Testing
1. Added `TestMephistoSingletonDB`, which passes the base Database Tests.
2. Can successfully run example task using the singleton DB:
`python static_test_script.py mephisto.database._database_type=singleton`
3. Original testing passes.

### DIFF
--- a/mephisto/abstractions/database.py
+++ b/mephisto/abstractions/database.py
@@ -92,6 +92,22 @@ class MephistoDB(ABC):
             provider = ProviderClass(self)
             provider.cleanup_qualification(qualification_name)
 
+    def optimized_load(
+        self,
+        target_cls,
+        db_id: str,
+        row: Optional[Mapping[str, Any]] = None,
+    ):
+        """
+        Load the given class in an optimized fashion, if this DB has a more
+        efficient way of storing and managing the data
+        """
+        return None
+
+    def cache_result(self, target_cls, value) -> None:
+        """Opportunity to store the result class from a load"""
+        return None
+
     @abstractmethod
     def shutdown(self) -> None:
         """Do whatever is required to shut this server off"""
--- a/mephisto/abstractions/blueprints/abstract/static_task/static_agent_state.py
+++ b/mephisto/abstractions/blueprints/abstract/static_task/static_agent_state.py
@@ -9,6 +9,7 @@ from mephisto.abstractions.blueprint import AgentState
 import os
 import json
 import time
+import weakref
 
 if TYPE_CHECKING:
     from mephisto.data_model.agent import Agent
@@ -38,7 +39,7 @@ class StaticAgentState(AgentState):
         Static agent states should store
         input dict -> output dict pairs to disc
         """
-        self.agent = agent
+        self.agent = weakref.proxy(agent)
         self.state: Dict[str, Optional[Dict[str, Any]]] = self._get_empty_state()
         self.load_data()
 
--- a/mephisto/abstractions/blueprints/mock/mock_agent_state.py
+++ b/mephisto/abstractions/blueprints/mock/mock_agent_state.py
@@ -8,6 +8,7 @@ from typing import List, Optional, Dict, Any, TYPE_CHECKING
 from mephisto.abstractions.blueprint import AgentState
 import os
 import json
+import weakref
 
 if TYPE_CHECKING:
     from mephisto.data_model.agent import Agent
@@ -21,7 +22,7 @@ class MockAgentState(AgentState):
 
     def __init__(self, agent: "Agent"):
         """Mock agent states keep everything in local memory"""
-        self.agent = agent
+        self.agent = weakref.proxy(agent)
         self.state: Dict[str, Any] = {}
         self.init_state: Any = None
 
--- a/mephisto/abstractions/blueprints/parlai_chat/parlai_chat_agent_state.py
+++ b/mephisto/abstractions/blueprints/parlai_chat/parlai_chat_agent_state.py
@@ -13,6 +13,7 @@ from mephisto.data_model.packet import (
 import os
 import json
 import time
+import weakref
 
 if TYPE_CHECKING:
     from mephisto.data_model.agent import Agent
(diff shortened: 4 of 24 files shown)


mojtaba-komeili (commented):
It generally looks OK to me.
But I may not be able to provide useful feedback because of my unfamiliarity with Python's weak references and many classes that this PR touches (basically everything in `_cached_classes`).

mephisto/abstractions/database.py
```diff
@@ -92,6 +92,22 @@ def delete_qualification(self, qualification_name: str) -> None:
...
+        """
+        Load the given class in an optimized fashion, if this DB has a more
+        efficient way of storing and managing the data
+        """
+        return None
+
+    def cache_result(self, target_cls, value) -> None:
+        """Opportunity to store the result class from a load"""
+        return None
+
```
pringshia, line 110: I wonder if adding these optimization overrides to the abstract class may make it too specific/be a leaky abstraction. It seems like from the perspective of an abstract class, it doesn't matter if a "load" is optimized or not, a load is just a load.

I'm not sure of the full architecture here, but maybe another route could be the ability to create an OptimizedSQLLiteDatabase that inherits from a SQLLiteDatabase class which then overrides certain getters, setters, and loaders. And then you could use the optimized class instead. Just a thought.

pringshia (approved):
Overall LGTM. Speed optimizations are great. I'm curious do you have any sort of metrics on how fast of a speedup this is?

mephisto/abstractions/database.py
JackUrb: The difficulty here is that we want to be able to optimize the load with our most basic API - direct initialization (`Agent(db, db_id)`). I'm not sure how to expose such a possibility to the DB other than defining a method in the abstract base class. I suppose I could be more exact here in noting that this would be the approach to do a `possibly_load_from_cache`, as this is the only optimization form we have at the moment, but I'm not sure if another DB may do something differently in the future. In this case it's less the SQLite load that's a problem as much as the object initialization at this point.

JackUrb:
> Overall LGTM. Speed optimizations are great. I'm curious do you have any sort of metrics on how fast of a speedup this is?

It's hard to disentangle the results of this from #341 and #435, but overall these together are expected to cut away ~90% of the previous execution time. Using this particular change without the other two would likely net around 3/4 of that, but the changes elsewhere mean that we still see ~40% improvement running on the old db.

stephenroller:
I do feel a bit strange that meta classes are being abused essentially to implement a singleton pattern here. I reckon that's just for backwards compatibility?

JackUrb:
> I do feel a bit strange that meta classes are being abused essentially to implement a singleton pattern here. I reckon that's just for backwards compatibility?

Correct - doing this again from scratch I'd have the api be:
```
Agent:
  @classmethod
  get(db, db_id):
    ...
```
And then we could do `Agent.get(db, id)` rather than `Agent(db, id)`.

stephenroller:
Ya but why not just do that, have init reference a singleton counter and have a warning recommending migration to the new pattern if the counter goes above 1?

JackUrb:
At the moment we're not forcing people over to the singleton pattern, and in fact there are times you explicitly do not want to use it (say, using an external process to examine a currently running task), so I'm not sure how to put the warning in.

I at once point considered using `Agent.__call__` right now to _redirect_ to `Agent.get`, though that would make it such that `Agent.get` for now needs to directly invoke `__new__` and `__init__`. I could then slap a deprecation warning on calling the data model initializers directly? I'd guess this is a happy medium for now, and would make it possible to drop the metaclasses in the future?

PR #435, opened 2021-04-12
Setting cache invalidation for sync_hit_mapping

JackUrb:
## Overview
As it's important for us not to assign two Units the same HIT id, we have a lock around the database reads for the `MTurkDatastore`. Unfortunately, this means that every time we check the status of a HIT-Unit pairing in this DB, we're locked. We tend to do these syncs in a sequential fashion, so this turns out to be a rather large time lock.

We _could_ do something with read/write specific locks, but I feel that overcomplicates the setup - luckily we don't do the writes very often, so we can instead move to a model where readers can skip the read (and the lock) entirely so long as they've synced more recently than the most recent invalidation. 

## Implementation
`MTurkUnit`s now keep track of their `_last_sync_time`, which corresponds to the last time they got their `hit_id` and `assignment_id` from the db. When asked to sync, if this time is more recent than the last invalidation, they skip the sync.

The `MTurkDatastore` now also keeps a time for the last time a change was made to the table corresponding to this table, and as such the `MTurkUnit` can query `is_hit_mapping_in_sync` to ensure it's still valid.

## Testing
1. Ran local tests
2. Launched and completed an MTurk sandbox job.

### DIFF
--- a/mephisto/abstractions/providers/mturk/mturk_datastore.py
+++ b/mephisto/abstractions/providers/mturk/mturk_datastore.py
@@ -76,7 +76,7 @@ class MTurkDatastore:
         self.db_path = os.path.join(datastore_root, "mturk.db")
         self.init_tables()
         self.datastore_root = datastore_root
-        self._last_hit_mapping_update_time = time.time()
+        self._last_hit_mapping_update_time = time.monotonic()
 
     def _get_connection(self) -> sqlite3.Connection:
         """Returns a singular database connection to be shared amongst all
@@ -90,6 +90,13 @@ class MTurkDatastore:
             conn.row_factory = sqlite3.Row
             self.conn[curr_thread] = conn
         return self.conn[curr_thread]
+
+    def _mark_hit_mapping_update(self) -> None:
+        """
+        Update the last hit mapping time to mark a change to the hit
+        mappings table and allow dependents to invalidate caches
+        """
+        self._last_hit_mapping_update_time = time.monotonic()
 
     def init_tables(self) -> None:
         """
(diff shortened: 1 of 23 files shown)


mephisto/abstractions/providers/mturk/mturk_datastore.py
```diff
@@ -122,6 +130,7 @@ def new_hit(self, hit_id: str, hit_link: str, duration: int, run_id: str) -> Non
                 ) VALUES (?, ?);""",
                 (hit_id, run_id),
             )
+            self._last_hit_mapping_update_time = time.time()
```
mojtaba-komeili, line 133: nit: maybe define a new function for all the updates like this. In case you wanted to log updates or something later. For example,
```
def set_cache_updated():
  self._last_hit_mapping_update_time = time.time()
```

mephisto/abstractions/providers/mturk/mturk_unit.py
```diff
@@ -68,6 +72,10 @@ def _sync_hit_mapping(self) -> None:
             self.hit_id = None
             self.mturk_assignment_id = None
             self.assignment_time_in_seconds = -1
+        # We update to a time slightly earlier than now, in order
+        # to reduce the risk of a race condition caching an old
+        # value the moment it's registered
```
mojtaba-komeili, line 77: Nice 🆒  trick

mojtaba-komeili (approved):

bottler:
It might be better to use `time.monotonic()` instead of `time.time()`, both because you'll avoid any problems if system time does get adjusted and because maybe any duplicates problem goes away.

PR #437, opened 2021-04-15
Deprecate old initializers

JackUrb:
## Overview
Such that we can eventually move _away_ from using metaclasses to enable usage of optimized loading, this PR prepares us to move to `DataModelItem.get(db, id)` rather than what we had. Once that's the standard method of initializing, we can remove the metaclasses cleanly. Target date for that removal is late June 2021.

## Implementation
Creates the `MephistoDataModelComponentMixin` class which provides the `get` method. I may need to do something in the future for getting the typechecking of `get` to return the correct class, but that's future work.

## Testing
CI tests


stephenroller (approved):
:D

PR #438, opened 2021-04-15
Update internal inits

JackUrb:
## Overview
Following the steps in #437 to deprecate the old method of initializing, this PR simply updates all of _our_ callsites to use the new version. This is excepting Database calls to the original `__init__` methods, as we assume any requests made _inside_ a database class could be optimized inside if they wanted to, and so the choice to not use `get` is to be respected.

## Implementation details
Ctrl-F for the various inits internally. 

## testing
CI tests


JackUrb:
Abandoning in favor of including these commits directly with #437 so that we don't need to rebase two commits.

PR #439, opened 2021-04-19
Fixing parlai onboarding

JackUrb:
## Overview
#421 included a bug for specific cases of `ParlAIBlueprint` onboarding. This fix resolves the issue, and also resolves an issue where the timeout status wasn't being pushed forward in onboarding.

## Details
#421 introduced a bug _specifically_ for `ParlAIBlueprint` onboarding worlds, as the `ParlAIBlueprint` is the only type of live task wherein _either_ the backend (the `World`) or the frontend (via `handleSubmit`) can submit the onboarding. In the case that the world exited cleanly, this would lead to a message to the frontend telling it to submit.

Unfortunately, #421 improved the onboarding flow to _ensure_ that the transition from running onboarding to registering would happen in the same order every time. Specifically, the requirement is that `_on_submit_onboarding` now has to _trigger_ the termination of the onboarding runner. In the case where the backend triggered this for parlai onboarding by just exiting an onboarding world, that _wasn't_ the order things happened in. 

This updates the `ParlAChatITaskRunner` to _first_ mark the agent as done, then _wait_ for the submit packet (to come through `_on_submit_onboarding` before terminating.

## Testing
Launched a task world with onboarding worlds that submit on the backend, was able to transition into the main world with no refreshes, no disconnects, and no other behavior.
Launched a task world with onboarding, timed out, then ensured the timeout was pushed to the frontend.


mephisto/abstractions/blueprints/parlai_chat/parlai_chat_task_runner.py
```diff
@@ -199,6 +200,13 @@ def run_onboarding(self, agent: "OnboardingAgent") -> None:
                     },
                 )
             )
+        # Mark the agent as done, then wait for the incoming submit action
+        agent.mark_done()
+        while not agent.has_action.is_set():
+            done_act = agent.act()
+            if done_act is not None:
+                break
+            time.sleep(0.3)
```
mojtaba-komeili, line 209: question: could we have this as a variable in case we wanted to find the optimal value?

mojtaba-komeili:
The code looks good to me. I tried this in mock, sandbox and live. Mock worked well. In sandbox, when it was transitioning it looked like it was refreshing a bit too long, but it was generally good. The live (prod) version is going fine but a bit too slow (that might be related to other factors though).

mojtaba-komeili (approved):

Issue #431, opened 2021-04-12
Expiring a mostly-finished task run.


bottler:
If I have a large task run, expiring it expires all its units, which sets the expiry on mturk even if the unit is completed. This can take a long time. I am wondering if this is worth optimizing. If so, would the following change in mephisto, in `mturk_unit.py`, be safe?

```
     def expire(self) -> float:
         """
         Send a request to expire the HIT, and if it's not assigned return,
         otherwise just return the maximum assignment duration
         """
         delay = 0

+        status = self.get_status()
+        if status in [AssignmentState.EXPIRED, AssignmentState.COMPLETED]:
+            return delay

-        if self.get_status() == AssignmentState.ASSIGNED:
+        if status == AssignmentState.ASSIGNED:
```

JackUrb:
This makes sense to me, completed or expired HITs shouldn't need to be re-expired. The only concern would be ensuring that these units indeed have an associated `HIT` already, but if they have a completed `Agent` associated with them I don't see how you could get here with a `Unit` that has no mapped HIT.

bottler:
I've been testing this out, running large runs for a few days, and it seems to be working. I will make a PR.

There are often cases where I expire a run* and wait a while (much longer than the assignment duration) and find that there's a hit or two in the "launched" state. In this case, expiring the run again succeeds in cleaning up. 

* i.e. `operator.get_running_task_runs()[...].task_launcher.expire_units()`

JackUrb:
Interesting to note that there are hits that are still in LAUNCHED - I wonder if they transitioned _back_ to that state after being expired, or if something else happens for these. I'd imagine that on the MTurk side, the HIT is actually already expired.

bottler:
Yes, they are expired on the mturk side, so something has gone wrong in keeping track of the status inside mephisto. Waiting doesn't lead mephisto to correct itself, though, and I don't think calling `get_status` on them does it either.

JackUrb:
I'm betting that the task gets returned after it has been 'expired', which is causing the transition to the `LAUNCHED` state. Will need to see how to reconcile the transition of a return if the underlying task is no longer available.

federicoruggeri:
I'm not sure if this comment is directly related to above issue. I've noticed that if timeout occurs, then the unit state follows the same state change: from `assigned` to `expired` and then to `launched`. This occurs everytime timeout expires (tested with local and heroku architect and mock agent on dialogue task).

[...]

PR #440, opened 2021-04-21
Make expire fast for completed units

bottler:
As I mentioned in https://github.com/facebookresearch/Mephisto/issues/431, we don't need to contact boto to expire a hit if it is already expired or completed.


JackUrb (approved):
Thanks for this improvement @bottler!

PR #441, opened 2021-04-23
Adding assignment-unit cache to singleton DB

JackUrb:
## Overview
`Assignment.get_units` is currently the largest remaining call in the Mephisto profile when running with `MephistoSingletonDB` This PR looks to remove this chunk by adding a cache for this specific call.
Screen Shot 2021-04-23 at 5 00 10 PM

## Implementation
Creates a dict in the `MephistoSingletonDB` to hold a mapping between the given `assignment_id`s queried and the returned list of `Unit`s. Also adds a wrapper around `new_unit` to ensure that if a new unit is added to a given `assignment_id` we clear that cached value. Considering we're within a singleton, this is sufficient to ensure that the local units will always be up-to-date.

## Testing
In python shell:
```
>>> from mephisto.abstractions.databases.local_singleton_database import MephistoSingletonDB
>>> db = MephistoSingletonDB()
>>> from mephisto.data_model.task_run import TaskRun
>>> len(TaskRun(db, 1188).get_assignments())
300
>>> TaskRun(db, 1188).get_assignment_statuses()
{'created': 0, 'launched': 2, 'assigned': 0, 'completed': 0, 'accepted': 0, 'mixed': 0, 'rejected': 0, 'soft_rejected': 0, 'expired': 298}
>>> TaskRun(db, 1188).get_assignment_statuses()
{'created': 0, 'launched': 2, 'assigned': 0, 'completed': 0, 'accepted': 0, 'mixed': 0, 'rejected': 0, 'soft_rejected': 0, 'expired': 298}
```
The second call to `get_assignment_statuses()` was instant, where the first took about a second.

### DIFF
--- a/mephisto/abstractions/databases/local_singleton_database.py
+++ b/mephisto/abstractions/databases/local_singleton_database.py
@@ -68,6 +68,7 @@ class MephistoSingletonDB(LocalMephistoDB):
 
         # Create singleton dictionaries for entries
         self._singleton_cache = {k: dict() for k in self._cached_classes}
+        self._assignment_to_unit_mapping: Dict[str, List[Unit]] = {}
 
     def shutdown(self) -> None:
         """Close all open connections"""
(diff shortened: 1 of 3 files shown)


mephisto/abstractions/databases/local_singleton_database.py
```diff
@@ -128,3 +129,93 @@ def new_agent(
...
+        requester_id: str,
+        assignment_id: str,
+        unit_index: int,
+        pay_amount: float,
+        provider_type: str,
+        task_type: str,
+        sandbox: bool = True,
+    ) -> str:
+        """
+        Create a new unit with the given index. Raises EntryAlreadyExistsException
```
mojtaba-komeili, line 206: Does throwing the exception happen in the super class?

JackUrb: Yes

mephisto/abstractions/databases/local_singleton_database.py
```diff
@@ -128,3 +129,93 @@ def new_agent(
...
+        if there is already a unit for the given assignment with the given index.
+        """
+        if assignment_id in self._assignment_to_unit_mapping:
+            del self._assignment_to_unit_mapping[assignment_id]
```
mojtaba-komeili, line 210: nit: I usually prefer popping dict keys instead of del. I am a bit scared of introducing bugs that may remove the whole dict instead of that single entry.

mephisto/operations/operator.py
```diff
@@ -306,7 +306,7 @@ def _track_and_kill_runs(self):
                     tracked_run.architect.shutdown()
                     tracked_run.task_launcher.shutdown()
                     del self._task_runs_tracked[task_run.db_id]
-            time.sleep(2)
+            time.sleep(10)
```
mojtaba-komeili, line 309: Could you add this as hardcoded constant (probably all caps) at the top of the file or in a constants file (if you are using one). Specially having a descriptive name helps someone who may want to debug later see what is readily available for tweaking.

mojtaba-komeili (approved):
Thanks a lot Jack. Can't wait to run with new changes 😃

mephisto/abstractions/databases/local_singleton_database.py
JackUrb: I tend to prefer `del` as explicitly not caring about the value, `pop` if the value at the entry is important.

PR #442, opened 2021-04-26
[band-aid] Agent is none logging + ignore

JackUrb:
# Overview
During recent bouts of Heroku spottiness, we started receiving errors of the form `Agent is None`, which points to a `Unit` being in the `COMPLETED` state, while the `Agent` is not (eventually disconnecting). This would mean that a worker was able to submit, but we didn't receive the message. 

We _have to_ allow the worker to submit in this instance, but there's no clear way to reconcile the state at the moment. This PR also includes a change that _attempts_ to do this reconciliation, however the saved data is not guaranteed to be the same format as the original data.

# Implementation
In the case we find a `MTurkUnit` transitioning to `COMPLETED`, but agent hasn't advanced passed `IN_TASK` we try a reconciliation of loading the assignment data from MTurk directly. We then parse it to a form as close as we can get to the form it would've come in from the server, and pass it along as an action to get the job marked completed.

The specific call is `MTurkAgent.attempt_to_reconcile_submitted_data(MTurkUnit.get_mturk_hit_id())` - this method can be called to reconcile an agent _after_ a task is completed and no longer running with the following:
```
agent.attempt_to_reconcile_submitted_data(unit.get_mturk_hit_id())
agent.act()
agent.mark_done()
```

# Discussion
This comes with a number of risks:
1. The reconciliation isn't guaranteed to work, and if MTurk changes their format substantially we'll have to adjust
2. The format we get from reconciliation may not be the same format as the original, and the task runner may crash anyways. This is still _better_ than before, but only marginally
3. If the data is saved, but doesn't 100% cover the form that any further along parsing scripts expect, it's possible that a future `Agent.state.get_data()` call doesn't have all of the fields expected, and this could mess with downstream scripts. For now I've added a key for `MEPHISTO_MTURK_RECONCILED` to the data, but I'm not sure this is sufficient.

I feel these should be discussed a bit and given some time to iterate before this change is merged.

### DIFF
--- a/mephisto/abstractions/providers/mturk/mturk_unit.py
+++ b/mephisto/abstractions/providers/mturk/mturk_unit.py
@@ -150,7 +150,7 @@ class MTurkUnit(Unit):
                 elif agent_status == AgentState.STATUS_SOFT_REJECTED:
                     found_status = AssignmentState.SOFT_REJECTED
             else:
-                logger.warning("Agent is None")
+                logger.warning(f"Agent for unit {self} is None")
             if found_status != self.db_status:
                 self.set_db_status(found_status)
             return self.db_status
--- a/mephisto/abstractions/providers/mturk/mturk_unit.py
+++ b/mephisto/abstractions/providers/mturk/mturk_unit.py
@@ -109,6 +109,36 @@ class MTurkUnit(Unit):
             self.__requester = cast("MTurkRequester", super().get_requester())
         return self.__requester
 
+    def set_db_status(self, status: str) -> None:
+        """
+        Set the status reflected in the database for this Unit
+        """
+        super().set_db_status(status)
+        if status == AssignmentState.COMPLETED:
+            agent = self.get_assigned_agent()
+            if agent is not None:
+                agent_status = agent.get_status()
+                if agent_status == AgentState.STATUS_IN_TASK:
+                    # Oh no, MTurk has completed the unit, but we don't have
+                    # the data. We need to reconcile
+                    logger.warning(
+                        f"Unit {self} moved to completed, but the agent didn't... "
+                        f"Attempting to reconcile with MTurk directly"
+                    )
+                    try:
+                        agent.attempt_to_reconcile_submitted_data(
+                            self.get_mturk_hit_id()
+                        )
+                    except Exception as e:
+                        logger.warning(
+                            f"Was not able to reconcile due to an error, {e}. "
+                            f"You may need to reconcile this specific Agent manually "
+                            f"after the task is completed. See here for details: "
+                            f"https://github.com/facebookresearch/Mephisto/pull/442"
+                        )
+            else:
+                logger.warning(f"No agent found for completed unit {self}...")
+
     def clear_assigned_agent(self) -> None:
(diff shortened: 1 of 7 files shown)


mephisto/abstractions/providers/mturk/mturk_unit.py
```diff
@@ -109,11 +109,52 @@ def get_requester(self) -> "MTurkRequester":
...
+    def set_db_status(self, status: str) -> None:
+        """
+        Set the status reflected in the database for this Unit
+        """
+        super().set_db_status(status)
+        if status == AssignmentState.COMPLETED:
+            agent = self.get_assigned_agent()
+            if agent is not None:
+                agent_status = agent.get_status()
+                if agent_status == AgentState.STATUS_IN_TASK:
```
mojtaba-komeili, line 121: Do we switch agent status from in_task to something else, regardless of reconcile failing or not?

mojtaba-komeili:
Trying to compile the collected data from the last couple of days (with Mephisto on `singleton-find-unit-cache` 400af99), I get an error that indicates the save_data function from worlds had not included the full data dump. Could that be related to this bug? I paste the relevant part of the stack trace here:

```
  File "/private/home/komeili/dev/ParlAI/parlai/crowdsourcing/utils/analysis.py", line 188, in get_units_data
    task_data.append(data_browser.get_data_from_unit(unit))
  File "/private/home/komeili/dev/Mephisto/mephisto/tools/data_browser.py", line 88, in get_data_from_unit
    "data": agent.state.get_parsed_data(),
  File "/private/home/komeili/dev/Mephisto/mephisto/abstractions/blueprints/parlai_chat/parlai_chat_agent_state.py", line 99, in get_parsed_data
    if "WORLD_DATA" in messages[-1]
```

mojtaba-komeili (approved):
I will try to have my next round of data collection on this branch. Provided that we collect the stdout and profiler data from that one, we see how this works in practice. I will let you know when they were ready.
Thanks again for resolving the issue so quickly.

mephisto/abstractions/providers/mturk/mturk_unit.py
JackUrb: The reconciliation should put a final task act into the message queue for the agent being reconciled, which may be sufficient to move them to complete in the `TaskRunner`. This is definitely not a guarantee though.

JackUrb:
> Trying to compile the collected data from the last couple of days (with Mephisto on singleton-find-unit-cache 400af99), I get an error that indicates the save_data function from worlds had not included the full data dump. 

Did this happen very frequently? Definitely could be related to this bug on a failed submit event.

mojtaba-komeili:
> > Trying to compile the collected data from the last couple of days (with Mephisto on singleton-find-unit-cache [400af99](https://github.com/facebookresearch/Mephisto/commit/400af996038c118408b0c530d058904879aeb90d)), I get an error that indicates the save_data function from worlds had not included the full data dump.
> 
> Did this happen very frequently? Definitely could be related to this bug on a failed submit event.

Not a lot but a few of the units had this issue.

PR #443, opened 2021-04-26
Simple Supervisor shutdown array fix

JackUrb:
Right now shutdown isn't being particularly safe iterating over the agents to close. This updates that iteration to pre-populate the list of threads to join.

### DIFF
--- a/mephisto/operations/supervisor.py
+++ b/mephisto/operations/supervisor.py
@@ -201,6 +201,8 @@ class Supervisor:
 
     def shutdown(self):
         """Close all of the channels, join threads"""
+        # Prepopulate agents and channels to close, as
+        # these may change during iteration
         channels_to_close = list(self.channels.keys())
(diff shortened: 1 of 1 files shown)


mephisto/operations/supervisor.py
```diff
@@ -211,7 +211,8 @@ def shutdown(self):
         if self.sending_thread is not None:
             self.sending_thread.join()
         logger.debug(f"Joining agents {self.agents}")
-        for agent_info in self.agents.values():
+        agents_to_close = list(self.agents.values())
```
mojtaba-komeili, line 214: Could you add a comment explaining the reason for separating it like that here. If I am refactoring this code and I get to to this line, I definitely change it back to what it was before this PR.

mojtaba-komeili (approved):
Please add that comment to avoid future confusions.

mojtaba-komeili (approved):
Great, thank you!

PR #446, opened 2021-04-27
Getting complete MTurk pay for individual units

JackUrb:
# Overview
At the moment `MTurkUnit`s are not properly reporting their pay amount.  In order to actually calculate the amount of spend on a particular `Unit`, we need to accommodate both the MTurk fees for the base pay and bonuses, as well as query the assignment for bonuses in the first place. This PR introduces the `get_bonuses_for_assignment` method which returns the sum payment of bonuses for a given assignment, and then uses that in the `MTurkUnit`'s implementation for `get_pay_amount`.

### DIFF
--- a/mephisto/abstractions/providers/mturk/mturk_unit.py
+++ b/mephisto/abstractions/providers/mturk/mturk_unit.py
@@ -13,6 +13,9 @@ from mephisto.abstractions.providers.mturk.mturk_utils import (
     expire_hit,
     get_hit,
     create_hit_with_hit_type,
+    get_bonuses_for_assignment,
+    calculate_mturk_task_fee,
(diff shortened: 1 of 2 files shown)


pringshia:
This looks fine to me - question: are `get_bonuses_for_assignment` and `get_pay_amount` intended to be printed out anywhere? Or are they helpers for Mephisto task devs to use as part of their run scripts?

JackUrb:
This function is collected over by `TaskRun.get_total_spend()`, but isn't strongly documented yet (as interaction with sandbox and pilot runs is shaky at the moment). As we build out bookkeeping I'm sure we'll return to this more, starting with `get_total_spend()` itself.

PR #447, opened 2021-04-27
Extract all annotation-toolkit css into a single file at `build/main.css`

pringshia:
This is a webpack build update to extract all css for the annotation toolkit into a single file @ `/build/main.css`.
This is useful for single imports and CDN style imports (e.g. with iframe usage).

This is important to get iframe usage working internally with `react-frame-component`.


s0mya (approved):
Looks good to me.

PR #454, opened 2021-04-29
annotation-toolkit v1.1.0 - bug fixes, debugging enhancements, usability enhancements

pringshia:
Release notes are also available in `packages/annotation-toolkit/release-notes.md`. I find it useful to keep a running list there as I'm building up a PR. Pasted here for convenience:

This PR was made possible by dogfooding the `annotation-toolkit` with the following PR: https://github.com/fairinternal/mephisto-internal/pull/11

# annotation-toolkit v1.1.0
- **ENHANCEMENT** A helpful message is shown when `<AppShell />` is missing a `layers={...}` prop.

Screen Shot 2021-04-27 at 5 57 24 PM

- **NEW** Major update to the Debug Panel. Incorporates `react-json-inspector` for the "Inspect State" tab. A search bar also lets you search through the global state quickly.

Screen Shot 2021-04-28 at 10 25 33 PM

---

- **FIX** Fix a bug where actions could show up twice in the Actions Panel for `alwaysOn` layers that are also currently selected.
- **ENHANCEMENT** Add a '⬩' marker next to layers in the actions panel that are selected by virtue of active selection, as opposed to the `alwaysOn` prop. Helpful for debugging and intuiting why certain actions are currenlty shown, both for task creators and annotators.
- **NEW** Add a `hideName` boolean prop to `<Layer />` which can be used to prevent that layer from showing up in the Layers Panel. Useful for layers with a visual component, but no annotation-based interactivity. Default: `false`.
- **NEW** New `layerButtons` prop for `<AppShell />` to allow end-users to create toolbar buttons. Format: `{title: string, icon: string.blueprint-icon, action: fn}`
- **FIX** Fix a bug where actions specified for toolbar buttons were not being fired on click.
- **ENHANCEMENT** Add `dataPathBuilderFor(layerId) -> fn(...layerPathArgs)` to helpers. Usage: `const dataPath = dataPathBuilderFor('VideoLayer'); dataPath('playedSeconds');`
- **NEW** Add a `hideActionsIfUnselected` prop to `<Layer />`. Useful for `alwaysOn` or `onWithGroup` layers that should hide their actions from the Actions Panel when unselected.

---

- **BREAKING** Rename the `requestQueuePath` helper to `requestsPathFor`.
- **BREAKING** The global state schema for layers has been updated so that it contains three top level properties: `data`, `config`, and `requests`. `requests` has been moved out of `data` into it's own top-level property. Otherwise, `data` is the same as before. All remaining properties that were specified by the layer config are now moved under the `config` property.
    - To update, all accessors that were not part of the data prop must be updated, e.g. `layer.Video.displayName` -> `layer.Video.config.displayName`.
    - Accessors of `requests` also need to be updated. If you were using `requestQueuePath` before, you should be fine, as long as you apply the other breaking change fix above and rename it to `requestsPathFor`.

# Other

- Added a "how-to" guide for front-end development with the repo, specifically focusing on working with multiple packages and how to link them
- Added some more documentation for the annotation-toolkit components, still a WIP


packages/annotation-toolkit/src/helpers.js
```diff
@@ -2,6 +2,13 @@ export const frameToMs = (frame, fps) => (frame * 1000) / fps;
 
 export const msToFrame = (ms, fps) => Math.round((ms * fps) / 1000);
 
-export const requestQueue = (layer) => {
-  return ["layers", layer, "data", "requests"];
+export const requestsPathFor = (layerId) => {
+  return ["layers", layerId, "requests"];
 };
+
+export const dataPathBuilderFor = (layerId) => (...args) => [
```
s0mya, line 9: Worth adding a comment explaining this const?

packages/annotation-toolkit/src/utils.js
```diff
@@ -0,0 +1,5 @@
+export function isFunction(functionToCheck) {
+  return (
+    functionToCheck && {}.toString.call(functionToCheck) === "[object Function]"
```
s0mya, line 3: Wow, I'm surprised that's how you detect a function. 

Curious if the second answer from here also works? https://stackoverflow.com/questions/5999998/check-if-a-variable-is-of-function-type

s0mya (approved):
LGTM..

packages/annotation-toolkit/src/utils.js
JackUrb: I think I've used that before:
```
if (typeof v === 'function') {
    // do something
}
```
but unsure if the context matters.

packages/annotation-toolkit/src/helpers.js
pringshia: added some comments!

packages/annotation-toolkit/src/utils.js
pringshia: switched to the simpler version Jack mentioned, as underscore.js also uses that implementation

PR #456, opened 2021-05-05
Preventing data browser crash, if there is no matching unit

mojtaba-komeili:
1. While retrieving HITs for an MTurk client, there are situations that `boto` throws and exception because of not being able to retrieve a HIT. This may crash the `data_examiners` that retrieve the assignment units. 
2. (minor) some debug logging message to track available units when a worker requests HITs.

### DIFF
--- a/mephisto/abstractions/providers/mturk/mturk_unit.py
+++ b/mephisto/abstractions/providers/mturk/mturk_unit.py
@@ -162,6 +162,8 @@ class MTurkUnit(Unit):
         requester = self.get_requester()
         client = self._get_client(requester._requester_name)
         hit = get_hit(client, mturk_hit_id)
+        if not hit:
+            return AssignmentState.SOFT_REJECTED
         hit_data = hit["HIT"]
 
         local_status = self.db_status
--- a/mephisto/abstractions/providers/mturk/mturk_unit.py
+++ b/mephisto/abstractions/providers/mturk/mturk_unit.py
@@ -162,8 +162,8 @@ class MTurkUnit(Unit):
         requester = self.get_requester()
         client = self._get_client(requester._requester_name)
         hit = get_hit(client, mturk_hit_id)
-        if not hit:
-            return AssignmentState.SOFT_REJECTED
+        if hit is None:
+            return AssignmentState.EXPIRED
         hit_data = hit["HIT"]
 
         local_status = self.db_status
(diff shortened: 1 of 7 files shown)


mephisto/abstractions/providers/mturk/mturk_unit.py
```diff
@@ -162,6 +162,8 @@ def get_status(self) -> str:
         requester = self.get_requester()
         client = self._get_client(requester._requester_name)
         hit = get_hit(client, mturk_hit_id)
+        if not hit:
```
JackUrb, line 165: Nit: explicitly check for `if hit is None`, as that's the only other value we expect here. Also means that after we add typechecking, not entering this condition will ensure that hit is not none, and is the other type we expect as a return value.

mephisto/abstractions/providers/mturk/mturk_unit.py
```diff
@@ -162,6 +162,8 @@ def get_status(self) -> str:
...
+            return AssignmentState.SOFT_REJECTED
```
JackUrb, line 166: Nit: The more I think about it, the more `AssignmentState.EXPIRED` is more appropriate for this.

mephisto/abstractions/providers/mturk/mturk_utils.py
```diff
@@ -551,7 +551,12 @@ def delete_sns_topic(session: boto3.Session, topic_arn: str) -> None:
 
 def get_hit(client: MTurkClient, hit_id: str) -> Dict[str, Any]:
     """Get hit from mturk by hit_id"""
-    return client.get_hit(HITId=hit_id)
+    hit = None
+    try:
+        hit = client.get_hit(HITId=hit_id)
+    except ClientError:
+        logger.warning(f"Unable to retrieve HIT {hit_id}. Skipping!")
```
JackUrb, line 558: Rather than throw away the client error, include it: 
```
except ClientError as e:
  logger.warning(f"Unable to retrieve HIT {hit_id} due to {e}.  Skipping!")
```

JackUrb (approved):
This kind of change _generally_ makes sense to me. It's not the best behavior for when this actually happens during a live task, but that is an edge case that's hard to deal with.

mojtaba-komeili:
@jju for some reason I ended up with a bunch of formatting tasks that I had to fix to make sure the checks pass.

PR #467, opened 2021-05-13
Idempotent mturk quals

JackUrb:
# Overview
The bugfix in #385 only resolved the step of creating a qualification multiple times in the Mephisto DB, but didn't actually cover the same step for the MTurk side. As such, if two calls to `find_or_create_qualification` happened simultaneously, both may actually trigger the request to create the qualification. 

This change catches the failed call, and then re-queries the MTurk database to get the correct qualification information.

# Testing
Tested by stepping through in a repl, ensuring that the error formatting lined up on a failed create, and then running the full function.


mojtaba-komeili (approved):
Looks good. Thanks!

PR #468, opened 2021-05-24
Context manager in clear_hit_from_unit

bottler:
One fairly new function, clear_hit_from_unit in mturk_datastore, was missing a context manager when it modifies the database. Theoretically, if the update were to fail, this could leak a lock, although I doubt this is actually realistic.

(I am seeing a problem sometimes where workers cannot accept hits on one run after another run has finished, which led me to this. However I don't think this is actually the cause.)


JackUrb (approved):
Thanks for fixing this up (before it had become an issue too).

PR #480, opened 2021-06-11
Static Tasks with Generated Assignments

JackUrb:
# Overview
As a partial step towards model-in-the-loop for enqueuing tasks, and more task control, this PR makes it possible to provide a `Generator` to `SharedStaticTaskState.static_task_data` such that Mephisto can run with an indeterminate amount of tasks, generated on the fly. This can be used in the short term by tasks that intend to get a full set of annotations, but have some quality control on what is considered good.

# Implementation
Much of this was already implemented in an earlier PR, but this finishes a few of the threads:
1. Updates `StaticBlueprint` to wrap the `get_initialization_data` method with a generator-friendly version
2. Notes in the `StaticHTMLBlueprint` that this behavior is not supported (as we inject HTML into the task data, and I'm not really interested in keeping full HTML support as it is an onboarding ramp)
3. Updates the `Operator` to not shut down if there are still assignment generators running for a task.

# Testing
I replaced the `static_react_task`'s `static_task_data` with the following generator:
```
    def task_data_generator():
        for x in range(5):
            time.sleep(5)
            yield {"text": f"This text comes from task number {x}"}
            input()
```
And ensured that only one task was available per newline sent in, and that Mephisto didn't shut down when no tasks were available.

This should have a more deliberate test case, but that will be included in the more complete model-in-the-loop update.

### DIFF
--- a/examples/static_react_task/run_task.py
+++ b/examples/static_react_task/run_task.py
@@ -21,6 +21,7 @@ import hydra
 from omegaconf import DictConfig
 from dataclasses import dataclass, field
 from typing import List, Any
+import time
 
 TASK_DIRECTORY = os.path.dirname(os.path.abspath(__file__))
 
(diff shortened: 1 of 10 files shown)


pringshia (approved):
Great! This is a solid step towards full model-in-the-loop support.

PR #481, opened 2021-06-15
Upgrading boto requirements

JackUrb:
As titled, simply upgrading the version of `boto3` we are using

Doesn't seem to have broken any MTurk tests and manual runs on sandbox look fine.


EricMichaelSmith (approved):
Makes sense

Issue #93, opened 2020-04-23
[TODO Roundup] Logging overhaul


JackUrb:
# Summary
Mephisto doesn't yet have a good system for logging of information. As such we'd like to add loggers via the [python `Logger` class](https://docs.python.org/3/library/logging.html). At a high level, modules should each have their own loggers. Initially we should be able to configure the log level across the whole application (from `DEBUG` to `CRITICAL`), and should be able to write both to terminal and to file. NOTE - the logger by default should be configured to use `NullHandler`, and mephisto should have a setting to toggle this to `StreamHandler` (for writing to terminal) or `FileHandler` (for writing to file).

Some helpful information for logging best practices can be found across the web ([1](https://timber.io/blog/the-pythonic-guide-to-logging/), [2](https://fangpenlin.com/posts/2012/08/26/good-logging-practice-in-python/), [3](https://docs.python-guide.org/writing/logging/)...)

# Requirements:
- [x] Create a `core.logger` class for Mephisto to store any initializers and helper methods required for creating loggers. 
- [x] Add a logger object to every class in the `core`, `data_model`, `server`, and `providers` modules.
- [x] Replace existing print statements with logs, and add additional marked logs.
- [x] Add logging to unintended except blocks
# Missing logs
- [ ] Do a check to see if the database schema is not as expected, and log this to the user.
- [x] Update the wrapper around the call to heroku functions to only surface the "You have too many apps" message if there are actually too many applications, otherwise surface the contained error. This can be tested by creating a heroku account, launching 5 apps without shutting them down, and then launching another.
- [ ] Do propper logging around failed MTurk Approve and Reject calls, described [here](https://github.com/facebookresearch/Mephisto/pull/166/files#r436175412)
# Wishlist
- [x] Peruse the flow of the `Supervisor`, `Operator`, and `TaskLauncher` classes to see if there are important control flow operations that might be improved with logging.
- [x] Add logging to the data model for similar important control flow paths, things that might require inspection. (Essentially, while debugging other parts of Mephisto, when you find the important variable to introspect, add that to `DEBUG` logging. Important control flow goes into `INFO`, and then errors above that).
- [ ] Update the logging in `print_run_details` to actually have useful details about the progress of tasks currently being run.
- [ ] Write `debug` level logging to file, and provide a utility for cleaning up old Mephisto logs.

pringshia:
Does #365 resolve any of the tasks in this?

JackUrb:
[...]

PR #486, opened 2021-06-28
Further clarifying misleading heroku exception.

JackUrb:
As part of resolving heroku logging in #93, adds more context to the failure log to not lead people in the wrong direction.


pringshia:
Should be serviceable as a temporary band-aid for now, with the intention of propagating through underlying exceptions/errors at some point in the future.

pringshia (approved):

jxmsML:
thanks for this quick fix!

PR #489, opened 2021-06-30
Adding ping-pong disconnect to node router

### LINKED ISSUE #485: Probably getting null Worker ID in the frontend
Hi,

Thank you for the great crowdsourcing platform. We are using Mephisto for a chat-style data collection.

When the task was live on AMT, many of the workers reported that they saw the following message:

`Sorry, you have already worked on the maximum number of these tasks available to you, or are no longer eligible to work on this task.`

However, we did not set the property `maximum_units_per_worker` in our config.
We did not face any errors in `localhost` or AMT Sandbox.

From the [code](https://github.com/facebookresearch/Mephisto/blob/c571bf9c2e6395553e71ce922c14a14e363ad7d5/packages/mephisto-task/src/index.js#L109), it looks like this message arises when the `workerId` received is `null`.
Is there any other scenario, apart from exceeding `maximum_units_per_worker`, that can result in a null `workerId`?

JackUrb:
# Overview

The node server running on Heroku hasn't been able to detect disconnects in the same way that it does when running locally. This leads to delays on marking tasks as disconnected, expiring tasks, and generally a mismatch between `Agent` and `Unit` status on MTurk compared to Mephisto.

This PR moves to record ping times on the router to allow it to determine without a hard disconnect whether an agent has disconnected, currently set to not having received a heartbeat in 15 seconds.

Resolves #485.

# Implementation details
- Added `last_ping` to the `LocalAgentState` of the node router, and update it whenever a `HEARTBEAT` is received.
- Check liveliness on the router during the periodic status checks from the main Mephisto server, updating relevant agents to `STATUS_DISCONNECT` when relevant.

Additionally, found and resolved the following bugs:
- Disconnect events which led to expiring tasks would prevent Mephisto from shutting down, even if there were no tasks left (as the task had expired). This is cleaned up by ensuring that, after a `Unit` transitions to `EXPIRED` it cannot move back to `ASSIGNED`.
- Fixed reconnecting to the node router after a disconnect and receiving the correct status

# Testing
Launched a server on Heroku with the parlai chat task, connected to it as two different agents, and then disconnected one. Ensured that, while this disconnect wasn't detected before, it is now.

Launched a static react task to ensure connection functionality on this hadn't changed (as no heartbeats are sent). Connected, and found myself still connected after 15 seconds.

Ensured both tasks shut down properly when completed.

### DIFF
--- a/mephisto/abstractions/architects/router/deploy/server.js
+++ b/mephisto/abstractions/architects/router/deploy/server.js
@@ -56,13 +56,24 @@ const server = http.createServer(app);
 // ======= <Sockets and Agents> ========
 
 const FAILED_RECONNECT_TIME = 10000;
+const FAILED_PING_TIME = 15000;
 
 // TODO can we pull all these from somewhere, make sure they're testable
 // and show they're the same as the python ones?
-const STATUS_INIT = "none";
-const STATUS_CONNECTED = "connected";
-const STATUS_DISCONNECTED = "disconnect";
+const STATUS_NONE = "none";
+const STATUS_ACCEPTED = "accepted";
+const STATUS_ONBOARDING = "onboarding";
+const STATUS_WAITING = "waiting";
+const STATUS_IN_TASK = "in task";
 const STATUS_COMPLETED = "completed";
+const STATUS_DISCONNECT = "disconnect";
+const STATUS_TIMEOUT = "timeout";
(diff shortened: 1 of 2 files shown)


pringshia:
Looks good to me. Thanks for the thorough PR description. I see quite a few new statuses introduced (and in the diff I don't see some of them being used - is that an intentional ommision?)

pringshia (approved):

Issue #488, opened 2021-06-29
Running static_test_script.py: Usage of deprecated keyword in package header


ninadu07:
Seems like this is a compatibility issue with hydra 1.1.0? I've changed static_test_script.py like so:
`@hydra.main(config_path=".", config_name="scriptconfig")`
and deleted the first line config/example.yaml as instructed by hydra.cc, after running the script I'm still getting the following error:
`/home/ninacdu/.local/lib/python3.8/site-packages/hydra/core/default_element.py:122: UserWarning: In 'mephisto/provider/mock': Usage of deprecated keyword in package header '# @package _group_'.
See https://hydra.cc/docs/next/upgrades/1.0_to_1.1/changes_to_package_header for more information
  warnings.warn(
/home/ninacdu/.local/lib/python3.8/site-packages/hydra/core/default_element.py:122: UserWarning: In 'mephisto/architect/local': Usage of deprecated keyword in package header '# @package _group_'.
See https://hydra.cc/docs/next/upgrades/1.0_to_1.1/changes_to_package_header for more information
  warnings.warn(
/home/ninacdu/.local/lib/python3.8/site-packages/hydra/core/default_element.py:122: UserWarning: In 'mephisto/blueprint/static_task': Usage of deprecated keyword in package header '# @package _group_'.
See https://hydra.cc/docs/next/upgrades/1.0_to_1.1/changes_to_package_header for more information
  warnings.warn(
In 'scriptconfig': Validation error while composing config:
Merge error: BlueprintArgs is not a subclass of StaticHTMLBlueprintArgs. value: {'_blueprint_type': '???', 'onboarding_qualification': '???', 'block_qualification': '???'}
    full_key: 
    object_type=dict

Set the environment variable HYDRA_FULL_ERROR=1 for a complete stack trace.
`
Is there something I'm doing wrong here?

JackUrb:
I hadn't realized 1.1.0 was out yet - excited to upgrade, but for now Mephisto isn't compatible yet.

ninadu07:
I downgraded hydra to 1.0.6, and after running the script the error I got is as follow:
`[2021-06-28 18:42:20,597][mephisto.operations.operator][ERROR] - Ran into error while launching run: 
Traceback (most recent call last):
  File "/media/ninacdu/Workspace/convqa/Mephisto/mephisto/operations/operator.py", line 424, in validate_and_run_config
    return self.validate_and_run_config_or_die(
[...]

PR #490, opened 2021-06-30
Specifying the SharedTaskState used by a specific blueprint

JackUrb:
# Overview
The fixes introduced in #480 to allow for static tasks to have generated assignments on the fly introduced a bug when the `SharedTaskState` was not directly specified. By default, `Operator`s were creating a default `SharedTaskState` rather than the `SharedTaskState` that a specific `Blueprint` might specify.

This was surfaced in #488 

# Fix
This simply ensures that `SharedStateClass` is set on all of our blueprints correctly, and asks the `Operator` to generate the correct `SharedTaskState` using this attribute on startup.


pringshia (commented):
lgtm

PR #493, opened 2021-07-06
Add a Storybook for the annotation-toolkit

pringshia:
## Overview 

As we continue to build out components for the annotation-toolkit, it would be great to have a way to showcase them to users.

This PR lays the foundation of the annotation-toolkit storybook, also found deployed here: **https://annotation-toolkit-storybook.vercel.app/**

This also serves as a great way to document common use-cases. By looking at the code for individual stories, users can reference common recipes for getting up and running.

In the future, we can also implement ways to do regression testing using these stories, either via screenshot capture or re-leveraging the stories for base case jest assertions.

## Screenshots

Screen Shot 2021-07-05 at 11 49 07 PM
Screen Shot 2021-07-05 at 11 49 38 PM


JackUrb:
This storybook presentation is pretty compelling, especially with the docs included. Would be interesting to have automated testing similar to this, but this fast deploy is an improvement.

JackUrb (approved):

docs/storybook/src/stories/Button.jsx
```diff
@@ -0,0 +1,54 @@
+import React from "react";
+import PropTypes from "prop-types";
+import "./button.css";
+
+/**
+ * Primary UI component for user interaction
+ */
+export const Button = ({ primary, backgroundColor, size, label, ...props }) => {
```
JackUrb, line 8: Just for my understanding - is this part of the storybook demo, or a component we're actually displaying?

docs/storybook/src/stories/Header.jsx
```diff
@@ -0,0 +1,62 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+import { Button } from "./Button";
+import "./header.css";
+
+export const Header = ({ user, onLogin, onLogout, onCreateAccount }) => (
```
JackUrb, line 7: Same comment as with `Button`

docs/storybook/src/stories/Button.jsx
pringshia: yep, this is from the demo. it's not being displayed in the UI, but I sort of like to have it around as I'm learning the features of storybook, it makes for a good reference, particularly the `Button.stories.jsx` file

PR #509, opened 2021-08-09
Enable symlinking js bundle via Hydra prop to make for easier local development for static_react_task

pringshia:
## Overview

@ytsheng mentioned that it's annoying that `python run_task.py` requires a rerun every time a React code change is made. I agreed.

This PR update introduces a Hydra config property called `mephisto.blueprint.link_task_source` which can be set to `true`, to remediate this.`

From the accompanying README update in this PR:

> #### Local Development
>
> For local development, you may want to have changes made to your React code reflect locally without having to restart the server each time. To enable this, update the Hydra config's `mephisto.blueprint.link_task_source` value to `true` (default is `false`).
> 
> After running `python run_task.py`, you can then run `npm run dev:watch` in the webapp folder to auto-regenerate the task_source file. Since the task source file is now symlinked, simply refreshing the browser will show changes. You won't need to kill and restart the server each time anymore.

## Testing

1. Update `/examples/static_react_test/hydra_configs/conf/example.yaml` to include `link_task_source: true`: 
Screen Shot 2021-08-09 at 7 03 22 PM

2. Run `python run_task.py`
3. Run `npm run dev:watch` from the `/examples/static_react_test/webapp` folder in a separate console
4. Load the page at http://localhost:3000
5. Edit something in `/examples/static_react_task/webapp/src/app.jsx` for example, in the `isPreview` code block where the task description is written
6. Reload the page at http://localhost:3000 without restarting the python server
7. Ensure that the change is reflected

### DIFF
--- a/mephisto/abstractions/blueprints/abstract/static_task/static_blueprint.py
+++ b/mephisto/abstractions/blueprints/abstract/static_task/static_blueprint.py
@@ -87,15 +87,6 @@ class StaticBlueprintArgs(BlueprintArgs):
     data_csv: str = field(
         default=MISSING, metadata={"help": "Path to csv file containing task data"}
     )
-    extra_source_dir: str = field(
-        default=MISSING,
-        metadata={
-            "help": (
-                "Optional path to sources that the HTML may "
-                "refer to (such as images/video/css/scripts)"
-            )
-        },
-    )
 
 
 class StaticBlueprint(Blueprint, OnboardingRequired):
--- a/mephisto/abstractions/blueprints/static_react_task/static_react_blueprint.py
+++ b/mephisto/abstractions/blueprints/static_react_task/static_react_blueprint.py
@@ -61,9 +61,9 @@ class StaticReactBlueprintArgs(StaticBlueprintArgs):
         default=False,
         metadata={
             "help": """
-                Symlinks the the task_source file in your development folder to the
+                Symlinks the task_source file in your development folder to the
                 one used for the server. Useful for local development so you can run
-                a watch-based build for your task_source so that the UI code can
+                a watch-based build for your task_source, allowing the UI code to
                 update without having to restart the server each time.
             """,
             "required": False,
--- a/mephisto/abstractions/blueprints/static_react_task/static_react_blueprint.py
+++ b/mephisto/abstractions/blueprints/static_react_task/static_react_blueprint.py
@@ -100,7 +100,15 @@ class StaticReactBlueprint(StaticBlueprint):
         assert (
             found_task_source is not None
         ), "Must provide a path to a javascript bundle in `task_source`"
+
         found_task_path = os.path.expanduser(found_task_source)
         assert os.path.exists(
             found_task_path
         ), f"Provided task source {found_task_path} does not exist."
+
+        link_task_source = args.blueprint.link_task_source
+        current_architect = args.architect._architect_type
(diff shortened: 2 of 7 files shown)


mephisto/abstractions/blueprints/abstract/static_task/static_blueprint.py
```diff
@@ -87,15 +87,6 @@ class StaticBlueprintArgs(BlueprintArgs):
...
     )
-    extra_source_dir: str = field(
-        default=MISSING,
-        metadata={
-            "help": (
-                "Optional path to sources that the HTML may "
-                "refer to (such as images/video/css/scripts)"
-            )
-        },
-    )
```
pringshia, line 98: This is a duplicate field, it's already defined higher up. Cleaning up.

ytsheng (approved):
Confirmed that it works on my end. Thank you!!!

JackUrb (commented):
I feel we should do a bit more to prevent accidental mismatch between input configuration arguments. If someone tries to launch externally with this option set, the behavior may not be as expected. For this, you can add a case to `StaticReactBlueprint.assert_task_args` that throws if launching with `cfg.architect._architect_type == heroku` and this option set to true. 

In the longer run, I think we'll need to dig deeper on compatibility _between_ the different interfaces, but I don't want to block this implementation for that discussion.

JackUrb (approved):
Approved pending the compatibility changes.

pringshia:
Makes sense @JackUrb - I was thinking the same thing, as this config option only really makes sense for when architect=local and for blueprint=static_react_task, even though it lives within the blueprint config.

I wasn't sure whether to silently ignore the irrelevant config when architect != local or whether to fail. I don't have a strong opinion either way, happy to add in the check in `assert_task_args`. 

I think it may make more sense to have it be `cfg.architect._architect_type != local` vs `cfg.architect._architect_type == heroku`. any thoughts on that? ofc right now we only have 2 but this will make it so we dont have to change it again in the future if we add something like ec2

JackUrb:
I was wondering if there would ever be a situation where we want to use `_architect_type == mock` for this implementation with the above suggestion. Indeed we only have a few right now, perhaps the safer option is to gate to local for now.

pringshia:
`mock` could be a candidate in the future. for now, I'll assert that it only works for `local` (i prefer explicit allowlists over inexplicit blocklists), and in the future as we test and integrate `mock` we can add that to the allowlist as well

mephisto/abstractions/blueprints/static_react_task/static_react_blueprint.py
```diff
@@ -100,7 +100,15 @@ def assert_task_args(
...
         assert os.path.exists(
             found_task_path
         ), f"Provided task source {found_task_path} does not exist."
+
+        link_task_source = args.blueprint.link_task_source
+        current_architect = args.architect._architect_type
+        allowed_architects = ["local"]
+        assert link_task_source == False or (
+            link_task_source == True and current_architect in allowed_architects
+        ), f"`link_task_source={link_task_source}` is not compatible with architect type: {args.architect._architect_type}. Please check your task configuration."
```
pringshia, line 114: Added task args validation to ensure `link_task_source` is not accidentally specified with something like `heroku`

JackUrb (approved):
Thanks for adding the check!

PR #534, opened 2021-09-01
[fix] make frame height to be 0

ytsheng:
https://blog.mturk.com/tasks-can-now-scale-to-a-workers-browser-window-size-c6e66f4bdfc9

Making frame height to be zero maximizes the annotation area. Please see screenshot: 

after copy


JackUrb:
@ytsheng I updated this PR to write the frame height to the runs table on creation, this way we can respect tasks that define specific heights in their `get_frontend_args()` functions. Some downstream tasks have been built to 650, so this should satisfy both. Let me know if you have any thoughts.

Issue #536, opened 2021-09-01
Problem in get_reviewable_task_runs() about jsonify DictConfig 


wayne163:
I came with a problem when using client scripts. I think maybe it's a type conflict bug. 
Version: 0.3.4
Steps
1. start app with ```python cli.py web```
2. send get request to ```http://localhost:5000/api/v1/task_runs/reviewable``` and ten flask will throw error
```
[2021-09-01 12:54:36,582] ERROR in app: Exception on /api/v1/task_runs/reviewable [GET]
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python3.8/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/Mephisto/mephisto/client/api.py", line 68, in get_reviewable_task_runs
    return jsonify({"task_runs": dict_tasks, "total_reviewable": reviewable_count})
  File "/usr/local/lib/python3.8/site-packages/flask/json/__init__.py", line 370, in jsonify
    dumps(data, indent=indent, separators=separators) + "\n",
  File "/usr/local/lib/python3.8/site-packages/flask/json/__init__.py", line 211, in dumps
    rv = _json.dumps(obj, **kwargs)
  File "/usr/local/lib/python3.8/json/__init__.py", line 234, in dumps
    return cls(
  File "/usr/local/lib/python3.8/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/local/lib/python3.8/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/local/lib/python3.8/site-packages/flask/json/__init__.py", line 99, in default
    return text_type(o.__html__())
[...]

PR #547, opened 2021-09-06
fix task_config type conflict bug(#536)

wayne163:
fix bug issue [536](https://github.com/facebookresearch/Mephisto/issues/536)

### DIFF
--- a/mephisto/data_model/task_config.py
+++ b/mephisto/data_model/task_config.py
@@ -108,7 +108,7 @@ class TaskConfig:
             return
 
         # Parse out specific arguments for the task_config
-        self.args: Dict[str, Any] = args.task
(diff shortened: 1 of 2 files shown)


JackUrb:
Hi @wayne163, thanks for the PR! Though this doesn't actually take the approach you described in the issue. We should expect that `TaskConfig.args` will be of type `DictConfig`, and then containerize it during the write out in `to_dict`. Happy to take this over if that's unclear though!

wayne163:
> to_dict

Hi @JackUrb , sorry I misunderstood before. I have updated this PR,  set task_config.args as DictConfig Type not Dict.

JackUrb (approved):
Thanks for making the changes here @wayne163! This looks great 👍

PR #548, opened 2021-09-07
Creating required files for pypi release

### LINKED ISSUE #524: Release Mephisto versions on PyPI
Right now the suggested approach for getting Mephisto running is cloning the repo and running `pip install -e .` 

For better or for worse, this approach requires users to "run on the edge" and thus breaking changes may be easily introduced into their projects. For example, in this case from one of Mephisto's dependencies: #508

Although we are mindful to use git tags to mark releases, it would be good to also publish versions of Mephisto on PyPI so that users can more explicitly select and upgrade their Mephisto runtime.

Starting with the 1.0 release, we could enable `pip install Mephisto`

JackUrb:
# Overview
In order to get PyPI releases working, this sets up the minimum packaging setup that also includes all of the static files we require to run Mephisto. Once this is merged I'll release `0.4.1`, which should be our first full PyPI release.

# Concerns & Future work
Right now I've set this up to be able to upload via the classic `setup.py` route, rather than using `poetry` as we have tried to for other parts of the project. I think this signals a degradation of the developer experience, however I couldn't see a quick resolution that allows us to use `pip install -e .` for local development using just poetry before. Digging in, it appears to be blocked by pip here: https://github.com/pypa/pip/issues/6950, and should be resolved by [PEP 660](https://www.python.org/dev/peps/pep-0660/) which is targeted for pip 21.3 (October). We should revisit consolidating around poetry then.

Closes #524


pringshia (approved):
Looks reasonable to me. Thanks for getting this key piece together.

PR #554, opened 2021-09-09
Fixing parlai chat height bug

JackUrb:
# Summary 
Moving to 0 chat height in #534 broke ParlAI chat-based tasks. In going through I found our order of override resolution to be somewhat strange for this task, so I clarified. 

# Testing
Launching a parlai chat task before this change launched with 0 height and was invisible. Afterwards I get 650 height, though after the bootstrap-chat change one could override with `SharedTaskState` arguments to be able to provide a `frame_height` of 0 to get full browser height on MTurk.

### DIFF
--- a/mephisto/abstractions/blueprints/parlai_chat/parlai_chat_blueprint.py
+++ b/mephisto/abstractions/blueprints/parlai_chat/parlai_chat_blueprint.py
@@ -286,16 +286,19 @@ class ParlAIChatBlueprint(Blueprint, OnboardingRequired):
         to the client for use by the task's frontend
         """
         # TODO move frontend args in
-        frontend_task_config = {
-            "task_description": self.full_task_description,
-            "preview_html": self.full_preview_description,
-            "frame_height": 0,
-            "chat_title": self.args.task.task_title,
-            "has_preview": self.args.blueprint.get("preview_source", None) is not None,
-            "block_mobile": True,
-            "frontend_task_opts": self.shared_state.frontend_task_opts,
-        }
-        frontend_task_config.update(super().get_frontend_args())
+        frontend_task_config = super().get_frontend_args()
+        frontend_task_config.update(
+            {
+                "task_description": self.full_task_description,
+                "preview_html": self.full_preview_description,
+                "frame_height": 650,
+                "chat_title": self.args.task.task_title,
+                "has_preview": self.args.blueprint.get("preview_source", None)
+                is not None,
+                "block_mobile": True,
+                "frontend_task_opts": self.shared_state.frontend_task_opts,
+            }
+        )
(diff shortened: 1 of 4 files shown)


pringshia (approved):
LGTM

mephisto/abstractions/blueprints/parlai_chat/parlai_chat_blueprint.py
```diff
@@ -285,17 +285,23 @@ def get_frontend_args(self) -> Dict[str, Any]:
...
+        # Start with standard task configuration arguments
+        frontend_task_config = super().get_frontend_args()
+        # Add ParlAI standards
+        frontend_task_config.update(
+            {
+                "task_description": self.full_task_description,
+                "preview_html": self.full_preview_description,
+                "frame_height": 650,
+                "chat_title": self.args.task.task_title,
+                "has_preview": self.args.blueprint.get("preview_source", None)
```
pringshia, line 297: this is an intentional change right? whereas before it would enforce a boolean this may let non-boolean values flow through? could this cause breakages?

packages/bootstrap-chat/src/composed/BaseFrontend.jsx
```diff
@@ -35,10 +35,12 @@ function BaseFrontend({
   const { connectionStatus, agentStatus, taskConfig } = mephistoContext;
   const { appSettings } = appContext;
   const sidePaneSize = appSettings.isCoverPage ? "col-xs-12" : "col-xs-4";
+  const heightStyle =
+    taskConfig.frame_height == 0 ? {} : { height: taskConfig.frame_height };
```
pringshia, line 39: i generally like to use `===` triple equals. for example:

```
false == 0 // true
false === 0 // false

```

mephisto/abstractions/blueprints/parlai_chat/parlai_chat_blueprint.py
JackUrb: Ah it's on the next line - `black` formatting decided to make this less clear unfortunately

packages/bootstrap-chat/src/composed/BaseFrontend.jsx
JackUrb: Interesting, I hadn't considered this when I wrote the code, but think that the implicit conversions in this case are actually reasonable.

mephisto/abstractions/blueprints/parlai_chat/parlai_chat_blueprint.py
pringshia: ah lol I see it now... interesting choice `black`...

PR #555, opened 2021-09-10
Save ParlAI chat start/end times

EricMichaelSmith:
Use existing `AgentState` methods to track the start and end times of ParlAI chat HITs


JackUrb (approved):
Great use of this interface!

PR #556, opened 2021-09-13
[ParlAI] Pushing timestamps into the messages too

JackUrb:
Like #555, I think it's relevant to have timestamps in the messages themselves. Currently using this in a review script for LIGHT that I think is really a useful metric for automated evals. 

@EricMichaelSmith this will probably be a key that you'll have to drop as well in your PR that can clean this data for test purposes.


EricMichaelSmith (approved):
Makes sense, good to have!

PR #566, opened 2021-09-18
Refactor Blueprint Mixins, create ScreenTaskRequired mixin

JackUrb:
# Overview
Creates the first full-fledged `Blueprint` mixin `ScreenTaskRequired`, in a similar vein to `OnboardingRequired`, but with clearer delineation for workflow-specific functionality. This PR also refactors both of these to follow the new `BlueprintMixin` structure, which lets us standardize how workflows for blueprints can be mixed in. The goal here is to create a task that, for workers that haven't completed any of the task yet, you can put out a test task for which you can more easily validate the results to than a full task. This launches new tasks on the fly, such that you can have validation tasks be separate from the rest of the pool.

# Implementation
- Creates the `ValidationRequired` mixin, which defines behaviors to properly validate and initialize blueprints with this functionality. Defines some logic hooks that Mephisto's ops layer can use to interact with this workflow. Also defines utility methods that are useful in user scripts for those utilizing this workflow.
- Updates `TaskRun` to ignore special units (for gold, validators, etc) when determining eligible units for workers
- Passes the `TaskLauncher` along to the `Supervisor` to be able to handle launching validator units. This _smells_ but I aim to address this kind of over-coupling in the coming refactor of the `Supervisor`.
- Updates the `Supervisor` to run validation tasks for new workers on blueprints that have `ValidationRequired`. Ensures that these are expired when abandoned, to not pass to others.
- Creates a function in the `TaskLauncher` that creates and launches the special validator units.

# Considerations
- This implementation is just getting things working, but really displays the need for a general supervisor refactor. Moving some of the business logic into blueprints and feature mixins will likely help encapsulate things.
- This introduces a specific way to determine if launched tasks are used for special purposes, keyed on the `unit_index`. I think this is a _decent_ way to go about this, but I wonder if mixins should also have considerations for review scripts? Likely need to update the `DataBrowser` to handle these better in the future.
- Right now I've dumped the mixins with the `Blueprint` file, but I think these should go somewhere better. Will break out before merging

# Testing
In internal
- [x] Fix automated testing broken by passing task_launcher to supervisor
- [x] write automatic testing for this new blueprint mixin


pringshia:
Overall looks good! Will re-review pending some of the changes we discussed, e.g. adding max_validations @ the blueprint level, as well as perhaps asserting that this value is necessary to re-emphasize to the user that there is a cost for these tasks

mephisto/operations/task_launcher.py
```diff
@@ -198,11 +200,46 @@ def _launch_limited_units(self, url: str) -> None:
...
+            task_run.requester_id,
+            assignment_id,
+            SCREENING_UNIT_INDEX,
+            task_config.task_reward,
+            task_run.provider_type,
+            task_run.task_type,
+            task_run.sandbox,
+        )
+        validator_unit = Unit.get(self.db, unit_id)
+        validator_unit.launch(self.launch_url)
```
pringshia, line 240: shall we rename this to `screening_unit` as well?

test/core/test_supervisor.py
```diff
@@ -846,6 +853,185 @@ def test_register_job_with_onboarding(self):
...
+            if worker_2.is_qualified(FAILED_QUALIFICATION_NAME):
+                break
+            time.sleep(0.1)
+        self.assertLess(
+            time.time() - start_time, TIMEOUT_TIME, "Did not qualify in time"
+        )
+
+        # Submit screening from the agent
+        validation_data = {"success": True}
+        self.architect.server.send_agent_act(agents[2].get_agent_id(), validation_data)
```
pringshia, line 1006: rename these `validation_data` variables as well

mephisto/abstractions/blueprints/mixins/screen_task_required.py
```diff
@@ -0,0 +1,202 @@
...
+            failed_qualification_name is not None
+        ), "Must supply an block_qualification in Hydra args to use a qualification task"
+        assert hasattr(shared_state, "generate_screening_unit_data"), (
+            "You must supply a generate_screening_unit_data function in your SharedTaskState to use "
+            "qualification tasks."
+        )
+        max_screening_units = args.blueprint.max_screening_units
+        assert max_screening_units is not None, (
+            "You must supply a blueprint.max_screening_units argument to set the maximum number of "
+            "additional tasks you will pay out for the purpose of validating new workers. "
```
pringshia, line 139: I would rephrase "validating new workers." => "screening new workers." here as well to keep terminology consistent.

This would also be a good place to remind workers of the charges associated.

We can probably add to this to the end, similar to the README: "Note that you **do pay** for screening units, they are just like any other units."

Unrelated: but I see units and tasks being used quite interchangeable in the docs.

pringshia (approved):
I have a few copy and terminology related notes, but otherwise this looks good to me.

mephisto/abstractions/blueprints/README.md
```diff
@@ -31,6 +31,14 @@ A blueprint is able to create a container that handles any shared data that is i
...
+### `OnboardingRequired`
+This mixin allows for blueprints that require people to complete an onboarding task _before_ they're even able to start on their first task. Usually this is useful for providing task context, and then quizzing workers to see if they understand what's provided. Tasks using this mixin will activate onboarding mode for new `Worker`s whenever the `mephisto.blueprint.onboarding_qualification` hydra argument is provided.
+### `ScreenTaskRequired`
+This mixin allows for blueprints that require people to complete a _test_ version of the real task the first time a worker does the task. This allows you to validate workers on a run of the real task, either on your actual data (when providing `SharedTaskState.generate_screening_unit_data=False`) or on test data that you may more easily validate using (when providing a generator to `SharedTaskState.generate_screening_unit_data`). The tasks should be the same as your standard task, just able to be easily validated. You **do pay** for screening tasks, and as such we ask you set `mephisto.blueprint.max_screening_units` to put a cap on how many screening units you want to launch.
```
pringshia, line 39: Really like this API!

`screening_data_factory` could be another possible name here... just throwing out ideas

mephisto/abstractions/blueprints/mixins/screen_task_required.py
JackUrb: Yes, I need to bring the terminology to units across everywhere that makes sense! Will also catch the remaining validation name changes

PR #570, opened 2021-09-20
Add support for ref based state-retrieval

### LINKED ISSUE #521: Add on-demand state retrieval via a ref from `global-context-store`'s `<Store />`
### Overview

In certain integration use-cases, we could have a container app that needs to handle the submission for an embedded annotation-toolkit app. 

On submission, we need to allow the parent or sibling component to get access to the state of the `annotation-toolkit` application.

### Proposal

Our proposal for doing so is via React refs. This will allow on-demand state-retrieval functionality using `forwardRef` and `useImperativeHandle`.

### Motivating Use Case

We'd like to be able to do something like the below where on an event such as a button click, we can retrieve the current state of the annotation-toolkit app.

```jsx

function handleSubmit(state) {
  console.log(state)
}

function ParentApp() {
  const appRef = useRef();
  return (
    <div>
      <button onClick={ () => handleSubmit(appRef.current.getState()) } />
      <EmbeddedApp ref={appRef} />
    </div>
  );
}
```

### Design & Considerations

pringshia:
Closes #521

More details in the issue, but essentially global application state can now be retrieved imperatively via the container application using a `ref`. This should make integrations easier.

https://user-images.githubusercontent.com/425059/134744276-a4a8db10-5030-468f-939d-cbbcad161983.mp4

### DIFF
--- a/.github/workflows/annotated-pkgs.yml
+++ b/.github/workflows/annotated-pkgs.yml
@@ -37,5 +37,8 @@ jobs:
       - name: Build packages
         run: yarn workspaces foreach -ptA run build
 
+      - name: Build packages with legacy setup
+        run: cd packages/global-context-store && npm install && npm run build
+
       - name: Test that storybook compiles
         run: yarn workspace sb build-storybook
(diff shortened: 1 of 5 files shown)


JackUrb (approved):
Thanks for waiting for this review, needed to take a moment to grasp the semantics. Looks good though! Definitely will be helpful to pull the state out where necessary.

Issue #563, opened 2021-09-17
Connection between the onboarding world and the main world 


ninadu07:
During local development, after turkers complete the onboarding tasks and world_data is submitted, the two turkers are connected into the main world for the task. However, after around 15 seconds, the right-bottom response input pane always changes to 
`<div id="response-type-done" class="response-type-module" style="padding: 20px 25px; float: left;"></div> `
and the input pane is hidden/removed from the chat pane.  

This happens every time the turkers complete onboard tasks (with a new onboarding_qualification string) then connect to the main world. With the same onboarding_qualification, if the turkers have gotten onboard previously, then they go straight to the main world and the input pane works fine. 

Any ideas what am I missing here?  Thanks in advance!

JackUrb:
This definitely sounds like a bug - will investigate, thanks for flagging!

kushalchawla:
Hi. I am facing something similar. I also observe this in the example parlai chat demo script when run with the conf=onboarding_example:

`python parlai_test_script.py conf=onboarding_example`


 The behavior is a bit odd: 

At the end of the onboarding parley, we send a final observe to the agent with episode_done=true ([Line 20 here](https://github.com/facebookresearch/Mephisto/blob/dfc985716ddaf552ff57d179d764bfb5d311edad/examples/parlai_chat_task_demo/demo_worlds.py#L20)). Once that happens, ideally I would expect the message box to become inactive and the control would wait to get connected to the partner. But instead, the message box is active and disappears in ~15 seconds. In the meantime that it does not disappear, if I mess around by sending another message, that message then actually interferes with the upcoming chat that happens when the two partners are paired up.

kushalchawla:
Hello. I did some digging. 

This is probably due to [Line 190-209 here](https://github.com/facebookresearch/Mephisto/blob/dfc985716ddaf552ff57d179d764bfb5d311edad/mephisto/abstractions/blueprints/parlai_chat/parlai_chat_task_runner.py#L190).

After calling parley, there is one more round of observe and act. I am not sure why this is the case, even for the onboarding session.

[...]

PR #573, opened 2021-09-22
Fixing some of the ParlAI problems

JackUrb:
# Overview
Begins work to resolve the issues raised in #563. In short, an issue there was surfaced wherein after transitioning from the onboarding world to the main world, after some number of seconds, the chat window would drop the ability to send messages (including inside of a task world!). This would make it impossible to complete the task.

Also raised in that issue is how, after completing onboarding, the input pane is available for entering a message, and messages sent in this waiting stage end up polluting what ends up in the task.

# Fixes
As of now, this PR addresses only the first part of this. The issue was that messages were still being sent on behalf of the onboarding world, which, after it closed, would then send a message to the frontend client noting the world was closed. This PR resolves the bug by closing the agent thread when onboarding is submitted (preventing the connection from the router to the client) and preventing sending a state change on onboarding agents when the status transitions from waiting to approved or rejected (as neither should be sent to the router).

# Fixes pending
~~@pringshia I've been unable to understand why the second behavior still occurs, and haven't been able to debug either. It's reproducible just by running the parlai chat conf with onboarding (`python parlai_test_script.py conf=onboarding_example`). Then go to visit onboarding (`http://localhost:3000/?worker_id=<PutNewIDHereEveryTest>&assignment_id=5775`), send a message, and observe that the chat ends up in the active `ready_for_input` state. This doesn't make sense to me, as in the `mephistoContext` it's easy to observe that the intended state is `waiting` and `wants_act` is `false`.~~

This PR now includes updates to these packages, and now works when linked.


pringshia:
@JackUrb

Is this the intended behavior:

https://user-images.githubusercontent.com/425059/134507801-82c8f1d0-d38d-4502-a204-be77bf9b396d.mp4

I was initially able to reproduce the issue you mentioned, however once I used a custom bundle with `bootstrap-chat` connected via `npm link`, the issue seemed to "resolve" and I get what is recorded above.

JackUrb:
Yup that's the desired behavior! Perhaps I fixed this in the code in the past, but never pushed a new version? I guess if we merge this with a version bump on `bootstrap-chat` and `mephisto-task` we'll be good to go.

pringshia (approved):
Looks good to me!

PR #587, opened 2021-10-15
Move defaults to yaml configs for example tasks

meganung:
Moved configs from run scripts to yaml config files for the example tasks. Decided to update them all after realizing I needed to update the defaults for the parlai_chat_task_demo so that the chat_demo task test passes in ParlAI.

Checked that the configs set are respected by running the run scripts with the flag [ --cfg job ](https://hydra.cc/docs/advanced/defaults_list/#debugging-the-defaults-list)

Note: the crowsourcing chat_demo test in ParlAI will fail until the corresponding [PR](https://github.com/facebookresearch/ParlAI/pull/4095) has also been merged (so we should wait until that PR is ready to go). The update changes the way overrides occur within the configs.

### DIFF
--- a/examples/static_react_task/hydra_configs/conf/example.yaml
+++ b/examples/static_react_task/hydra_configs/conf/example.yaml
@@ -1,4 +1,8 @@
 #@package _global_
+defaults:
+  - /mephisto/blueprint: static_react_task
+  - /mephisto/architect: local
+  - /mephisto/provider: mock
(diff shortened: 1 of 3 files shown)


JackUrb (approved):
Thanks for these updates @meganung - I think the new format is easier to reason about and cutting some of the boilerplate out of the run files is nice.

Issue #598, opened 2021-11-02
Restore `mypy` checking for python code


JackUrb:
# Overview
A number of simple issues that have cropped up recently would have been easily resolved with `mypy` being active. We had it running in `poetry` before the PyPI patch, but moving from it has proven problematic. (#593, #596, #597 in recent memory).

PR #601, opened 2021-11-03
Mypy backlog cleanup

JackUrb:
# Overview

Running through all of the mypy issues, as an initial start to #598 and to close some of the nearby issues. Mostly involved just telling `mypy` what was really going on in certain places, but actually did clarify logic and catch a minor bug in onboarding initialization.


pringshia:
Nicely done

pringshia (approved):

Issue #531, opened 2021-08-26
Improve MTurk credential override UX


JackUrb:
# Overview
At the moment, when a user registers an MTurk requester with a profile name that already has credentials for it, we treat this as a no-op even if the credentials don't match those that exist. This is an unexpected user experience, and requires the user manually going to the credentials file and purging the old credentials if they want to be able to update to newer ones. We'd like to improve this behavior.

More discussion in https://github.com/facebookresearch/Mephisto/pull/528

# Implementation
The basic idea would be to replace the warning with code that actually updates to the new credentials. For this, we could parse the `~/.aws/credentials` file, searching for the line that contains the formatted credentials that would be stored. The formatting can be seen later in the file:
https://github.com/facebookresearch/Mephisto/blob/3665e05088ffa8b556d7d1228cc8542670e076b8/mephisto/abstractions/providers/mturk/mturk_utils.py#L104-L110
Here we would parse and check to see if the provided keys match the ones that are already present. If so, this is a no-op. If they don't match, we should print a warning noting that we're updating the credentials, overwriting ones that already existed for the profile, delete those lines from the file, and then append new ones.

Yash621:
@JackUrb can I work on this ?

JackUrb:
Completed in #616

PR #602, opened 2021-11-09
added implementation

Yash621:
@JackUrb I have added the implementation in the mentioned file ,please review it ,if it is okay or not.

### DIFF
--- a/examples/static_react_task/README.md
+++ b/examples/static_react_task/README.md
@@ -119,3 +119,6 @@ In order to get started on your own task, it is a good idea to copy this `static
 7. Launch a small batch on a crowd provider to see how real workers handle your task.
 8. Iterate more.
 9. Collect some good data.
+
+
+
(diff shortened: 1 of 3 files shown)


Yash621:
@facebook-github-bot I have signed the Contributor License Agreement

JackUrb:
Hi @Yash621 - thanks for opening the PR. 

For a few pieces of advice on contributions, it's generally helpful to link to the issue you're trying to solve in your PR details (as you've asked about helping in a few areas). I'm assuming this is related to #531? 

It's also useful to explain your implementation thinking: noting the problem you're trying to solve and how your code provides the solution.

In this case, I don't think you've taken in the meat of what is important for this implementation. Taking details from #531:
> Here we would parse and check to see if the provided keys match the ones that are already present. If so, this is a no-op. If they don't match, we should print a warning noting that we're updating the credentials, overwriting ones that already existed for the profile, delete those lines from the file, and then append new ones.


To be clearer on what's required for this implementation:
1) The first block of `setup_aws_credentials` (wherein we just return `True` if there are credentials for the given profile name and we're able to make a `Session`) is the key place to work in, specifically in the part with the comment "Eventually we could manually re-parse the file and see...".
2) If a profile exists, we need to parse the file and find the details that are present. If they match the ones we would be inserting, this is a no-op and we return True.
3) If the existing profile doesn't match the registration details we are providing, we should _edit_ the file to put the new ones in, and then warn the user that such a change occurred (you can use the same `logger.warning` call

Yash621:
@JackUrb okay i will keep in mind all the points that you mentioned ,thank for the guidance :)

PR #604, opened 2021-11-11
Improved MTurk credential override UX 

Yash621:
@JackUrb This PR is respect to issue #531
As you suggested what I have done is I am accessing the aws_access_key_id present in the credentials folder ,if it matches i consider it a no-op and straight away return true 
if it doesn't I overwrite the credentials and print the warning message with the help of logger.warning 
Please review my PR ,and suggest changes if required.


mephisto/abstractions/providers/mturk/mturk_utils.py
```diff
@@ -67,15 +67,39 @@ def setup_aws_credentials(
...
-            )
+            aws_credentials_file_path = "~/.aws/credentials"
+            expanded_aws_file_path = os.path.expanduser(
+                aws_credentials_file_path)
+            aws_credentials_file_string = ""
+            if os.path.exists(expanded_aws_file_path):
+                with open(expanded_aws_file_path, "r") as aws_credentials_file:
+                    aws_credentials_file_string = aws_credentials_file.read()
+            # accessing the aws_credentials_file
+            aws_access_key_id = aws_credentials_file_string.split("\n")[1]
```
JackUrb, line 78: This won't get you the access key ID always, as the access key ID isn't always on the first line. If you add multiple profiles, there may be many lines to search through, and it will be hard to select the correct one by directly indexing.

Your best bet here is to do a regex search for:
```
"[<profile_name>]\naws_access_key_id=$1\naws_secret_access_key=$2\n
```
And see if the values at $1 and $2 match the given `aws_access_key_id` and `aws_secret_access_key`. If they match, you can just return True. If they don't, you can do a `aws_credentials_file_string.replace(<found_version>, <new_version>)`.

JackUrb (changes requested):
Closer, but not quite there yet! A few simple test cases you can use to ensure it's correct:
```
# Write first credentials to file
setup_aws_credentials('abc', OmegaConf.create({'access_key_id':'test1', 'secret_access_key':'test2'}))
# check the result file, it should have this profile

# Put a second profile
setup_aws_credentials('def', OmegaConf.create({'access_key_id':'test1', 'secret_access_key':'test2'}))
# The credentials file should have two profiles

# Put a third profile
setup_aws_credentials('ghi', OmegaConf.create({'access_key_id':'test1', 'secret_access_key':'test2'}))
# The credentials file should have three profiles

# Repeat the second profile
setup_aws_credentials('def', OmegaConf.create({'access_key_id':'test1', 'secret_access_key':'test2'}))
# This shouldn't warn about anything, as the credentials remain the same. The profile should look the same.

# Repeat the second profile again, but with differet params
setup_aws_credentials('def', OmegaConf.create({'access_key_id':'updated', 'secret_access_key':'updated'}))
# This should warn about updating the credentials, and if you check the file the keys should be updated.
```

Yash621:
@JackUrb I have done the updations that you suggested ,please review it

PR #609, opened 2021-11-17
EC2Architect - Various small improvements

JackUrb:
# Overview
In more testing, ran into two primary issues:
1. With `tmux` sessions sometimes the ssh-agent forwarding would become stale and then setting up the server would hang indefinitely. 
2. Determining the priority to use based on the number of rules meant that if some servers shut down (reducing the rule number) and then more were launched, two servers could attempt to be launched with the same rule priority.
3. The subdomain string could contain special characters not allowed by amazon

# Implementation
1. Clearing `SSH_AUTH_SOCK` in the env overwrites use of the ssh-agent for auth (which we don't need as we're referring to the `.pem` keypair anyways), which _should_ resolve `tmux` issues (can't directly reproduce yet).
2. Implements a search over possible priority numbers
3. Runs a regex replace to remove special characters from the provided subdomain
4. Also adds an `Owner` tag to all resources, to make them easier to track

cc @klshuster @ytsheng

### DIFF
--- a/mephisto/abstractions/architects/ec2/ec2_architect.py
+++ b/mephisto/abstractions/architects/ec2/ec2_architect.py
@@ -43,6 +43,15 @@ FINAL_SERVER_BUILD_DIRECTORY = "routing_server"
 DEPLOY_WAIT_TIME = 3
 
 
+def url_safe_string(in_string: str) -> str:
+    """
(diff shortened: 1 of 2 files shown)


JackUrb:
Merging without review after testing, to get as many people on the bugless version as possible.

PR #615, opened 2021-11-28
 Improved MTurk credential override UX 

Yash621:
@JackUrb I have done the updations that you suggested ,please review it !
rather than a regex search I am finding the profile among the multiple profile and then accessing the keys through indexing.
This PR is in respect to issue #531 and a continuation of PR #604


mephisto/abstractions/providers/mturk/mturk_utils.py
```diff
@@ -61,21 +61,55 @@ def check_aws_credentials(profile_name: str) -> bool:
...
+            aws_credentials = aws_credentials_file_string.split("\n")
+            # iterating to get the profile
+            profileFound = False
+            for credentialIndex in range(0, len(aws_credentials)):
+                if profileFound:
+                    aws_access_key_id = aws_credentials[credentialIndex]
+                    aws_secret_access_key = aws_credentials[credentialIndex+1]
+                    break
+                if str(aws_credentials[credentialIndex]).startswith("[{}]".format(profile_name)):
+                    profileFound = True
```
JackUrb, line 87: Nit - this body could definitely be simplified:
```
for credentialIndex in range(0, len(aws_credentials)):
    if str(aws_credentials[credentialIndex]).startswith("[{}]".format(profile_name)):
        aws_access_key_id = aws_credentials[credentialIndex+1]
        aws_secret_access_key = aws_credentials[credentialIndex+2]
    break  # credentialIndex now has the index into credentials we're using.
```

JackUrb: Note however that the contents of the lines in `aws_credentials[credentialIndex+1]` and `aws_credentials[credentialIndex+2]` are _not_ just the credentials, they are as follows:
```
aws_credentials_file.write(
    "aws_access_key_id={}\n".format(register_args.access_key_id)
)
```
Your code should anticipate this and look for the full string to compare to. (either by stripping the content before the first `=` or by comparing to strings containing this as well).

mephisto/abstractions/providers/mturk/mturk_utils.py
```diff
@@ -61,21 +61,55 @@ def check_aws_credentials(profile_name: str) -> bool:
...
+
+            if register_args.access_key_id != aws_access_key_id or register_args.secret_access_key != aws_secret_access_key:
```
JackUrb, line 89: As noted above, these aren't _quite_ right, and the way the code is currently written they will never be the same and this conditional will always be entered.

mephisto/abstractions/providers/mturk/mturk_utils.py
```diff
@@ -61,21 +61,55 @@ def check_aws_credentials(profile_name: str) -> bool:
...
+                    # overWrite login details
+                    aws_credentials_file.write("[{}]\n".format(profile_name))
+                    aws_credentials_file.write(
+                        "aws_access_key_id={}\n".format(
+                            register_args.access_key_id)
+                    )
+                    aws_credentials_file.write(
+                        "aws_secret_access_key={}\n".format(
+                            register_args.secret_access_key)
+                    )
```
JackUrb, line 108: The write process actually simplifies if you're trying to follow your current strategy, as you can insert directly:
```
aws_credentials[credentialIndex+1] = ... # update to the new value
aws_credentials[credentialIndex+2] = ... # update to the new value

with open(expanded_aws_file_path, 'w') as aws_credentials_file:
  aws_credentials_file.write("\n".join(aws_credentials))
```
This ensures that the old values are no longer in the file, and have been replaced by the new ones.

You can also skip the `# clean up file` code as the file should already be prepared if you're only making edits.

JackUrb (commented):
Getting closer! You're nearly there, after this next set I encourage you to run the test cases I supplied last time and ensure they work, as you'll likely be good to go afterwards!

Issue #614, opened 2021-11-24
`EC2Architect` fix warning when domain is in use


JackUrb:
# Overview:
One possible failure mode of the `EC2Architect` arises during setup, where if the chosen subdomain is already in use we may fail to push to the server.  This is currently broken due to the introduction of `url_safe_string`, so the checked subdomain isn't the one that's actually in use. 
After fixing this and adding a file check, we should catch both the easy instances where the server exists locally (in the `servers` folder) and the cases where the domain is already existing (with `rule_is_new`).

# Code Pointers:
Where we currently write server details to:
https://github.com/facebookresearch/Mephisto/blob/main/mephisto/abstractions/architects/ec2/ec2_architect.py#L274
Where we should check the subdomain in servers (note we should call `url_safe_string` to convert):
https://github.com/facebookresearch/Mephisto/blob/main/mephisto/abstractions/architects/ec2/ec2_architect.py#L170

Yash621:
@JackUrb  can I work on it ?

JackUrb:
Sure, it should be an easy fix. You just need to make sure the `url_safe_string` method gets called, then check the servers folder for files named `<subdomain>.json`.

Yash621:
@JackUrb  okay ,I will work on it !! secondly I have done the changes on the PR for issue #531 please review it :)

PR #617, opened 2021-12-06
EC2Architect fixed warning when domain is in use

Yash621:
This PR is in respect to issue #614 
@JackUrb please review it , I making sure that  url_safe_string method gets called, then checking the servers folder for files starting with "subdomain".json.


mephisto/abstractions/architects/ec2/ec2_architect.py
```diff
@@ -156,6 +158,17 @@ def download_file(self, target_filename: str, save_dir: str) -> None:
                 if chunk:
                     out_file.write(chunk)
 
+    def check_domain_exists(self, subdomain_name: str):
+        """
+          Checks if the domain name exists in Route53
```
JackUrb, line 163: Docstring can be more direct on what's happening within:
```
"""
Checks to see if we have an active local record for the given subdomain
"""
```

mephisto/abstractions/architects/ec2/ec2_architect.py
```diff
@@ -156,6 +158,17 @@ def download_file(self, target_filename: str, save_dir: str) -> None:
...
+        """
+        with open(self.server_detail_path, "r") as server_detail_file:
+            server_detail_file_files = server_detail_file.read().split('/n')
+            for file in server_detail_file_files:
+                if file.startswith(f'{subdomain_name}.json'):
+                    return True
+        return False
```
JackUrb, line 170: The check itself is really much simpler - we just need to see if the file at `self.server_detail_path` exists or not. If not, we're clear to create. If so, then there's potentially an issue.

mephisto/abstractions/architects/ec2/ec2_architect.py
```diff
@@ -166,8 +179,13 @@ def assert_task_args(cls, args: DictConfig, shared_state: "SharedTaskState"):
         assert ec2_helpers.check_aws_credentials(
             profile_name
         ), "Given profile doesn't have registered credentials"
+        #   Producing a domain string that is safe for use
+        #   in ec2 resources
+        subdomain = url_safe_string(args.architect.subdomain)
+
+        assert cls.check_domain_exists(
```
JackUrb, line 186: We'd actually like to be sure that the domain _doesn't_ exist yet.

JackUrb (changes requested):
Thanks for taking this one on! Some simple changes inline.

Also, more generally, I'm curious as to why you're making additional changes outside of the core files relevant to the feature, like those in `ec2_helpers.py`. These would be reverted by `black`, and are likely the cause of your branch not passing the style check at the moment.

Yash621:
@JackUrb i have done all the changes that you suggested 
please review it !

PR #636, opened 2022-01-05
Record StaticAgentState task start and end

EricMichaelSmith:
Use existing methods of `AgentState` to return the task start and end times for `StaticAgentState`


JackUrb (approved):
❤️

PR #639, opened 2022-01-09
Prevent EC2 architect target group collisions

JackUrb:
Target groups created with similar names due to long-named task series were sometimes mapping to the same group, say if `my-really-long-task-name-that-I-launched-version-1` and `my-really-long-task-name-that-I-launched-version-2` were launched at the same time.

This fixes the issue by adding a hash to the front of the target group of the full name (as we must truncate the name).

# Testing
Local proof of concept
```
>>> import hashlib
>>> group_name = "mephisto-fallback_group"
>>> group_name_hash = hashlib.md5(group_name.encode('utf-8')).hexdigest()
>>> anti_collision_group_name = f"{group_name_hash[:8]}-{group_name}"
>>> final_group_name = f"{anti_collision_group_name[:28]}-tg"
>>> final_group_name[:32]
'7fc4e428-mephisto-fallback-g-tg'
```
Ran a job on sandbox with this setup without issue as well

### DIFF
--- a/mephisto/abstractions/architects/ec2/ec2_helpers.py
+++ b/mephisto/abstractions/architects/ec2/ec2_helpers.py
@@ -12,6 +12,7 @@ import os
 import subprocess
 import json
 import getpass
+import hashlib
 from mephisto.abstractions.providers.mturk.mturk_utils import setup_aws_credentials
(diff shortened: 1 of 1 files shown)


pringshia:
nice fix

PR #640, opened 2022-01-09
EC2 resource cleanup warning

jxmsML:
1. `mephisto/abstractions/architects/ec2/cleanup_ec2_server_by_name.py` which clean local server by name and will ensure fallback server will not be cleanedup.
2. adding the following to `cleanup_all_ec2_resources.py`:
1) a warning saying "This is going to completely delete all resources for your EC2 architect, including the VPC, routes, and fallback server. Are you sure you want to do this?
2) a check to see how many rules are set up on the current server (perhaps using this: https://github.com/facebookresearch/Mephisto/blob/80baca64b53814bd14ce6b0edffef0dda6094868/mephisto/abstractions/architects/ec2/ec2_helpers.py#L677-L680) and ensuring there's only 1 rule, otherwise throwing a second warning "There are still existing rules on the router, which would imply that active jobs are running right now. Are you SURE you want to DELETE ALL?"

### DIFF
--- a/mephisto/abstractions/architects/ec2/cleanup_ec2_resources.py
+++ b/mephisto/abstractions/architects/ec2/cleanup_ec2_resources.py
@@ -14,6 +14,11 @@ from typing import Dict, Any
 
 # TODO Hydrize
 def main():
+    confirm = input(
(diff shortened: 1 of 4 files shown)


JackUrb (approved):
Thanks for adding these checks, will definitely make it much harder to remove the wrong server.

PR #643, opened 2022-01-19
[js] Use HTMLFormElement prototype func to submit forms

jacobkahn:
Don't use the property-based `submit` callable on the HTMLFormElement --
because tasktask data is attached as form `input` elements, an `input`
element with the property name `submit` will [clobber](https://github.com/facebookresearch/Mephisto/blob/bbc14aab6bef682f24912fb992c104d3f8b8c035/mephisto/abstractions/providers/mturk_sandbox/wrap_crowd_source.js#L66) the form's property.
- `caniuse` - https://caniuse.com/mdn-api_htmlformelement
- before/after behavior - https://jsfiddle.net/Lhtodmv7/31/
- also tested on a standalone repro with the static React app example


JackUrb (approved):
Great fix - thanks for committing this improvement!

Issue #476, opened 2021-06-04
Static onboarding flow fails with flask server type


Rebecca-Qian:
I was running the static HTML flow with onboarding from Mephisto's provided examples. I got this error when running with the flask server type, but not with the default node server.

The task UI loads, but form submission fails.

```
[2021-06-03 12:32:07,148][mephisto.abstractions.providers.mock.mock_unit][INFO] - Mock task launched: localhost:3000 for preview, localhost:3000/?worker_id=x&assignment_id=291 for assignment 283
[2021-06-03 12:32:27,457][mephisto.abstractions.blueprints.abstract.static_task.static_agent_state][INFO] - SAVED_DATA_TO_DISC at /private/home/rebeccaqian/Mephisto/data/data/runs/NO_PROJECT/109/onboarding/onboarding_13/agent_data.json
Unhandled exception in onboarding OnboardingAgent(13): AssertionError('Static tasks should only have final act')
Traceback (most recent call last):
  File "/private/home/rebeccaqian/Mephisto/mephisto/abstractions/blueprint.py", line 176, in launch_onboarding
    self.run_onboarding(onboarding_agent)
  File "/private/home/rebeccaqian/Mephisto/mephisto/abstractions/blueprints/abstract/static_task/static_task_runner.py", line 63, in run_onboarding
    agent_act = agent.act(timeout=self.assignment_duration_in_seconds)
  File "/private/home/rebeccaqian/Mephisto/mephisto/data_model/agent.py", line 523, in act
    self.state.update_data(act)
  File "/private/home/rebeccaqian/Mephisto/mephisto/abstractions/blueprints/abstract/static_task/static_agent_state.py", line 100, in update_data
    ), "Static tasks should only have final act"
AssertionError: Static tasks should only have final act
[2021-06-03 12:32:36,730][mephisto.operations.supervisor][INFO] - Onboarding agent onboarding_13 disconnected or errored, final status onboarding.
[2021-06-03 12:32:37,130][mephisto.operations.operator][INFO] - Operator running task ID = 109
```

Full trace available at Internal paste: P419933290

The command I used to run:
```
python static_run_with_onboarding.py mephisto.architect.server_type=flask mephisto/architect=local
```

JackUrb:
This should be resolved by #655, to be included in the 1.0 release!

Issue #419, opened 2021-03-16
[Clarifications] Testing in a local environment


federicoruggeri:
Dear All,
I've started working with ParlAI and Mephisto some days ago in order to define an interface for data collection. In particular, I've extend the parlai-chat-demo to meet my task-specific requirements.

I was wondering how to test the interface with users in a local network or by remotely connecting to my server. Thus, I've looked up for existing threads about the same topic: [#409](https://github.com/facebookresearch/Mephisto/issues/409) and [#303 ](https://github.com/facebookresearch/Mephisto/issues/303) as the main examples.

Since I'm no expert in this kind of stuff, I'm not completely sure what are the current networking setups that I can consider at the moment.  Therefore, I kindly ask you for a gentle summary/tutorial concerning the particular objective of testing locally/remotely with own server.

To summarize, I would like to know if there's an answer to these main questions:
Q: Is it possible to easily test the interface with other people in a local environment (same network)?
Q: How to do so?

Down below I report a list of additional questions with related comments:

1. My server has a static IP, can I test it locally on the same network? Connecting from the same network to the server brings up a 'disconnected' status in the interface.

2. Can I use the mentioned nginx reverse proxy to access to my interface? What are the steps? I've tried setting it up and then access to the server. I'm able to access to the interface but the connection status is still 'disconnected'.

3. Do I need a specific mephisto configuration for local testing? Up until now I've worked with `"mephisto/architect"="local"` and `"mephisto/provider"="mock"`.

Thanks in advance and apologies if this issue sounds just as a duplicate of some past issues.
Unfortunately, I'm not sure I've done everything right until now.

JackUrb:
Hi @federicoruggeri, sorry for the delay here. To answer your questions, testing locally _should_ be possible, and shouldn't require usage of a reverse proxy. If you have a static IP on the network, you can specify the IP itself as the `mephisto.architect.hostname` argument, which updates the `LocalArchitect` to be listening to more than just requests for `localhost`. This should resolve your issue, but let me know if you're still stuck after this.

JackUrb:
> I still get 'disconnected' status bar with the parlai_chat_task_demo app.

That's somewhat surprising - anything appear in your javascript console logs, or in the terminal logs, when you try to do this?

> what is the corresponding python version for setting the suggested property?

[...]

PR #655, opened 2022-02-07
Architect API v1.0 master PR

### LINKED ISSUE #651: Architect-API: Upgrade and Clarify responsibilities of Mephisto Routing Server
The original design of the routing server is starting to show some signs of growing pains. Now that we have a better idea of the types of packets sent, the responsibilities of the separate server components, and where we're heading next, it's time for an update.

JackUrb:
# Overview
The original communication protocol between Mephisto, the routing server, and JS clients is showing clear signs of age now that we're moving towards handling non-ParlAI types of live tasks (model in the loop especially), and as I try to add clearer logging throughout the system. This PR includes (in the included doc) a proposal for a new design for the system which should both clarify the responsibilities and stages of the system, as well as simplifying away some of the overhead that we incorporated into Mephisto but can generalize away from.

Closes #651, #476, #419.

# Issues
- [x] I still haven't been able to get linked use of local packages to work consistently across _all_ locally run Mephisto examples. May lead to some troubles as I'm testing things out.
- [x] ~~We're moving from 13 packet types down to 8, and doing so hits locations across 3 core areas of our infrastructure. I'm expecting it to be impossible to make some of the sub-changes non-breaking.~~ Ended up with 11 packets, and still some breaking changes that were addressed.
- [x] Major state transitions around "status complete" events can be somewhat hairy, as we have two methods of submission presently (backend forcing a task to be marked 'ready for submission', and frontend logic for 'ready for submission'). The backend-forced 'ready' state is quite overcomplicated and could see removal here.
- [x] ~~We've been asked a number of times to create a "post-task" state, where after completing the main task a worker may be able to submit feedback or a survey. This PR may not include that state, but should be mindful to make it easy enough to insert later.~~ Considered, but we didn't make this change. Live tasks can choose to include post-task info in the `final_submission` without forcing other tasks to await everyone to submit
- [x] While we're in, #419 would be easy to solve here too.

# Implementation
- [x] #656 Update the packet types, and change the `SocketIOHandler` to refer to the new calls. Update testing locations in the `MockServer` to follow the new registration protocol. (breaking). 
- [x] #657 Update the `Agent` class API to use the correct sending protocols (especially as `act` no longer sends a specific act-based packet anymore). Update `TaskRunner`s to clarify the weird 'done-but-waiting-submission' state (testable, but breaking for manual example tests).
- [x] #658 Update the node router to use the correct packet types as well, and to remove old responsibilities.
- [x] #660 Update `mephisto-task` to send things to the correct locations. Update all `status_done` related state such that there's only ever one done submission and _all_ tasks trigger it. Update the `StaticTaskRunner` to handle these as well (partially testable with the example Static Tasks)
- [x] #661 Update `bootstrap-chat` to use the new paradigm for sending, ideally in a way that doesn't change the interface users already have with this package. Update the `ParlAIChatTaskRunner` to handle these tasks now (fully testable)
- [x] (Unrelated QOL) #662 Quality of life fix for asyncio shutdowns, and clearer labelling 
- [x] (Refactor) #673
- [x] #674 
-  ~~Create the `RemoteQueryBlueprint` to prepare us for model-in-the-loop tooling~~ Will handle in a follow-up pr creating a `mephisto-remote-query` package.
- [x] #675 - noting version mismatches for `mephisto-task`
- [x] #676 - ensure that npm versions match what's on `main` 

# Review requirements:
- [x] For now, just checking up on the proposed new Architect plan. I'll start moving forward with some of this, but getting eyes on it before I commit too hard to any part of the design would be great!


docs/web/docs/explanations/architect_api.md
```diff
@@ -0,0 +1,108 @@
...
+4. The interaction layer should be simple. 
+
+This document aims to describe how we got from these requirements and goals to the system we have now.
+
+## Requirements
+
+### `Worker`/`Agent` registration
+Mephisto allows for multiple stages of filtering for a worker, which leads the registration process to be staged. The complete order is listed below:
+0. `CrowdProvider`-set qualifications may prevent a worker from even seeing a task.
+1. `Qualification`s may be set that a `CrowdProvider` is not aware of. Workers not meeting these qualifications should be filtered.
```
pringshia, line 22: This is the only one that isn't clear to me. What is meant by "should be filtered" - by whom? Does Mephisto take care of this, or is it up to the task author? What is a use case where you'd do this over the option above?

docs/web/docs/explanations/architect_api.md
```diff
@@ -0,0 +1,108 @@
...
+Over the course of a task, there are a few key states to consider:
+1. Connecting
+2. Onboarding
+3. Waiting (for other `Unit`s in an assignment)
+4. In-Task
+5. Completed (task complete, or partner disconnects)
+6. Disconnected (Server disconnect, timeout, return)
+7. Failed to connect (no available tasks)
+
+We've also considered a "post-task" state after the completion of a task for surveys or related content.
```
pringshia, line 54: Wonder if another way of reframing this is simply as a "live" task, e.g. #2 under 'Task data submission' above.

That is to say, we just send the final data back as "arbitrary task data", and then the post-task data could also be sent as either further "arbitrary task data" or completion data.

docs/web/docs/explanations/architect_api.md
```diff
@@ -0,0 +1,108 @@
...
+The core carrier for information in the Mephisto Architect API is the `Packet` class. Downstream, they act as the way to tie an `Agent` class in python to an actual human worker. 
+
+The `Channel` is the primary way of trasmitting packets, with the `WebsocketChannel` being the main implementation Mephisto currently uses with its `Architect`s. The `ClientIOHandler` is responsible for using and interpreting packets, so it defines the key types to be handled.
+
+### Packet Types
+There are a number of different packet types used by Mephisto:
+- `PACKET_TYPE_ALIVE`: Used to mark the success of a new socket connection.
+- `PACKET_TYPE_SUBMIT`: Used to handle submission of any subcomponent of a `Unit`, currently either onboarding or full task. 
+- `PACKET_TYPE_OUTGOING_LIVE_DATA`: Used to send any new live data to an agent.
+- `PACKET_TYPE_INCOMING_LIVE_DATA`: Used for the frontend to send any type of data to the backend, usually to be processed by a user-defined callback.
```
pringshia, line 78: [Request for consideration]

Since Outgoing/Incoming are relative terms, they require a subject and object to reduce ambiguity. Outgoing from what? Coming into what?

One proposition is **Client-bound** (or **Agent-bound**) & **Server-bound** instead. Yielding: `PACKET_TYPE_AGENT_BOUND_LIVE_DATA` and `PACKET_TYPE_SERVER_BOUND_LIVE_DATA`.

Grokking this immediately denotes the directional relationship without ambiguity.

docs/web/docs/explanations/architect_api.md
```diff
@@ -0,0 +1,108 @@
...
+The `Channel` is the primary way of trasmitting packets, with the `WebsocketChannel` being the main implementation Mephisto currently uses with its `Architect`s. The `ClientIOHandler` is responsible for using and interpreting packets, so it defines the key types to be handled.
+
+### Packet Types
+There are a number of different packet types used by Mephisto:
+- `PACKET_TYPE_ALIVE`: Used to mark the success of a new socket connection.
+- `PACKET_TYPE_SUBMIT`: Used to handle submission of any subcomponent of a `Unit`, currently either onboarding or full task. 
+- `PACKET_TYPE_OUTGOING_LIVE_DATA`: Used to send any new live data to an agent.
+- `PACKET_TYPE_INCOMING_LIVE_DATA`: Used for the frontend to send any type of data to the backend, usually to be processed by a user-defined callback. 
+- `PACKET_TYPE_REGISTER_WORKER`: Used to request a new `Agent` and `Unit` data for a specific worker on a task.
+- `PACKET_TYPE_AGENT_DETAILS`: Used to respond with the details of a worker registration request.
```
pringshia, line 80: [Request for clarification]

I find it interesting that the description for PACKET_TYPE_REGISTER_**WORKER** uses the term Agent.
And the description for PACKET_TYPE_**AGENT**_DETAILS uses the term Worker.

Possible to just coalesce the verbiage here, and just call it Agent. Proposition: changing the enum to  `PACKET_TYPE_REGISTER_AGENT`

docs/web/docs/explanations/architect_api.md
```diff
@@ -0,0 +1,108 @@
...
+- `PACKET_TYPE_REQUEST_STATUSES`: Used by Mephisto to poll for the current statuses for a worker.
+- `PACKET_TYPE_RETURN_STATUES`: Used by the router to return updates for all of the currently registered agents.
```
pringshia, line 82: [Nit]

Says statues instead of Statuses

docs/web/docs/explanations/architect_api.md
```diff
@@ -0,0 +1,108 @@
...
+- The `ClientIOHandler` must send a `PACKET_TYPE_ALIVE` whenever it opens a new channel (in this case, to a router).
+- For `Unit` Registration, in response to a `PACKET_TYPE_REGISTER_WORKER`, the handler must return a `PACKET_TYPE_AGENT_DETAILS` with the details of an agent and it's initialization data, or the failure status for why an agent couldn't be created.
+- During a `Unit` the handler must process `PACKET_TYPE_INCOMING_LIVE_DATA` and direct the content to the correct handlers, and should send a `PACKET_TYPE_OUTGOING_LIVE_DATA` for any `Agent.send_data()` call on a live connected `Agent`. The handler must also process `PACKET_TYPE_SUBMIT` packets for the key transitions of a `Unit` in progress, and should respond with `PACKET_TYPE_AGENT_DETAILS` for a submit on an `OnboardingAgent`.
+- Over any run, the handler should poll with `PACKET_TYPE_REQUEST_STATUS` and update local `Agent` statuses on disconnects from `PACKET_TYPE_RETURN_STATUS`. This also acts as a heartbeat from the Python core to the router. The handler should also take `PACKET_TYPE_ERROR` and log the contents if this ever occurs.
```
pringshia, line 91: [Nit]

It says _STATUS (singular) here whereas above it says _STATUSES (plural)

docs/web/docs/explanations/architect_api.md
```diff
@@ -0,0 +1,108 @@
...
+
+
+### Router Responsibilities
+
+The primary responsiblity of the router is to take incoming packets from client connections and direct them to the core Mephisto `ClientIOHandler` and to do the reverse as well. All packets will have a core `agent_id` field denoting either the sender or receiver of the packet, depending on the packet type. The only exception is the `PACKET_TYPE_ALIVE`, which is directed to the router and allows for any registration of an incoming connection.
+
+Secondarily, the router is responsible for converting RESTful `POST` requests from `mephisto-task` into socket messages, and relaying the response as a standard `POST` response. This behavior is _only_ for the `PACKET_TYPE_REGISTER_WORKER`, and `PACKET_TYPE_SUBMIT` packets, and both of them will be serviced by `PACKET_TYPE_AGENT_DETAILS` responses. For these it should be listening to `POST` requests at `/register_worker`, `submit_onboarding`, and `/submit_task`. `POST` requests to `/log_error` should result in forwarding a `PACKET_TYPE_ERROR`.
```
pringshia, line 98: [Nit]

`submit_onboarding` -> `/submit_onboarding`

docs/web/docs/explanations/architect_api.md
```diff
@@ -0,0 +1,108 @@
...
+
+Fourth, the router is responsible for serving the static `task_config.json` file, which allows the frontend to load certain details about the full task before going through any registration handshakes.
+
+### `mephisto-task` responsibilities.
+
+The `useMephistoTask` hook is responsible for allowing a worker to connect to a task and submit the relevant data. For this, it only needs to make `POST` requests related to the `PACKET_TYPE_SUBMIT` and `PACKET_TYPE_REGISTER_WORKER` events. The former should be triggered on `handleSubmit`, while the latter should trigger immediately on load.
+
+The `useMephistoLiveTask` hook is responsible for the rest of the packets. Data packets should be sent via `sendData` and handled with the `onLiveDataReceived` callback. Providing a callback to `sendData` will call that with any response data that sets the `source_request_id` to the initial send's `message_id`. Local connection state is managed by the `socket_handler.jsx` class, and surfaced through the hook variables for UIs to use.
```
pringshia, line 108: [Request for clarification]

> Providing a callback to `sendData` will call that with any response data that sets the `source_request_id` to the initial send's `message_id`.

I didn't understand this fully. Why would I want to do this? The reader is introduced to `source_request_id` and `message_id` for the first time here, so some more context could be helpful.

pringshia:
Thanks for this! I found the overall documentation clear to read and understand. Pretty happy with the architecture design. Some comments scattered throughout.

docs/web/docs/explanations/architect_api.md
JackUrb: Ah Mephisto takes care of filtering out workers by any qualifications that the task owner sets, regardless of if those qualifications are propagated to a `CrowdProvider`. I imagine this is a fairly uncommon use case, but if someone wants to keep _private qualifications_ of some kind, this would be the way to do it.

docs/web/docs/explanations/architect_api.md
JackUrb: The difficulty here is, under the current setup, "live tasks" with multiple workers consider disconnect events of partners to be the end of a task. As such a live task with multiple workers can't be fully submitted until everyone is done, given the current architecture. 

I'd love to take on an alternative, but I don't quite know what one would be.

docs/web/docs/explanations/architect_api.md
JackUrb: Sounds good, will address in the first PR creating these new packet types.

docs/web/docs/explanations/architect_api.md
JackUrb: Fine by me, and it's a good point. I was somewhat inheriting old verbiage, wherein first a worker is created, then an agent, but ultimately it is agent registration. Will update!

JackUrb: Under the new setup, both `Worker` and `Agent` creation happen on the same request, but the latter _always_ happens while the former is more a 'find-or-create' operation.

docs/web/docs/explanations/architect_api.md
JackUrb: Noted and fixed, thanks!

docs/web/docs/explanations/architect_api.md
JackUrb: Adding some clarifying text:
`Qualification`s may be set locally that a `CrowdProvider` is not aware of. Workers not meeting these Mephisto qualifications (as set by `SharedState.qualifications`) should be filtered out.

docs/web/docs/explanations/architect_api.md
JackUrb: Yeah this is somewhat vague, but I'm not sure how in-depth I wanted to get here. 

The main thought is that, upon sending a request to the backend, we (`mephisto-task`) attach a `message_id` to the outgoing message. Then the callback handler (yet-to-be-built) should send any resultant messages with a `source_request_id` set (or perhaps `context_id` may be more appropriate), such that `mephisto-task` can later call that specific callback. If no such callback is provided, then only the default `onLiveDataReceived` callback is used. Happy to flesh this out more though.

docs/web/docs/explanations/architect_api.md
pringshia: Ah - interesting caveat!

docs/web/docs/explanations/architect_api.md
pringshia: Curious to learn more about a concrete use case where this would be used. I think I may understand what you're getting at but there are still a few gaps in my understanding. The general idea seems to be to link outgoing requests to incoming responses and maintaining a sort of "correlation id" right?

docs/web/docs/explanations/architect_api.md
pringshia: Ah I see, this bit was the most helpful for me: "I imagine this is a fairly uncommon use case, but if someone wants to keep private qualifications of some kind, this would be the way to do it"

pringshia (approved):
Approved! (Already approved the sub-PRs merging into this)

PR #660, opened 2022-02-12
Updating `mephisto-task` to use the Architect 1.0 API, pushing to static tasks.

### LINKED ISSUE #419: [Clarifications] Testing in a local environment

JackUrb:
# Overview
As part of #655, this PR contains the full set of changes to update `mephisto-task` to work with the 1.0 API for architects and packets. Upgrades `mephisto-task` to version `2.0.0` to note this. Also happens to fix the bug in #419 while we're at it. Incorporates the changes from #659 to allow me to do the editing.

# Implementation
There are a lot of changes here, but the most critical surround the change to the API for `mephisto-task`'s `AgentState`. We used to directly keep track of a number of variables that would persist in our local storage, when these were frequently task-specific. By allowing these to be parsed from 'live data' packets instead, we can allow the downstream tasks to decide what to keep and how to parse, rather than leaking that through the abstraction.

One element that remains a bit leaky - triggering these kinds of state callbacks requires that the live data packet you send contains a `{ 'state': {...}}` field. I've noted this in the architecture doc. Ultimately I'm fine with having the packets serve two functions starting at the `mephisto-task` level, as I think it lets me abstract the Packets from the data that is sent with them. It _may be worthwhile_ to add some kind of function to the `Agent` class like `send_state` or something that just acts as a wrapper around this send, but that could be unnecessary. 

With this change, the callbacks that trigger for `useMephistoTask` and `useMephistoLiveTask` needed to be updated somewhat, but hopefully in a way that's clear with this context.

For a full list of changes:
- `useMephistoTask`
  - Sets `blockedExplanation` as part of the state, this way we can pull from an `AgentDetails` event when the Mephisto backend provides a reason for us to show the worker.
  - Updates both the regular registration handshake as well as the post-onboarding handshake to use the new one-step semantics
  - Deletes unused code from `utils.js`
- `useMephistoLiveTask`:
  - Updates the `socket_handler.jsx` to use the new packet types
  - Removes `onStateUpdate` as we no longer track this at the packet level
  - Fixes #419 by inheriting the location's protocol instead of the localhost hack
  - Creates an effect for receiving initial data that checks `raw_messages` to allow a `mephisto-task` task to replay them, restoring the original state and messages present.
  - Implements the `state` switch noted above for `onMessageReceived`, which is now pulled out for use with `raw_messages`.
- Removes `getWorkerRegistrationInfo` from the required functions of `wrap_crowd_source.js` as we only have a single step registration handshake now. Updates the documentation to reflect this.
- Fixes how `StaticAgentState` handles parsing submissions with files. Now that we directly pass the submitted data rather than the full context packet, we don't need to extract out from onboarding or task data. We can then check for files and do final file parsing.
- Minor bugfixes on the `StaticHTMLBlueprint` tasks. Transitioning from onboarding now works correctly in a way that required a refresh to fix before.
- Updates packages throughout to be using the newer versions of `webpack` and `webpack-cli`
- Small fixes to hydra config `.yaml`s that were actually broken :(

# Testing
Automated tests should still be passing, considering this isn't touched.
I updated both of the `StaticBlueprint` task examples, and tried out various configuration steps. Went fully from onboarding through to completing a task in both setups.
**React preview**
Screen Shot 2022-02-10 at 4 52 06 PM
**React task**
Screen Shot 2022-02-11 at 9 35 01 AM
**React onboarding**
Screen Shot 2022-02-11 at 9 54 40 AM
**HTML task (with file upload)**
Screen Shot 2022-02-11 at 10 45 17 AM
**HTML onboarding**
Screen Shot 2022-02-11 at 10 44 20 AM


packages/mephisto-task/src/utils.js
```diff
@@ -93,23 +93,8 @@ export function postProviderRequest(endpoint, data) {
...
-}
-
-// Sends a request to get the initial task data
-export function getInitTaskData(mephisto_worker_id, agent_id) {
-  return postProviderRequest("/initial_task_data", {
-    mephisto_worker_id: mephisto_worker_id,
-    agent_id: agent_id,
-  });
+export function requestAgent() {
+  return postProviderRequest("/request_agent", getAgentRegistration());
```
pringshia, line 97: Nice. I appreciate this single handshake, and the simplification over the multiple handshakes that needed to occur before.

packages/mephisto-task/src/live.js
```diff
@@ -24,27 +24,67 @@ const AGENT_STATUS = {
...
+
+  const { initialTaskData } = taskProps;
+
+  function onMessageReceived(message) {
+    const { state, message_id, timestamp, ...remaining_message } = message;
+    if (state !== undefined) {
+      props.onStateUpdate && props.onStateUpdate({ state });
+    }
+    if (Object.keys(remaining_message).length != 0) {
+      props.onMessageReceived && props.onMessageReceived(remaining_message);
```
pringshia, line 67: - I suggest we rename the local `onMessageReceived` function here to `handleMessageReceived`. It's serving more as [a handler than a callback](https://stackoverflow.com/a/2069800) in that its functionality is not passed around the code. The `on*` prefix initially confused me as it led me to think otherwise.
  - This also more clearly differentiates it from `props.onMessageReceived` - which is indeed a callback given that it's passed via the `props` argument.

- Is there any reason we don't want to pass down `message_id` and `timestamp` down to the callback? Could there be a use case where the end user requires them?
- Given the gated clause of `Object.keys(remaining_message).length != 0`, we'll never invoke onMessageReceived when there's only a state update. Just verifying that this is intended behavior?

- I want to challenge the object model here a bit. Correct me if I'm wrong: We have `raw_messages` (e.g. `initialTaskData.raw_messages`), which produce either `state`s or `message`s or both. A `message` is defined simply as any `raw_message` that has a field other than `state, message_id, timestamp`. A state update is simply any message that includes a `state` field. Correct?
  - What's also a bit confusing here is that [in the backend code](https://github.com/facebookresearch/Mephisto/blob/6e2a02d94ee29dac7d5ba501bd3c5b17eea4a973/mephisto/abstractions/blueprints/parlai_chat/parlai_chat_agent_state.py#L61), the `raw_messages` being sent over are called retrieved from a variable called `messages`. I think we could maybe use this refactoring as an opportunity to be more consistent, precise and defined about the ontology of messages.
  - Note: I think `state` here should be `status` (see earlier comment in this file), though I'm using `state` in this comment to keep consistent with the code in reference, even if a typo
  - Also are all `packets` a `raw_message` or is there some distinction between those two? My guess is no, because packets also include things such as heartbeats that secure the connection.

packages/mephisto-task/src/live.js
```diff
@@ -24,27 +24,67 @@ const AGENT_STATUS = {
...
+  }, [initialTaskData]);
+
   const socketProps = useConnectionHook({
     onConnectionStatusChange: (connectionStatus) => {
       setConnectionStatus(connectionStatus);
     },
-    onStateUpdate: ({ state, status }) => {
-      setAgentState(state);
+    onStatusUpdate: ({ status }) => {
       setAgentStatus(status);
```
pringshia, line 84: Really happy about this simplification and consolidation. States and status always felt a little confusing to me.

packages/mephisto-task/src/live.js
```diff
@@ -24,27 +24,67 @@ const AGENT_STATUS = {
...
   const useConnectionHook = props.customConnectionHook || defaultConnectionHook;
 
   const taskProps = useMephistoTask();
+
+  const { initialTaskData } = taskProps;
+
+  function onMessageReceived(message) {
+    const { state, message_id, timestamp, ...remaining_message } = message;
+    if (state !== undefined) {
+      props.onStateUpdate && props.onStateUpdate({ state });
```
pringshia, line 64: I thought we got rid of state and are using statuses? Is this a typo or am I missing something?

packages/mephisto-task/README.md
```diff
@@ -166,16 +166,18 @@ function MyApp() {
         destroy,
         sendMessage
 
-        agentState,
         agentStatus,
 
         connectionStatus,
     } = useMephistoLiveTask(
         onConnectionStatusChange: (connectionStatus) => {
```
pringshia, line 173: Just noticed a typo here, these props should be passed wrapped in an object, this is invalid javascript.

Fix:

```js
    } = useMephistoLiveTask({   // <-- notice the { here to start the object definition
        onConnectionStatusChange: (connectionStatus) => {
```

mephisto/abstractions/blueprints/static_html_task/source/dev/app.jsx
```diff
@@ -99,9 +96,14 @@ function MainApp() {
   );
 }
 
-function SubmitFrame({ children, onSubmit }) {
+function SubmitFrame({ children, onSubmit, currTask }) {
```
pringshia, line 99: small nit: we tend to use fully formed names in the js code by convention, so I would recommend renaming `currTask` to `currentTask`

pringshia (commented):
Wow, I'm loving these changes and the reduction in complexity incurred by them. Overall I think it will make Mephisto easier to maintain, more versatile, and friendlier to newcomers. Appreciate all of the fixes also rolled in here.

Left a few comments throughout.

packages/mephisto-task/src/live.js
JackUrb: So this was noted a bit in the PR description:
> One element that remains a bit leaky - triggering these kinds of state callbacks requires that the live data packet you send contains a { 'state': {...}} field...

I imagine that, while we're dropping `state` as an official field that we track in the package, I expect that downstream tasks may still want to use `state` for their local task context. That type of `state` under the hood is really just a regular message, rather than being something akin to status which is Mephisto-specific. Here I'm exposing a way for Mephisto to handle splitting these out such that, to the protocol, there are only `liveData` messages, but to a `mephisto-task` user they can register separate `state` and `message` callbacks. 

If this is undesirable, I can push this further along and just have `bootstrap-chat`'s message function do this split instead.

packages/mephisto-task/src/live.js
JackUrb: > rename the local onMessageReceived function

Surely!

> Is there any reason we don't want to pass down message_id and timestamp down to the callback?

These are added by Mephisto, so I stripped them back out. I can add them back in though, as I imagine they're useful to some.

> we'll never invoke onMessageReceived when there's only a state update

Correct - this is if we choose to split messages and state updates on the `mephisto-task` level. Can discuss more on this though.

> object model for `raw_messages`

Yeah this is a fair discussion and worth refactoring. `raw_messages` in this case cover all of what we consider to be "live data" which includes messages that are 'state' updates (Both `LIVE_DATA` packet types). All other packets aren't included in this. I think it makes sense for me to rename all occurrences across everywhere to refer to these received 'messages' as receiving 'live_data', and then 'raw_messages' can become 'past_live_data' instead.

packages/mephisto-task/src/live.js
pringshia: I'm still not sold on us first-classing state as part of the library, and think it may be better for users to implement that and choose to use `liveData` in that way if they wish.

packages/mephisto-task/src/live.js
```diff
@@ -58,20 +58,17 @@ const useMephistoLiveTask = function (props) {
 
   const { initialTaskData } = taskProps;
 
-  function onMessageReceived(message) {
-    const { state, message_id, timestamp, ...remaining_message } = message;
-    if (state !== undefined) {
-      props.onStateUpdate && props.onStateUpdate({ state });
-    }
+  function handleLiveData(liveData) {
+    const { message_id, timestamp, ...remaining_message } = message;
```
pringshia, line 62: I think the variable here should be `liveData` now that you've renamed

packages/mephisto-task/src/live.js
```diff
@@ -58,20 +58,17 @@ const useMephistoLiveTask = function (props) {
...
     if (Object.keys(remaining_message).length != 0) {
```
pringshia, line 63: i think we can remove this clause now too right, now that `state` is gone? unless there's a case where people would purposefully send blank liveData payloads?

packages/mephisto-task/src/live.js
```diff
@@ -58,20 +58,17 @@ const useMephistoLiveTask = function (props) {
...
+      props.onLiveDataReceived && props.onLiveDataReceived(liveData);
     }
   }
 
   React.useEffect(() => {
-    if (initialTaskData?.raw_messages) {
-      for (const raw_message of initialTaskData.raw_messages) {
-        onMessageReceived(raw_message);
+    if (initialTaskData?.past_live_data) {
+      for (const raw_message of initialTaskData.past_live_data) {
```
pringshia, line 70: do we still want to use the `raw_message` verbiage here?

PR #661, opened 2022-02-12
Updating `bootstrap-chat` to be inline with the Architect 1.0 API, test with ParlAI

JackUrb:
# Overview
Another key step of #655, this PR contains the changes to `bootstrap-chat` that bring it in line with the `mephisto-task@2.0.0` version changes. It also updates `ParlAIChatBlueprint` components to be able to test this live, and fixes a few bugs along the way.

# Implementation
Most of the changes surround how `mephisto-task` has dropped the `agentState` variable, and starts to pull this information from elsewhere (locally constructing this state). Now with `onStateUpdate` we do any expected changes to the state, and push the full thing to `taskContext`, in the same way `task_data` used to be pushed. This means that we have a 'legacy' way of pushing state-related data, but I'm not too concerned about the implications here. 

Full changes:
- `bootstrap-chat`:
  - `onStatusUpdate` and `onStateUpdate` are now split, with clearer semantics. Includes reading out the state-to-text updates for terminal statuses.
  - Bugfix: checks into `currentAgentNames` fix mislabelled names on reconnect, and help keep track of `live_data_requested` state following the `raw_messages` implementation in `mephisto-task`
  - Bugfix: `taskContext` reducer now actually creates a new object, triggering rerenders that were missed before.
  - Pull task done context from `taskContext` rather than `agentState` in `BaseFrontend.jsx`
- `ParlAIChatBlueprint`:
  - Updates the `TaskRunner` to properly deal with the way that completion is marked: requiring a submit. For onboarding this is managed by setting the agent status to WAITING (triggering an implicit submit), and for the full task this is handled by telling the frontend to show the done button to submit with.
- Updates to more modern `webpack` and `webpack-cli`, and related fixes.
- Fixes to ParlAI chat hydra `.yaml` configs.

# Testing
Automated tests are still passing, but don't provide coverage on the demos.
I ran all of the ParlAI chat demos with various configuration and disconnect scenarios, and ensured everything was working.
**Connecting to task**
Screen Shot 2022-02-11 at 12 37 09 PM
**Completing standard task**
Screen Shot 2022-02-11 at 3 49 14 PM
**Onboarding transition to full world**
Screen Shot 2022-02-11 at 3 55 56 PM
**In task after onboarding**
Screen Shot 2022-02-11 at 3 56 59 PM
**Partner disconnect**
Screen Shot 2022-02-11 at 4 14 51 PM
**Partner reconnect after disconnect**
Screen Shot 2022-02-11 at 4 14 38 PM
**Prebuilt loads**
Screen Shot 2022-02-11 at 4 20 16 PM
**Custom simple loads**
Screen Shot 2022-02-11 at 4 44 46 PM


packages/bootstrap-chat/src/composed/ChatApp.jsx
```diff
@@ -75,39 +78,55 @@ function ChatApp({
...
 
   function trackAgentName(agentName) {
-    if (agentName) {
-      const previouslyTrackedNames = taskContext.currentAgentNames || {};
-      const newAgentName = { [agentId]: agentName };
-      const currentAgentNames = { ...previouslyTrackedNames, ...newAgentName };
-      updateContext({ currentAgentNames: currentAgentNames });
-    }
+    const previouslyTrackedNames = taskContext.currentAgentNames || {};
+    const newAgentName = { [agentId]: agentName, agentName: agentName };
```
pringshia, line 82: what is the purpose of `agentName` here? currently it will be overwritten by each subsequent agent that is tracked, since they are spread onto `currentAgentNames`. something seems off here

packages/bootstrap-chat/src/composed/ChatApp.jsx
```diff
@@ -75,39 +78,55 @@ function ChatApp({
...
           AGENT_STATUS.TIMEOUT,
+          AGENT_STATUS.PARTNER_DISCONNECT,
           AGENT_STATUS.MEPHISTO_DISCONNECT,
         ].includes(status)
       ) {
         setInputMode(INPUT_MODE.INACTIVE);
-      } else if (state.wants_act) {
+        updateContext({
+          done_text: STATUS_TO_TEXT_MAP[status],
+          task_done: status == AGENT_STATUS.PARTNER_DISCONNECT,
```
pringshia, line 102: perhaps we can use this opportunity to clean up some of the names here: e.g. `isTaskDone` to denote boolean and `doneText` to keep consistent with the js camel casing? however if this variable is coming from Python, then it's probably fine to keep as is. I'm unsure if these new updates remove these from the backend entirely.

packages/bootstrap-chat/src/composed/ChatApp.jsx
```diff
@@ -75,39 +78,55 @@ function ChatApp({
...
+    const newAgentName = { [agentId]: agentName, agentName: agentName };
+    const currentAgentNames = { ...previouslyTrackedNames, ...newAgentName };
+    updateContext({ currentAgentNames: currentAgentNames });
   }
 
   let mephistoProps = useMephistoLiveTask({
-    onStateUpdate: ({ state, status }) => {
-      trackAgentName(state.agent_display_name);
-      if (state.task_done) {
-        setInputMode(INPUT_MODE.DONE);
```
pringshia, line 90: why don't we need this done case anymore?

packages/bootstrap-chat/src/composed/ChatApp.jsx
pringshia: Also this doesn't make much sense to me `task_done: status == AGENT_STATUS.PARTNER_DISCONNECT`

why is a task done when the partner disconnects? done in my head meant finished to completion, perhaps here we're signifying that the task failed and is now "done" in that sense? also what about the other error status cases other than partner disconnect?

mephisto/abstractions/blueprints/parlai_chat/parlai_chat_task_runner.py
```diff
@@ -263,6 +270,9 @@ def run_unit(self, unit: "Unit", agent: "Agent") -> None:
         while not world.episode_done() and unit.db_id in self.running_units:
             world.parley()
 
+        # Ensure agent can submit after completion
+        agent.observe({"state": {"task_done": True}})
```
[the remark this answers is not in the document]
pringshia: this seems more concise: `agent.observe({"task_done": True})`

or perhaps even this: `agent.update_state({"task_done": True})`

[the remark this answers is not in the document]
pringshia: Maybe this is related to my comments [here](https://github.com/facebookresearch/Mephisto/pull/660/files#r806126451) and [here](https://github.com/facebookresearch/Mephisto/pull/660/files#r806108770) in #660.

Instead of first-classing a `state` key, at the library level as convention, what if we didn't? I'm happy keeping the `state` key for bootstrap-chat, but maybe we should just consider state updates as a special class of `messages`?

packages/bootstrap-chat/src/composed/ChatApp.jsx
JackUrb: This ends up covering both current and previous names that we should consider to be "isSelf", but it also has a mapping for us to represent incoming messages with their correct ids. My intent was to have, for instance: 
```
{
   onboarding_1: "OnboardingAgent 1",
   "OnboardingAgent 1": "OnboardingAgent 1",
   task_agent_1: "In-task Agent",
   "In-task Agent": "In-task Agent",
}
```
be constructed by an agent first receiving `onboarding_1: "OnboardingAgent 1",` and later getting `task_agent_1: "In-task Agent",`. Then when we get incoming messages either with sender id `onboarding_1` or `OnboardingAgent 1` we know it was actually from ourself. The cause for having both allows us to get this detail on reconnect, as `ParlaIChatBlueprint` doesn't retain the mephisto-based sender ID in the raw messages later sent on reconnect.

mephisto/abstractions/blueprints/parlai_chat/parlai_chat_task_runner.py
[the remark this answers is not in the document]
JackUrb: >  instead of first-classing a state key, at the mephisto-task library level as convention, what if we didn't? I'm happy keeping the state key for bootstrap-chat, but what if we just considered state updates as a special class of messages?

I've been trying to figure this one out as well. The one thing that seems to *require* state is what I considered a common pattern of [requesting an act](https://github.com/facebookresearch/Mephisto/blob/ddc5095a55f51e1664cd325dd1e89bf514fb3c34/mephisto/operations/client_io_handler.py#L326-L331) when [acting and none is available](https://github.com/facebookresearch/Mephisto/blob/ddc5095a55f51e1664cd325dd1e89bf514fb3c34/mephisto/data_model/agent.py#L308-L314). If we remove state entirely, I'm not sure what we change for this - do we require `Blueprint`s to identify and recreate this pattern on a case-by-case basis?

packages/bootstrap-chat/src/composed/ChatApp.jsx
pringshia: Perhaps you meant to have brackets around `[agentName]` as it is now, your object would look like this:

```
{
   onboarding_1: "OnboardingAgent 1",
   task_agent_1: "In-task Agent",
   agentName: "In-task Agent",
}

```

packages/bootstrap-chat/src/composed/ChatApp.jsx
JackUrb: In this case, it's because all of the other disconnect statuses are "incomplete" in that the agent can't submit, but `PARTNER_DISCONNECT` allows submission b/c it wasn't this worker's fault.

I can clean up the casing for these though!

mephisto/abstractions/blueprints/parlai_chat/parlai_chat_task_runner.py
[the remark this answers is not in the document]
JackUrb: So in the end I'm using `task_data` as most other messages sent by parlai agents are `{'text': ..., 'task_data': ...}`, so it makes sense to maintain this design pattern rather than introducing one (considering state is task specific now!)

packages/bootstrap-chat/src/composed/ChatApp.jsx
JackUrb: Indeed - good catch!

packages/bootstrap-chat/src/composed/ChatApp.jsx
JackUrb: We still have it, just in `remaining_state.task_done` below. will be editing shortly though

mephisto/abstractions/blueprints/parlai_chat/parlai_chat_task_runner.py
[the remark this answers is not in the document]
pringshia: gotcha, makes sense!

PR #673, opened 2022-02-15
Renaming `LiveData` to `LiveUpdate` across the codebase

JackUrb:
As titled and following a discussion with @pringshia, we're renaming away from `liveData` to cover these as "update"s to the initial data packet sent for a static task.

Part of the #655 stack

# Testing

Automated and follow-up deploy tests.


pringshia:
Really happy with where we landed on this in terms of cleaning up and clarifying our terminology. Thanks for the updates

pringshia (approved):

examples/parlai_chat_task_demo/custom_input_and_messages/main.js
```diff
@@ -93,7 +93,7 @@ function RenderChatMessage({
         }
         message={message.text}
         taskData={message.task_data}
-        messageId={message.message_id}
+        messageId={message.update_id}
```
pringshia, line 96: Not sure if we want to change the `messageId` here, it's probably fine as it is maybe as that's more of `bootstrap-chat`'s data model?

JackUrb: Exactly my thought - `messageId` is more what you expect for what `bootstrap-chat` considers the updates to be.

Issue #525, opened 2021-08-23
[WIP] [Proposal] Enable mephisto-task to be aware of the version of Mephisto server running for compatibility messaging


pringshia:
wip.

JackUrb:
Closed in #655

PR #675, opened 2022-02-16
Warn when `mephisto-task` is out of sync.

JackUrb:
# Overview
In the past, we've had issues with the frontend and backend packages falling out of sync. Issue #525 touched on this, and this PR takes a step to resolve the issue by warning both on the frontend and backend when `mephisto-task` is out of sync with the version that the `mephisto` python package expects.

# Implementation
First, we include a new `CURR_MEPHISTO_TASK_VERSION` constant in a few places in the codebase that will track the current package version (more on that later). This allows us to include the expected version into the `task_config.json` file in `build_router.py`, as well as have values for the current task version in both of our `router` implementations. 

We update the `GET` request for `task_config.json` to include the current version of the `mephisto-task` package running in the frontend (thanks @pringshia for the pointer to babel `preval` and the code snippet for this). This then lets the routers check that value against the local version, and log an error to Mephisto if there's a mismatch.

We can also check the value inside of `task_config.json` on the frontend, and log there as well if there's a mismatch.

As it would be annoying to manually keep these up-to-date, I've included the `scripts/sync_mephisto_task.py` script which pulls the current version of the `mephisto-task` package and pushes that to the three files that contain `CURR_MEPHISTO_TASK_VERSION`.

# Testing
When pretending to run on version 3.0.0 in the backend, I see these errors:
Screen Shot 2022-02-16 at 12 56 52 PM
Screen Shot 2022-02-16 at 12 56 21 PM

After running `scripts/sync_mephisto_task.py` and the files update to 2.0.0, these resolve.


mephisto/abstractions/architects/router/flask/mephisto_flask_blueprint.py
```diff
@@ -35,6 +35,8 @@
 
 # Constants
 
+CURR_MEPHISTO_TASK_VERSION = "2.0.0"
```
pringshia, line 38: So it seems that this value is hardcoded in the file and then replaced when [scripts/sync_mephisto_task.py](https://github.com/facebookresearch/Mephisto/pull/675/files#diff-33575a8925bdd74eb5beca27cea0b35f16a38b02f857e253dd575225b13f47ff)

My initial expectation was that we would have the version lookup happen at run-time, e.g. when running `python run_task.py`? Any downsides to that? That approach would make sure that the number is "fresh" at all times.

If we do stick with this approach: how can we ensure that this value doesn't go stale, e.g. someone forgets to run the script when updating code? should we add a Github Action or pre-commit step to run  `sync_mephisto_task.py`?

mephisto/abstractions/architects/router/node/server.js
```diff
@@ -500,6 +501,28 @@ app.post("/log_error", function (req, res) {
...
+      data: {
+        error_type: "version-mismatch",
+        text:
+          "Package `mephisto-task` version mismatch. Expected " +
+          "version " +
+          CURR_MEPHISTO_TASK_VERSION +
+          " but frontend is currently using " +
+          mephisto_task_version +
+          ". This may cause unexpected errors, be sure to " +
+          "update your `mephisto-task` JS package.",
```
pringshia, line 519: Perhaps we can give explicit guidance around next steps in the error message:

"...be sure to update your `mephisto-task` dependency with `npm install mephisto-task@latest --save`. If this warning still persists please open an issue at `https://github.com/facebookresearch/Mephisto/issues`"

The latter issue portion is added to cover the case where we somehow miss deploying a package to npm, but the version exists on the `main` branch.

mephisto/abstractions/architects/router/flask/mephisto_flask_blueprint.py
JackUrb: I think one issue with at runtime is if someone is using the PyPI package, they may not have the local dir and package structure to even be loading the `mephisto-task` package from, so I wanted to ensure consistency. I think having a precommit step for `sync_mephisto_task.py` would be sufficient. Do you have a pointer for adding that?

pringshia: PR now available at #676

PR #677, opened 2022-02-17
Mephisto remote query blueprint, example, and package

JackUrb:
# Overview
This pr introduces the `RemoteProcedureBlueprint`, and example task, and a js package to power tasks like it. It's a v0 implementation of the standard suggested in #655, though moved _out_ from the core architect API into just _using_ the tools of `liveUpdates` to power it. It's not _pretty_, but it's functional and powerful, and may prove a good reference for people who are trying to get 'ahead of the curve' of what we offer in more polished forms.

The main idea of the implementation is to establish a list of requests and callbacks on the frontend side, as well as registered functions on the backend/python side, and use a common language to communicate between them. This is done via request packets, which carry a `requestId` for requests and `handles` for responses to that request. Registering callbacks allows us to wait on the response for one of these calls, but by making the registry take `targetEvent` names we provide a more templated way to consider how one would make these requests. 

If someone wants to do something more fancy, they can always return to the base `useMephistoLiveSocket`.

# Implementation
- `mephisto-task`
  - Exposes a wrapper interface around `useMephistoLiveTask` that operates more using requests and callbacks. This way people can create things that _mostly_ appear to be a static task, but have access to the `remoteProcedure` function and can register callbacks to it as well. More details in the README for this.
- `RemoteProcedureBlueprint`
  - Creates the `RemoteProcedureTaskRunner`, which essentially sets up small polling servers for incoming requests per task via `_run_server_timestep_for_agent`. Definitely not an _ideal_ setup, but functional for the initial case and establishes an API we can iterate on if this blueprint sees more use. 
  - Creates a `RemoteProcedureAgentState` that keeps track of all of the incoming requests and outgoing responses over the duration of a task, as well as the initial and final data. Stores these in a `requests` dict that is later indexable.
    - As a note, this dict is currently sorted by the request id key, which is random. It may be nice to sort these by the timestamp instead?
  - Creates the `SharedRemoteProcedureTaskState` which includes the `function_registry`.
  - The core `RemoteProcedureBlueprint` file operates almost directly like a `StaticReactBlueprint`, plus a function registry. Thus there isn't much interesting to discuss here.
- `remote_procedure` template example
  - Creates extremely simple task that registers a single function to the registry, demonstrates how local state can be accessed within it, and then passes a response back.
  - The frontend for the task requires sending a few requests, and then allows a submit afterwards.
- `mnist` example
  - creates an MNIST model-in-the-loop task to evaluate a local model. Leverages `CanvasDraw` element and some `PIL` image processing on the results. Includes some things for rate limiting and such.

# Testing
**mnist task**
Screen Shot 2022-03-08 at 4 20 40 PM
Screen Shot 2022-03-08 at 4 25 35 PM

**template task**
Screen Shot 2022-02-17 at 12 14 16 PM
Screen Shot 2022-02-17 at 12 14 10 PM
Screen Shot 2022-02-17 at 12 21 59 PM


packages/mephisto-remote-query/src/index.js
```diff
@@ -0,0 +1,154 @@
...
+    (previousRequestCallbacks, newRequestCallback) => {
+      let requestId = newRequestCallback.request_id;
+      let newRequestCallbacks = {
+        ...previousRequestCallbacks,
+        [requestId]: newRequestCallback,
+      };
+      return newRequestCallbacks;
+    },
+    {}
+  );
```
pringshia, line 47: If we end up using a ref anyway, there's no real reason to have this reducer here to manage the state. i would just use the ref.current to manage the underlying state, and also just have the `makeRemoteCall` callback use ref.current instead of addRequestCallback in the dependencies array,

packages/mephisto-remote-query/src/ConnectionIndicator.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+  return (
+    <button
+      id="connected-button"
+      className="btn btn-lg connection-indicator"
+      style={{ backgroundColor: backgroundColor }}
+      disabled={true}
+    >
+      {text}
+    </button>
```
pringshia, line 44: I'm not sure how I feel about introducing a JSX dependency here

packages/mephisto-remote-query/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "mephisto-remote-query",
```
pringshia, line 2: As mentioned in our 1:1 earlier, I'm leaning towards keeping this functionality bundled with `useMephistoTask` and `useMephistoLiveTask` in  the`mephisto-task` pkg. It allows folks to easily opt-in to add'l functionality as their tasks evolve in design without installing new deps, it makes operations easier around deploying packages as well as reducing complexity around peer-dep version mismatches w/ `mephisto-task`. Since this package is also pretty lightweight and somewhat related in functionality, I think it's fine to keep it with the other code.

packages/mephisto-remote-query/src/index.js
```diff
@@ -0,0 +1,154 @@
...
+    ...otherMephistoProps
+  } = mephistoProps;
+  let { agentId } = mephistoProps;
+
+  // They're still exposed in a separate manner though
+  const _fullSocketProps = {
+    connect,
+    destroy,
+    sendLiveUpdate,
+  };
```
pringshia, line 104: what if someone wants to use remote procedure calls, while also having a live task. any reason we shouldn't just support that? this means that presenting remote calls as another tool in the toolbelt, instead of a separate modality (different toolbelt altogether).

what this means in practice is something like `useMephistoLiveTask` just exposing a `makeRemoteCall` method, that exists in case people want to use it. is there something i'm missing there, maybe around packets not being differentiated correctly if they're traveling over the same channel?

examples/remote_query/webapp/src/components/core_components.jsx
```diff
@@ -0,0 +1,71 @@
...
+
+      <button
+        className="button"
+        onClick={() => {
+          setQueryCount(queryCount + 1);
+          onRemoteCall({
+            targetEvent: "handle_with_model",
+            args: { arg1: "hello", arg2: "goodbye", arg3: queryCount },
+            callback: (response) => alert(JSON.stringify(response)),
+          });
```
pringshia, line 51: i think this could be a js promise based API. it also would also give some language level support for error handling (and also unlocks the async/await keywords). For example:

```js
onRemoteCall({
  targetEvent: "handle_with_model",
  args: { arg1: "hello", arg2: "goodbye", arg3: queryCount }
})
 .then((response) => alert(JSON.stringify(response)))
 .catch(err => alert(JSON.stringify(err)));
```

I also would suggest an API like the following:

```js
remoteProcedure("run_mnist_classifier")
   .call("my_img.jpg")
   .then(res => console.log(res))
   .catch(err => console.error(err))
```

This opens up possibilities like this where remote calls are made frequently. Here they almost look as if you're calling native js functions:

```js
const classifyNumber = remoteProcedure("run_mnist_classifier").call;

const result = await classifyNumber("my_image.jpg");
const result2 = await classifyNumber("my_image2.jpg");
```

which looks pretty nice.

packages/mephisto-remote-query/src/ConnectionIndicator.jsx
JackUrb: Happy to leave out, we don't even use it in an example at the moment.

packages/mephisto-remote-query/src/index.js
JackUrb: > maybe around packets not being differentiated correctly if they're traveling over the same channel?

Yeah this is the exact issue, the current setup requires that people use an extension of the `RemoteQueryBlueprint` (or something that implements its API) in order to properly use this package, as that blueprint converts the `liveUpdate` format to be for RPCs. `useMephistoLiveTask` is a strict superset of this functionality, so I'm not sure what the use case would be for allowing both at the same time.

examples/remote_query/webapp/src/components/core_components.jsx
JackUrb: This looks much nicer to me, though I imagine we still need to pass the arguments. Or is there a way to specify the arguments to repackage later?

pringshia: yeah the idea here is to that remoteProcedure just returns a function reference and then invoke is what passes the argument and makes the socket call, returning a promise. Is this what you mean?

JackUrb: Oh I mean that we need the value that's sent across to be an object with argument names, like:
```
remoteProcedure("fn").invoke({val1: 1, val2: 2, val3: 3});
```

pringshia: ah yeah, I see the type signature is `args: Dict[str, Any]` in handle_with_model, in that case we would enforce an object param, correct. if it was just `args: Any[]` then we could support both `invoke(1,2,3)` or `invoke({a: 2})` with the retrieval being `args[0], args[1], args[2]`, and `args[0].a` for the latter. Don't really have a strong preference for ergonomics here, happy to keep the current type signature and force a naming of keys.

examples/remote_query/mnist/webapp/src/app.jsx
```diff
@@ -0,0 +1,95 @@
...
+    isPreview,
+    previewHtml,
+    isLoading,
+    handleSubmit,
+    remoteProcedure,
+    isOnboarding,
+    handleFatalError,
+  } = mephistoProps;
+
+  const classifyDigit = remoteProcedure("classify_digit");
```
pringshia, line 50: we say `remoteProcedure` here but `useMephistoRemoteQueryTask`, should we align on either query or procedure? unless there's some reason to have differentiation

examples/remote_query/template/README.md
```diff
@@ -0,0 +1,9 @@
+# Mephisto Remote Query Template
+
+This task is an _extremely rough_ bare-bones example of being able to set up a task where the frontend directly connects to the backend using the `mephisto-remote-query` package. It should serve as a decent example of how to get a `RemoteQueryBlueprint` task up off the ground.
```
pringshia, line 3: need to update now that there's no `mephisto-remote-query` package

pringshia (approved):
A few small notes but otherwise looks good

examples/remote_query/template/webapp/src/app.jsx
```diff
@@ -0,0 +1,99 @@
...
+
+import React from "react";
+import ReactDOM from "react-dom";
+import { BaseFrontend, LoadingScreen } from "./components/core_components.jsx";
+
+import {
+  MephistoContext,
+  useMephistoRemoteQueryTask,
+  ErrorBoundary,
+} from "mephisto-remote-query";
```
pringshia, line 17: outdated now that there's no `mephisto-remote-query`

pringshia:
P.S. The MNIST example is awesome!

examples/remote_query/mnist/webapp/src/app.jsx
JackUrb: I can convert to `remoteProcedure` for all, this is a reasonable suggestion!

examples/remote_query/template/README.md
JackUrb: Ah right! Spent a lot of time testing MNIST and less on the template...

PR #678, opened 2022-02-22
Add GH action for checking npm version consistency on merge to main

pringshia:
## Overview

On PR merges into the `main` branch, we run a GH action to check that packages on npm `mephisto-task@latest` and `bootstrap-chat@latest` match the version in the repo. 

## Details

This is done to reduce a class of errors where the main repo is out of sync with npm, e.g. when we fix a bug in the repo and tell users to update their packages to incorporate the fix.

This is also a stopgap measure until we roll out an automated npm deploy process. IMO it is a sufficiently effective one and should catch and alert for most errors with minimal overhead created.

#### Side note:

It also doesn't consider the case where we make edits to `mephisto-task` and `bootstrap-chat` but we forget to update the version number.

Perhaps in the future we want to update the workflow to also run the following test:

```
[[ $(git diff ${{ github.head_ref }}..${{ github.base_ref }} --name-only -- packages/ | wc -l) -ne 0 ]]
```

...and if true, check that a package version change has occurred.

There may also be easier GH workflow-esque ways of doing this. Would need to do some research.

## Testing

Note that commit [86cfd51](https://github.com/facebookresearch/Mephisto/pull/678/commits/86cfd51a098727a122b8018d1ceca84e2601d8e4) below has the `npm-check` test run, while after commit [**ddf6271**](https://github.com/facebookresearch/Mephisto/pull/678/commits/ddf62714879cf832d071e5ec96f345d880e7d25c) is introduced, the test doesn't run, since that commit introduces the run-when-merging-against-`main` branch functionality.

Ran the `/scripts/check_npm_package_versions.py` file locally, including while internet was disconnected to verify error when npm can't be accessed.

### DIFF
--- a/.github/workflows/npm-check.yml
+++ b/.github/workflows/npm-check.yml
@@ -14,4 +14,4 @@ jobs:
       - name: Install Mephisto
         run: pip install -e .
       - name: Run version sync script
-        run: python scripts/check_npm_package_version.py
+        run: python scripts/check_npm_package_versions.py
--- a/.github/workflows/npm-check.yml
+++ b/.github/workflows/npm-check.yml
@@ -1,8 +1,7 @@
 name: npm-check
 
 on:
-  push:
-  pull_request:
+  pull_request_target:
     branches: [main]
 
 jobs:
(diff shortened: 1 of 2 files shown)


JackUrb (approved):
I love these improvements to dev experience - thanks for putting this script and action together!

PR #683, opened 2022-02-25
Add new workflows for deploying docs

pringshia:
Adds a test and deploy workflow for auto-deploying docs to https://mephisto.ai:

https://docusaurus.io/docs/deployment#triggering-deployment-with-github-actions

## Details:

- A build test runs only when changes are made in `docs/web/**` in a PR targeting `main`
- Auto-deploy is run whenever there are changes in `docs/web/**` pushed on to `main`

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -1,0 +1,40 @@
+name: Deploy docs to GitHub Pages
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  deploy:
+    name: Deploy docs to GitHub Pages
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 16.x
+          cache: yarn
+
+      - name: Move to docs directory
+        run: cd docs/web
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+      - name: Build website
+        run: yarn build
+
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,40 @@
...
+        run: yarn install --frozen-lockfile
+      - name: Build website
+        run: yarn build
+
+      # Popular action to deploy to GitHub Pages:
+      # Docs: https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-docusaurus
+      - name: Deploy to GitHub Pages
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
JackUrb, line 31: Do we have a token configured already?

JackUrb (approved):
Great to have this as an action! Just one question on secrets.

.github/workflows/deploy-docs.yml
pringshia: I believe it is auto-configured via Github Actions.

> For newbies of GitHub Actions: Note that the GITHUB_TOKEN is NOT a personal access token. A GitHub Actions runner automatically creates a GITHUB_TOKEN secret to authenticate in your workflow. So, you can start to deploy immediately without any configuration.

\- [README from the actions-gh-pages package](https://github.com/peaceiris/actions-gh-pages#supported-tokens)

PR #687, opened 2022-03-01
Mephisto metrics overhaul [feat Prometheus/Grafana]

JackUrb:
# Overview
Introducing Mephisto, now with actual metrics. Using Prometheus and Grafana, we can now setup a dashboard (one that can be customized too!) to monitor a job over its lifecycle. The primitives set here _are_ somewhat rigid (only one operator can have metrics at a time, ports are preset, etc.) but can certainly be extended in the future.

# Implementation
Adds individual metrics for:
- keeping track of `MephistoDB` database latency (which required making the implementations `_<internal` so that the API layer is responsible for tracking).
- `ClientIOHandler` response latency per packet
- changes to the `mephisto-task` package and router servers to include timestamps that `ClientIOHandler` can log as well
- active task count and runtime durations in the `TaskRunner`
- current active `Agent` statuses, and `Agent`s per worker (stratified by onboarding or in-task)
- current active `Unit` statuses (stratified by unit type (gold, validation, regular))
- Onboarding outcomes, responses to agent requests, and latency of various function in `WorkerPool`

For actually installing the requirements (Prometheus and Grafana both need binaries), includes an `install_metrics.sh`  script that download and unpack requirements, and then load in default contents from `resources` to initialize the Prometheus data source and default mephisto dashboard.

Creates utility scripts for managing whether or not the metrics servers are up and on what process in `mephisto.utils.metrics`, allowing us to build out the interface for this functionality as we move to support multiple instances, greater configuration, etc.

Creates scripts for launching and cleaning up the servers in `view_metrics.py` and `shutdown_metrics.py`. Links functionality into the cli:
```
>>> mephisto metrics <install|view|cleanup>
install: Installs prometheus and grafana to /Users/jju/mephisto/metrics
view: Launches a Prometheus and Grafana server, and shuts down on exit
cleanup: Shuts down prometheus and grafana resources that may have persisted
```

# Testing
Automated tests passing
Ran a job without launching the server beforehand, asserted that the details got logged and could be checked on later with `view_metrics.py` or `mephisto metrics view`:
```
> (py38) root-mbp:metrics jju$ python view_metrics.py
Waiting for grafana server to come up.
Dashboard is now running, you can access it at localhost:3032/d/v6aIqWYnk/default-mephisto-monitoring
Once you're no longer using it, and no jobs need it anymore, you can shutdown with `shutdown_metrics.py`
```
Screen Shot 2022-02-28 at 3 36 13 PM
```
(py38) root-mbp:metrics jju$ python shutdown_metrics.py 
Using stored PIDs to shutdown metrics servers
Servers shut down!
```

Ran `mephisto metrics cleanup` during a task run, ensured that metrics were still logged (as the server was relaunched).


mephisto/data_model/agent.py
```diff
@@ -9,6 +9,7 @@
 from queue import Queue
 from mephisto.tools.misc import warn_once
 from uuid import uuid4
+from prometheus_client import Gauge
```
pringshia, line 12: is there any way to use a plugin architecture-type pattern here so that the prometheus dependency is opt-in and not a hard requirement. Perhaps we can have Gauge also be a noop if prometheus isn't found? I ask this to push these core data_model classes to be as lean and dependency-free as possible. I do understand the need for adding the prometheus_client here though, but perhaps it shouldn't be a rigid requirement for running Mephisto?

pringshia:
😍😍😍

mephisto/data_model/agent.py
JackUrb: I'm of two minds on this: On the one hand, I agree that we shouldn't be introducing dependencies throughout that aren't required for Mephisto to run. On the other hand, I think the value add of this kind of monitoring (especially once we add alerts and start having health checks and dependencies on these metrics (for instance, automatic scaling of `max_num_concurrent_units` based on server health)) means it _should_ be the default, and I worry making it opt-in would mean that the shortest path to average-case functionality is harder.

Perhaps there's a way to make it _opt-out_? I'm not sure what the build pattern for something like this is... But certainly I could move the prometheus classes into `mephisto.utils.metrics` and create no-op wrappers there if imports fail (though _that_ may also be complicated, as `Gauge` and other classes are classes and functions and also `with` contexts as well as decorators...).

I do want to make sure I'm not making this decision just because making it optional is more complex however.

pringshia: I think I'm reluctantly OK with making Prometheus a first class dependency, as the dashboard does seem useful. I did some research a few days back. In some ways Prometheus works well here, but in others it's not a perrrrfect fit. In the ideal case, I think Prometheus likes to be running all the time while polling long-running tasks (usually dedicated service instances). For things that are short-lived "batch jobs" (which maybe is a task run? idk.) they mention using something called the [Pushgateway](https://prometheus.io/docs/practices/pushing/#when-to-use-the-pushgateway). Not sure if you've looked into this.

I was also trying to see if there's a way for a Mephisto run to output it's metrics to a data file, which could later be replayed into Prometheus, but I found nothing like this after a bunch of research. Am I correct here? Is there no way to do a "replay", i.e. persist/share logging data?

JackUrb: > In some ways Prometheus works well here, but in others it's not a perrrrfect fit.

Agreed especially for the simple case of short pilot jobs, but logging every 5-10m over the course of a 3 day job, or (in the future) monitoring 3-4 different simultaneous jobs seems to work.

> For things that are short-lived "batch jobs" (which maybe is a task run? idk.) they mention using something called the [Pushgateway](https://prometheus.io/docs/practices/pushing/#when-to-use-the-pushgateway).

Yup did some reading on this - in the end I think going with a faster polling rate ends up better than the benefits that the `Pushgateway` really provides. From their docs: 'A "service-level" batch job is one which is not semantically related to a specific machine or job instance' - all of the metrics for a specific TaskRun have an associated 'job' on the Prometheus side. It also still requires a running server.

> Am I correct here? Is there no way to do a "replay", i.e. persist/share logging data?

There _is_, but it's a little silly (still intending to support it in a follow-up in 1.1 though): The prometheus *server* is able to do a dump that can be later imported, so the procedure for this would be to run a server during launch (which we automatically do) and then dump the metrics from the server that are tagged with the specific task run on close (which we don't do yet, but can).

JackUrb: > Assuming Prometheus is a hard dependency for all Mephisto runs - how does this affect the user, if in any way? Is installing Prometheus first via install_mac.sh or install_linux.sh first a MUST? If so that is a severe detriment to the new user experience.

It's not a must - the system will run but skip logging. There will be a warning to this effect:
```
if not os.path.exists(PROMETHEUS_EXECUTABLE):
        logger.warning(
            f"Cannot collect and display metrics without installing metrics. "
            f"Use the scripts in the metrics dir ({METRICS_DIR}) to install. "
        )
        return False
```
I can probably soften this a little bit if you think it's necessary though.

> What is the story around when a user accesses the data and how? Can they access it after the task ends? Is there something they have to do before launching a task to opt-in to the metrics, or no? You touch a bit on this in your testing section, but i'm still unclear on what happens after the run ends? Is it too late at that point?

So I've set this up so that, if the user has run one of the metrics install scripts, upon launching a job we also launch the relevant server. This server will persist (and the data will persist after the job is shutdown), so then they can use the `view_metrics.py` script to see the results.

pringshia: Gotcha, so if I understand correctly: Once Prometheus is installed, the first run launches the Prometheus server. The server stays alive after the job ends, and then must be manually closed.

Correct me if I'm wrong: Subsequent jobs will reuse the existing running server if it's already up? If someone launches Job A and then Job B, we still differentiate the metrics instead of commingling right? What happens if people run multiple jobs at a time - do we support that easily or do they have to do some configuration?

mephisto/utils/metrics.py
```diff
@@ -0,0 +1,206 @@
...
+                logger.warning(
+                    "Clearing prometheus pid as the server isn't running. "
+                    "Use `shutdown_prometheus_server` in the future for proper cleanup."
+                )
+        else:
+            logger.info("Prometheus server appears to be running at 9090")
+            return True
+    if not os.path.exists(PROMETHEUS_EXECUTABLE):
+        logger.warning(
+            f"Cannot collect and display metrics without installing metrics. "
```
pringshia, line 105: Should this read "without installing Prometheus" instead of "without installing metrics"?

Yeah I think I would rephrase this as more of a positive offering than a lack of a feature. For example a message like:

"Mephisto supports rich run-time metrics visualization through Prometheus. If you'd like to use this feature, use the scripts in the metrics directory <...> to install"

Addendum: What happens when people install Mephisto via pip, they won't have access to the metrics directory in that case then right? In that case mentioning the metrics dir could be confusing to them.

JackUrb: > What happens when people install Mephisto via pip, they won't have access to the metrics directory in that case then right

They do have access, but it may not be where they expect. This is why I get `METRICS_DIR`. I imagine we'll want to add things to the `mephisto` cli to access metrics in the future though.

pringshia: Wondering if we could quickly augment the CLI to support `mephisto script view_metrics` which will use the logic here: https://stackoverflow.com/a/5790381 to run the script file from within the package.

mephisto/data_model/agent.py
JackUrb: > If someone launches Job A and then Job B, we still differentiate the metrics instead of commingling right? 

Correct, upon completion of Job A, the metrics will be separate on Job B

> What happens if people run multiple jobs at a time - do we support that easily or do they have to do some configuration?

This will come in 1.1

mephisto/utils/metrics.py
JackUrb: Added scripts to the CLI:
```
>>> mephisto metrics
Usage: mephisto metrics <install|launch|cleanup>
install: Installs prometheus and grafana to /Users/jju/mephisto/metrics
view: Launches a Prometheus and Grafana server, and shuts down on exit
cleanup: Shuts down prometheus and grafana resources that may have persisted
```

mephisto/client/cli.py
```diff
@@ -319,9 +319,9 @@ def metrics_cli(args):
...
     )
 
-    if len(args) == 0 or args[0] not in ["install", "launch", "cleanup"]:
+    if len(args) == 0 or args[0] not in ["install", "view", "cleanup"]:
         click.echo(
-            "Usage: mephisto metrics <install|launch|cleanup>\n"
+            "Usage: mephisto metrics <install|view|cleanup>\n"
             f"install: Installs prometheus and grafana to {METRICS_DIR}\n"
             f"view: Launches a Prometheus and Grafana server, and shuts down on exit\n"
             f"cleanup: Shuts down prometheus and grafana resources that may have persisted"
```
pringshia, line 327: small nit: inconsistent capitalization. Prometheus and Grafana not capitalized for `install`/`cleanup` but are for `view`

pringshia (approved):
LGTM. Thanks for making this happen. Should be extremely useful for folks.

PR #690, opened 2022-03-02
Resolving or dropping todos related to #93

JackUrb:
# Overview
Picks up some remaining logging TODOs from #93. Overall most of these have been resolved in some form in other places, or are usurped by the broad metrics capabilities introduced by #687.

# Testing
Automated tests

### DIFF
--- a/mephisto/abstractions/databases/local_database.py
+++ b/mephisto/abstractions/databases/local_database.py
@@ -255,10 +255,6 @@ class LocalMephistoDB(MephistoDB):
         """
         Run all the table creation SQL queries to ensure the expected tables exist
         """
-        # TODO(#93) maybe raise flag when the schema of existing tables isn't what we expect
-        # it to be?
-        # "How to know that schema changes?"
-        # logger.warning("some message")
         with self.table_access_condition:
             conn = self._get_connection()
             conn.execute("PRAGMA foreign_keys = 1")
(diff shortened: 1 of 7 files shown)


mephisto/abstractions/databases/local_database.py
```diff
@@ -255,10 +255,6 @@ def init_tables(self) -> None:
         """
         Run all the table creation SQL queries to ensure the expected tables exist
         """
-        # TODO(#93) maybe raise flag when the schema of existing tables isn't what we expect
-        # it to be?
-        # "How to know that schema changes?"
-        # logger.warning("some message")
```
pringshia, line 261: what happens now in the event of a schema change?

pringshia (approved):

pringshia:
The version sync test above is failing with `ModuleNotFoundError: No module named 'prometheus_client'` any idea why that may be?

mephisto/abstractions/databases/local_database.py
JackUrb: We're launching 1.0 and I'm not making any schema changes anymore

JackUrb:
I must not have dependencies quite right on the `prometheus-metrics` branch. Will resolve.

pringshia:
I would expect the Python test to also fail here, interesting that it doesn't

JackUrb:
I think the python test is relying on `requirements.txt` which I updated, but github actions are relying on `pyproject.toml` which I missed.

PR #694, opened 2022-03-04
Cleanup - Utils and general housekeeping

### LINKED ISSUE #98: [TODO Roundup] Utils cleanup
# Summary
Mephisto doesn't have a clear internal utility structure for code and functions that exist outside of the main functionality. We can consider two primary divisions of utilities:
1) Those that are useful for end-users to be able to do special functionality inside of their own scripts and classes
2) Those that are useful for Mephisto internals and aren't meant to be used by the majority of users.

I propose splitting these up into `tools` and `utils` respectively. Prior to 1.0, it would be relevant to do a pass through the codebase to find functionality that belongs in one of these two, and to make the move.

JackUrb:
# Overview
In preparation for 1.0, this PR includes changes to finally organize our utility functions as noted way back in #98. Everything is grouped into `tools` (which are for Mephisto user's use) and `utils` which are utility functions for Mephisto developers to use. 

Alongside these changes, I've also fixed some documentation issues surrounding the `Supervisor`, cleaned up some `mypy`-related typing issues, and updated callsite locations.

Closes #98

# Details
- Fills out the `utils` directory with `logger_core.py`, `dirs.py`, `testing.py`, and `qualifications.py`. These join the `metrics.py` file that was introduced in #687. Updates documentation to talk about each. Most of these came from the misc utils in the `operations.utils` file, however hydra-specific configuration utilities have ended up remaining in `hydra_config.py`.
- Update callsite locations to all of the newly moved utils.
- `mypy` fixes, mostly adding typing to the data model and `#type: ignore` for new things that have been introduced w/o type files. Notable exceptions in `TaskLauncher` to ensure the correct iterator type for places we expect it to be a generator, as well as in `UseGoldUnit` mixin to use the newer function interface. 

# Testing
Automated tests pass
```
>> mypy mephisto test examples
Success: no issues found in 181 source files
```


mephisto/abstractions/blueprints/mixins/use_gold_unit.py
```diff
@@ -25,7 +25,7 @@
 from dataclasses import dataclass, field
 from omegaconf import MISSING, DictConfig
 from mephisto.data_model.qualification import make_qualification_dict, QUAL_NOT_EXIST
-from mephisto.operations.utils import find_or_create_qualification
+from mephisto.utils.misc import find_or_create_qualification
```
pringshia, line 28: I once read an angry rant about someone hating on `utils` lol: https://breadcrumbscollector.tech/stop-naming-your-python-modules-utils/

Not sure I completely agree with everything said there. That said maybe we should avoid the `misc` subsection for utils?


Some suggestions:
```
ensure_user_confirm -> utils.prompt
    - there's some similarity with logger_core, e.g. format_loud, but not entirely so didn't pick utils.logger_core
get_mock_requester -> utils.testing
warn_once -> utils.logger_core
find_or_create_qualification -> ?
    - signature looks similar to things like Worker.get(), etc. is there a Qualification class? e.g. Qualification.get()
```

JackUrb: Yeah I can break this out further, this point is fair

mephisto/abstractions/providers/mturk/mturk_requester.py
```diff
@@ -15,7 +15,7 @@
     setup_aws_credentials,
     get_requester_balance,
     check_aws_credentials,
-    find_or_create_qualification,
+    find_or_create_qualification as find_or_create_mturk_qualification,
```
pringshia, line 18: Just curious, is this alias just for clarity? My other guess would be to avoid a namespace collision but I don't see any possibility of that in the file

JackUrb: It's for clarity - `find_or_create_qualification` in _most_ of the mephisto codebase refers to the `mephisto.utils.qualifications` version, while this one comes from the mturk utils created for this provider.

pringshia (approved):
LGTM

PR #697, opened 2022-03-09
Fill Review CLI section in Custom React docs page, add default UI support

pringshia:
Mostly docs as mentioned in the title.

Also updates the review CLI to ship with a default view when no local interface path is specified. This makes for a much smoother, batteries-loaded initial user experience.

```bash

# In practice this means you can now just run this (no REVIEW_APP_DIR arg):
mephisto review --db custom-react-tutorial-iterating --stdout --all

# instead of something like this
mephisto review review-app/build/ --db custom-react-tutorial-iterating --stdout --all
```


JackUrb (approved):
I love the default reviewer, and thanks for updating the docs with it!

Issue #679, opened 2022-02-22
Add docs on how to get EC2Architect setup


pringshia:
https://mephisto.ai/docs/guides/quickstart#optional-set-up-heroku

^- we have instructions for Heroku setup, but not for EC2

PR #701, opened 2022-03-10
Missing 1.0 docs

JackUrb:
# Overview

As one of the last steps for Mephisto 1.0, this PR runs through some remaining docs pages that need content:
- [x] [worker_controls](https://mephisto.ai/docs/guides/tutorials/worker_controls)
- [x] [workflows](https://mephisto.ai/docs/guides/tutorials/workflows)
- [x] metrics tutorial
- [x] [setting up hydra profiles](https://mephisto.ai/docs/guides/how_to_use/efficiency_organization/reusing_configs)
- [x] [Automated API docs for some things](https://mephisto.ai/docs/reference/overview)

To come shortly after 1.0 would be the following docs:
- [x] [in-depth onboarding usage](https://mephisto.ai/docs/guides/how_to_use/worker_quality/using_onboarding/)
- [x] [in-depth validation usage](https://mephisto.ai/docs/guides/how_to_use/worker_quality/using_validation/)
- [x] [in-depth gold label usage](https://mephisto.ai/docs/guides/how_to_use/worker_quality/using_golds/)
- [ ] [ec2 architect](https://github.com/facebookresearch/Mephisto/issues/679)

For these de-prioritized docs, what would be the appropriate messaging to leave on the pages?

For autogenerated python docs, I've used `pdoc` and dropped the static docs into static. This included a number of changes to `__init__.py` files to autoinclude readmes. It also moved a lot of scripts behind `__main__` blocks, as these were being executed during the doc build process.

Example API docs:
Screen Shot 2022-03-11 at 1 58 42 PM

Note: the ergonomics aren't _great_ as this triggers a new tab load instead of being in the same tab, but I don't have any easy fix for this just yet.


docs/web/docs/guides/how_to_use/efficiency_organization/reusing_configs.md
```diff
@@ -4,4 +4,61 @@ sidebar_position: 1
...
+```
+
+Then augmenting your launch configs is as easy as doing:
+```
+python run_task.py +profile=local_testing
+...
+python run_task.py +profile=live_launch
+```
+
+Using `profile` can be an effective way to simplify the configuration for your most common workflows.
```
pringshia, line 64: Woah, didn't know this... so cool!

docs/web/docs/guides/how_to_use/efficiency_organization/reusing_configs.md
```diff
@@ -4,4 +4,61 @@ sidebar_position: 1
...
 
-TODO - talk about setting up hydra profiles
+As you begin launching many Mephisto tasks, you may find that there are some specific argument options that you frequently use _across_ multiple tasks. Mephisto provides a way to reuse these types of configurations with `profile`s.
+
+Setting up profiles is pretty easy, and makes it so that you don't have to be writing architect and crowdprovider args on every launch:
+```
+python run_task.py mephisto/architect=heroku mephisto/provider=mturk_sandbox mephisto.provider.requester_name=MY_REQUESTER
+```
+
+Instead you can move these common configurations into a file in your `~/.mephisto/hydra_configs/profile` dir.
```
pringshia, line 14: Maybe a provocative question, but is there any reason to have `hydra_configs` as part of the path? In the example folders we also have things like `parlai_chat_task_demo/hydra_configs/conf/`.... always as a nested folder. seems unnecessary to have the double besting. Perhaps `hydra_config` is part of the defaults but maybe we can use something straightforward here? this should also be easier now that we control the decorator.

Perhaps just `conf/`, e.g `~/.mephisto/conf/` that houses the profiles? or `~/.mephisto/profiles/`

pringshia:
So I see `"docs:static": "pdoc mephisto --logo https://mephisto.ai/img/logo.svg -o docs/web/static/python_api/ -t docs/web/pdoc_src"` in the root `package.json` but don't see the dependency specified anywhere? Were you installing pdoc through npm or python?

docs/web/docs/guides/how_to_use/efficiency_organization/reusing_configs.md
JackUrb: It's a hydra requirement, you need to point to a root folder, and then all sub folders of that folder are possible configuration options that can be set. As such it needs its own folder, or else hydra will consider all of the things present possible arguments. Agreed that it's not ergonomic, but I'm not sure how else we do it.

JackUrb:
> It would be nice to have better pdoc integration w/ Docusaurus, [like this example](https://wave.h2o.ai/docs/api/ui). Seems like this is the template they use: https://github.com/h2oai/wave/blob/master/py/docs/templates/text.mako
> 
> and this is how it's built via the `--template-dir` command:
> 
> https://github.com/h2oai/wave/blob/master/py/Makefile#L25-L31
> 
> I can take a look at wiring this up this weekend.

I spent about an hour or two walking through this rabbit hole - this is a super nice experience, but `pdoc` doesn't support output to markdown. `pdoc3` does, but is a fork of the original that isn't as well maintained and breaks down doing the conversion. I tried to make the markdown versions in 2-3 different ways using `pdoc3`, but each failed around a number of issues around typing documentation generating items of the form `<type>` which the html->markdown software totally dies on. If you want to submit a fix to `pdoc3` at some point around silently dropping failed tags though I'm sure we can change back over, but I don't think it's worth your time compared to this.

JackUrb:
> So I see:
> 
> ```
> "docs:static": "pdoc mephisto --logo https://mephisto.ai/img/logo.svg -o docs/web/static/python_api/ -t docs/web/pdoc_src"
> ```
> 
> in the root `package.json` but don't see the dependency specified anywhere? Were you installing pdoc through npm or python?

It's installed through python, I didn't know the right place to put this otherwise if our other scripts are here.

pringshia:
Saw your message too late, that's a few hours I'll never get back 😆 Man what a mess trying to use pdoc3.

pringshia:
> It's installed through python, I didn't know the right place to put this otherwise if our other scripts are here.

Unsure whether dev dependencies like this belong in the requirements.txt. Can't think of any other place so leaning towards yes?

pringshia:
> this triggers a new tab load

FYI: https://github.com/facebook/docusaurus/issues/3309#issuecomment-893839723

mhils:
Hi folks! I found this thread as I monitor potential pdoc issues in downstream dependencies via GitHub search.

> I spent about an hour or two walking through this rabbit hole - this is a super nice experience, but `pdoc` doesn't support output to markdown.

I have no experience with docusaurus, but FWIW many Markdown-based frameworks will happily accept a blob of HTML in Markdown files. pdoc does that quite happily, all CSS is specifically scoped to not affect the rest of the page. You can find an example for mkdocs at https://github.com/mitmproxy/pdoc/tree/main/examples/mkdocs. We follow the same approach for mitmproxy ([demo](https://docs.mitmproxy.org/dev/api/mitmproxy/http.html)). 

We also have some users that use a custom template to emit Markdown (https://github.com/codeghetti/seagulls-py/tree/main/.pdoc/templates), but ultimately it doesn't end up looking as nicely.

pringshia:
Thanks for the pointers @mhils - glad to know of the different approaches here. @JackUrb - I'm happy to revisit these approaches in a future update. As of now, I think this works for the 1.0 release, especially with the `_target` link fix.

pringshia (approved):

JackUrb:
I did a quick attempt at the implementation that `mitmproxy` uses, but found that Docusaurus isn't quite happy to have embedded HTML as it treats markdown files as `React`. In any case, we were able to find an implementation that works serving the static HTML on its own 'sub-site' of our docs. We can revisit the full customization choice that `seagulls-py` chose in the future if it turns out there's a strong need to colocate the API docs within the Docusaurus react tree (perhaps for automatic linking or something) in the future.

Issue #700, opened 2022-03-10
Update references to `Supervisor`?


pringshia:
Now that the `Supervisor` abstraction has been replaced with the `ClientIOHandler` and `WorkerPool` (#514), maybe we can update some of the remaining references (usually in comments in the codebase):

(The Supervisor is also mentioned in the [Explanations docs](https://mephisto.ai/docs/explanations/architecture_overview#supervisor))

### References

https://github.com/facebookresearch/Mephisto/blob/904b4e761354937f0d83ff519d70379e9102706e/mephisto/abstractions/_subcomponents/agent_state.py#L77

https://github.com/facebookresearch/Mephisto/blob/e824a99c13c7bdd7bee637ecf2ddf7a18ff7411d/mephisto/abstractions/_subcomponents/channel.py#L66

https://github.com/facebookresearch/Mephisto/blob/79314456c2fadc1225dd71add8a6c8082c3198aa/mephisto/abstractions/architects/README.md#L11

https://github.com/facebookresearch/Mephisto/blob/9a2b63053d111d0d1099a1771a09288116789fad/mephisto/abstractions/architects/channels/websocket_channel.py#L62

https://github.com/facebookresearch/Mephisto/blob/904b4e761354937f0d83ff519d70379e9102706e/test/abstractions/blueprints/test_mock_blueprint.py#L30

Perhaps this test should just be called `test_initialize_operator`:

https://github.com/facebookresearch/Mephisto/blob/384457950bd8725d9c9c0b453ad9a4736fc0246a/test/core/test_operator.py#L105

https://github.com/facebookresearch/Mephisto/blob/384457950bd8725d9c9c0b453ad9a4736fc0246a/test/core/test_operator.py#L143

https://github.com/facebookresearch/Mephisto/blob/384457950bd8725d9c9c0b453ad9a4736fc0246a/test/core/test_operator.py#L203

PR #702, opened 2022-03-10
Removing (almost all) supervisors

JackUrb:
I missed a few (lowercase) mentions of supervisor when I last attempted to remove all mentions. This leaves the only one in the Architecture Overview, which would require recreating the flow charts (out of scope for right now).

Thanks for finding in #700


pringshia (approved):
Thanks for the fixes!

PR #703, opened 2022-03-10
Fixing bugs related to installing Mephisto metrics

JackUrb:
# Overview
Made a small mistake merging the two scripts into one, both in forgetting to update the load directory and improperly assigning variables. I also forgot to specify what to process the file with, as this isn't required on the command line. 

# Testing
Deleted my local metrics dirs, reinstalled via the bash script, the cli, and the python util.


pringshia (approved):
Thanks for the quick fix

PR #704, opened 2022-03-10
Fixing Remote procedure on reload bug

JackUrb:
# Overview

I wasn't properly encoding remote procedure previous requests for delivery on reconnect, this PR fixes it. 

# implementation

We keep the requests in a dict with query key -> `RemoteRequest`, which can't be directly serialized. This converts it to a list, then sorts by arrival time, then converts to json-serializable dicts.

It also includes a note about this behavior in the docs.

# Testing

Ran the Mnist task, drew some digits, and refreshed. Found the previous requests in the state.


pringshia (approved):
👍

PR #717, opened 2022-03-21
Add 1.0 blog post, and Upgrade Guide checklist

pringshia:
Screen Shot 2022-03-21 at 6 51 03 PM

Screen Shot 2022-03-21 at 6 50 52 PM


JackUrb (approved):
Looks good to me, thanks for putting it together! I'll do a last pass after this is in but I think we'll be ready to go!

PR #728, opened 2022-03-31
Fixing package bug breaking liveUpdate tasks over https

JackUrb:
# Summary

Just changing `wws` => `wss` as it should have been. Unsure how I missed this in the test plan. Includes updates to prepare Mephisto for a 1.0.1 release and tag.

# Implementation

Mostly updating our packages to now refer to 2.0.1 and 1.0.1, other than a single character change.

# Testing

Ran the task in sandbox that was having issue before, it no longer does with 2.0.1

Thanks @jxmsML for surfacing the issue.


pringshia (approved):

JackUrb:
@pringshia  any idea why the mephisto-task-latest check is failing on this pr? I know pre-commit black is failing due to a circle error only recently resolved, but don't see what's wrong otherwise here. Will still merge as the referenced files in this PR have the new 2.0.1 version in the `package.json` files.

pringshia:
@JackUrb I'm not sure why the commit at [`9448ef28949b63e83a309c9d5cfbb6e4f11b2968 `](https://github.com/facebookresearch/Mephisto/runs/5779746328?check_suite_focus=true#step:2:459) is getting pulled down during that workflow.

And that commit has the older version of the package defined: https://github.com/facebookresearch/Mephisto/blob/9448ef28949b63e83a309c9d5cfbb6e4f11b2968/packages/mephisto-task/package.json#L3

pringshia:
@JackUrb - Only the remote procedure and static_react_task's packages were updated to 1.0.1 in this PR. What about parlai chat?

Issue #726, opened 2022-03-29
[Ongoing] Add an e2e integration test


pringshia:
Starting place could be for the static task or React task.

Initial deliverable could be a project plan or scope definition.

We can also reference the Cypress integration tests added with https://github.com/fossasia/visdom/pull/850

Etesam913:
#795 Is working on this

Etesam913:
Required branches to merge in to complete this
- [ ] #795 
- [ ] #800 
- [ ] #805

PR #743, opened 2022-04-02
[fix] Update `style-loader` to unbreak UI builds w/ newer webpack

### LINKED ISSUE #742: Frontend UI interface of chat task not loading properly for custom bundles
Hi,

Thank you for releasing Mephisto 1.0.0. I am currently trying to set up a [chat task](https://github.com/facebookresearch/Mephisto/tree/main/examples/parlai_chat_task_demo) but I am facing UI issues when I try to launch the task with custom bundles.

From within `Mephisto/examples/parlai_chat_task_demo`, I first use the following command
```
python parlai_test_script.py
```

On opening the link from the console logs, the interface looks something like this
Screen Shot 2022-04-01 at 6 34 06 PM

To try a custom build, I first run` npm install; npm run dev` from the`webapp` folder, then I run the following command
```
python parlai_test_script.py conf=custom_prebuilt
```

Now the UI interface gets messed up and looks like this
Screen Shot 2022-04-01 at 6 35 56 PM

The issue is reproducible on both Safari and Chrome web browsers. Neither the frontend nor the backend console shows any error.

Mentioning all relevant versions here
```
python --version
Python 3.7.13
npm -v
8.5.0
nvm -v
0.39.1
node -v
v16.14.2
```

From inside the `webapp` folder
```
npm ls mephisto-task
parlai-mturk-task-compiler@1.0.0 /Users/vaibhav.adlakha/ComplexQA-Mephisto/examples/complex_qa_task/webapp
└─┬ bootstrap-chat@2.0.0
  └── mephisto-task@2.0.0
```

I think this is a version issue. If the custom bundle interface on your end looks fine, can you let me know versions of npm, nvm, and node that you are using?

pringshia:
## Overview

Fixes #742 

A few example frontends had an older version of the `style-loader` dependency. Updates the `style-loader` from v0.23.x to v1.x to be compatible with the newer webpack 5 introduced with Mephisto 1.0.

Note: We may want to upgrade to `style-loader@latest` which is at v3 right now. However, other packages that use v1.x in the repo work fine so opting to use that for now for the quick fix, until we can test with the more later version. Ideally #726 would help with this.

## Testing

Ran the repo steps in #742 and confirmed that styles load now.

### DIFF
--- a/examples/parlai_chat_task_demo/webapp/package-lock.json
+++ b/examples/parlai_chat_task_demo/webapp/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.0.0",
       "dependencies": {
         "bootstrap": "^4.3.1",
-        "bootstrap-chat": "^2.0.0",
+        "bootstrap-chat": "^2.0.1",
         "rc-slider": "^8.6.3",
         "react": "16.13.1",
         "react-bootstrap": "^0.32.4",
@@ -34,7 +34,7 @@
         "eslint-plugin-jsx-a11y": "^6.2.3",
         "eslint-plugin-react": "^7.20.0",
         "eslint-plugin-react-hooks": "^4.3.0",
-        "style-loader": "^0.23.0",
+        "style-loader": "^3.3.1",
         "url-loader": "^4.1.1",
         "webpack": "^5.70.0",
         "webpack-cli": "^4.9.0"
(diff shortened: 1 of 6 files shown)


JackUrb (approved):
Thanks for finding and fixing - custom builds not using the simple build script are pretty uncommon. I thought I had checked this but perhaps not after the big package update.

PR #745, opened 2022-04-04
Adding assert to static tasks for data type

JackUrb:
# Overview
In upgrading to 1.0, we revealed some tasks that weren't following the expected format for Static Tasks. This adds an assertion to the underlying `AgentState` which is now responsible for noting when the expected data format isn't met. It also updates a comment in-code to point to the leaky abstraction where the `ClientIOHandler` contains some code that services a specific blueprint, however I don't have a short-term solution to resolve this one.

# Testing

Automated testing


pringshia (approved):
Thanks for adding. Good to have the validation closer to the data source.

Issue #746, opened 2022-04-04
Task loading endlessly and workers disconnected


claudiogreco:
Hello,

I am running a Mephisto data collection task hosted using Heroku. I am collecting annotations, but several workers (about 20% of the total number of workers) are complaining about the fact that they just saw a loading page or were disconnected at some point. In the terminal, I am seeing messages like: "Onboarding agent onboarding_2806 disconnected or errored, final status timeout.". The configuration file contains the following variables which might be relevant in order to solve the issue: allowed_concurrent: 1, max_num_concurrent_units: 0, and maximum_units_per_worker: 1. Would setting the variable max_num_concurrent_units equal to a value in the range [250-500] as suggested in the FAQs solve the issue? If yes, would it be possible to change that variable in the configuration file without having to restart the task from scratch?

Thank you very much in advance!

PR #747, opened 2022-04-04
Adding reasonable defaults to our examples for max concurrent units

JackUrb:
# Overview
Mostly just the title. As brought up in #746, not having these kinds of defaults means that if people use the default values they're more likely to hit a bad experience (too much server load) rather than be low utilization. To make the basic case better, I've selected some defaults that should be pretty good on this tradeoff (at least until we hit the point that we can use our `metrics` tooling to scale this _dynamically_... in version 2.0 or something).

### DIFF
--- a/examples/parlai_chat_task_demo/hydra_configs/conf/base.yaml
+++ b/examples/parlai_chat_task_demo/hydra_configs/conf/base.yaml
@@ -10,4 +10,4 @@ mephisto:
     num_conversations: 1
   task:
     # We expect to be able to handle 25 concurrent conversations without issue
-    max_num_concurrent_units: 50
+    max_num_concurrent_units: 50 # 25 convos * 2 people per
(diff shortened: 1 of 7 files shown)


examples/parlai_chat_task_demo/hydra_configs/conf/base.yaml
```diff
@@ -8,3 +8,6 @@ mephisto:
     world_file: ${task_dir}/demo_worlds.py
     task_description_file: ${task_dir}/task_description.html
     num_conversations: 1
+  task:
+    # We expect to be able to handle 25 concurrent conversations without issue
```
pringshia, line 12: is this a typo? Says 25 in the comment but 50 for the value...

examples/parlai_chat_task_demo/hydra_configs/conf/base.yaml
```diff
@@ -8,3 +8,6 @@ mephisto:
...
+    max_num_concurrent_units: 50
```
pringshia, line 13: 50 here^

examples/parlai_chat_task_demo/hydra_configs/conf/base.yaml
JackUrb: 25 conversations, but they require 2 connected people per. I see the comment is misleading here, can make explicit

pringshia (approved):

Issue #752, opened 2022-04-07
Mephisto metrics + docker integration misaligned


JackUrb:
# Overview
Something is happening around the metrics install and using docker for Mephisto, ending up with `OSError: [Errno 8] Exec format error: './prometheus'` as an outcome. Working theory is that the wrong version of Prometheus ends up being installed by the script within docker, though this needs to be investigated, reproduced, then resolved.

pringshia:
Did some more digging and it seems that [`$OSTYPE` is not detected by `sh`](https://stackoverflow.com/a/23425163) which is what the `mephisto metrics install` command uses, we should use `bash` instead.



```sh
$ sh mephisto/scripts/metrics/install_metrics.sh
Unable to automate install for
```

```sh
$ echo $OSTYPE
linux-gnu
```

---

With `bash`:

```bash
$ bash mephisto/scripts/metrics/install_metrics.sh

# Now it runs as expected
```

JackUrb:
The proposal here sounds good to me! For part 3 though, I'd guess that somehow the docker container is dropping the contents of the `resources/` directory due to `.dockerignore` not following the `!` exclude setup that `.gitignore` does.

pringshia:
Hmm, `!` *should* work with the `.dockerignore`...

> Lines starting with ! (exclamation mark) can be used to make exceptions to exclusions. The following is an example .dockerignore file that uses this mechanism:
>
> https://docs.docker.com/engine/reference/builder/#dockerignore-file

Issue #744, opened 2022-04-02
Cannot run example through docker Quickstart


znaeem:
I was not able to get the Mephisto project running on my machine. Particularly I cannot get the static_test_script.py to run, in the "Running your first task" guide , however the command `mephisto check` returns `Mephisto seems to be set up correctly.` I then tried to run it with docker, as did my colleague, and came across the same error, which is the following:

```
ERROR in ./dev/app.jsx 25:0-58
Module not found: Error: Can't resolve 'mephisto-task' in '/mephisto/mephisto/abstractions/blueprints/static_html_task/source/dev'
resolve 'mephisto-task' in '/mephisto/mephisto/abstractions/blueprints/static_html_task/source/dev'
  Parsed request is a module
  using description file: /mephisto/mephisto/abstractions/blueprints/static_html_task/source/package.json (relative path: ./dev)
    Field 'browser' doesn't contain a valid alias configuration
    resolve as module
      /mephisto/mephisto/abstractions/blueprints/static_html_task/source/dev/node_modules doesn't exist or is not a directory
      looking for modules in /mephisto/mephisto/abstractions/blueprints/static_html_task/source/node_modules
        single file module
          using description file: /mephisto/mephisto/abstractions/blueprints/static_html_task/source/package.json (relative path: ./node_modules/mephisto-task)
            no extension
              Field 'browser' doesn't contain a valid alias configuration
              /mephisto/mephisto/abstractions/blueprints/static_html_task/source/node_modules/mephisto-task doesn't exist
            .js
              Field 'browser' doesn't contain a valid alias configuration
              /mephisto/mephisto/abstractions/blueprints/static_html_task/source/node_modules/mephisto-task.js doesn't exist
            .json
              Field 'browser' doesn't contain a valid alias configuration
              /mephisto/mephisto/abstractions/blueprints/static_html_task/source/node_modules/mephisto-task.json doesn't exist
            .wasm
              Field 'browser' doesn't contain a valid alias configuration
              /mephisto/mephisto/abstractions/blueprints/static_html_task/source/node_modules/mephisto-task.wasm doesn't exist
        /mephisto/mephisto/abstractions/blueprints/static_html_task/source/node_modules/mephisto-task doesn't exist
      /mephisto/mephisto/abstractions/blueprints/static_html_task/node_modules doesn't exist or is not a directory
      /mephisto/mephisto/abstractions/blueprints/node_modules doesn't exist or is not a directory
[...]

PR #754, opened 2022-04-07
Remove prometheus/grafana executables from docker context

pringshia:
## Overview

Partially addresses a bug surfaced via #752 and #744 in that metrics executables saved down on the host computer would be copied into the Docker context and cause issues when running the docker image due to architecture mismatches.

This was fixed by updating the .dockerignore file

## Testing

```bash
docker build -t mephisto .
docker run -p 3000:3000 mephisto bash -c \
    'cd mephisto/examples/simple_static_task && python static_test_script.py'

...

[2022-04-07 21:04:03,109][mephisto.abstractions.providers.mock.mock_unit][INFO] - Mock task launched: localhost:3000 for preview, localhost:3000/?worker_id=x&assignment_id=4 for assignment 2
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,3 +4,11 @@
 **/build/*
 data/*
 **/*.mp4
+
+mephisto/scripts/metrics/*
+!mephisto/scripts/metrics/resources
+!mephisto/scripts/metrics/install_*.sh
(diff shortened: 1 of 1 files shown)


JackUrb (approved):
Thanks for fixing!

PR #767, opened 2022-04-22
Fixed issue in updating credentials file

### LINKED ISSUE #766: Mephisto register doesn't work properly when registering for both MTurk and MTurk Sandbox
This issue is regarding an issue with `mephisto register`. When registering for both MTurk sandbox and MTurk

Here are steps to reproduce the issue:

The original `~/.aws/credentials file` looks like this

```
[default]
aws_access_key_id=<ACCESS KEY>
aws_secret_access_key=<SECRET KEY>
```

Then I register on MTurk sandbox:
```
$ mephisto register mturk_sandbox name=my_mturk_user_sandbox access_key_id=<ACCESS KEY> secret_access_key=<SECRET KEY>
AWS credentials successfully saved in ~/.aws/credentials file
```

After this the `~/.aws/credentials file` file looks like the following:
```
[default]
aws_access_key_id=<ACCESS KEY>
aws_secret_access_key=<SECRET KEY>

[my_mturk_user]
aws_access_key_id=<ACCESS KEY>
aws_secret_access_key=<SECRET KEY>
```

Next, I register for MTurk:
```
$ mephisto register mturk name=my_mturk_user access_key_id=<ACCESS KEY> secret_access_key=<SECRET KEY>
We found an existing entry for my_mturk_user. As new credentials have been provided, we're updating the credentials, overwriting ones that already existed for the profile
Registered successfully.
```

Although the message says `Registered successfully`, the `~/.aws/credentials file` now looks like this:
```
[default]
aws_access_key_id=<ACCESS KEY>
aws_secret_access_key=<SECRET KEY>

[my_mturk_user]
aws_access_key_id=<ACCESS KEY>

aws_access_key_id=<ACCESS KEY>
```

`aws_secret_access_key` is no longer there for `my_mturk_user`, which causes the launch to fail

Python version: 3.8.9
OS Version: macOS Big Sur version 11.6.4

vaibhavad:
Fixes #766

### DIFF
--- a/mephisto/abstractions/providers/mturk/mturk_utils.py
+++ b/mephisto/abstractions/providers/mturk/mturk_utils.py
@@ -84,7 +84,9 @@ def setup_aws_credentials(
                     ] = "aws_access_key_id={}".format(register_args.access_key_id)
                     aws_credentials[
                         credentialIndex + 2
-                    ] = "aws_secret_access_key={}".format(register_args.secret_access_key)
+                    ] = "aws_secret_access_key={}".format(
+                        register_args.secret_access_key
+                    )
                     break
 
             with open(expanded_aws_file_path, "w") as aws_credentials_file:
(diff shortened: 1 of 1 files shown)


JackUrb (approved):
Thanks for fixing this re-registration bug! 

@Alex-Gurung, this resolves the bug you had reached out to me about

Issue #768, opened 2022-04-22
Mephisto CLI and ParlAI Error


ncmeade:
Hi,

Thank you for the useful code!

I've found that the Mephisto CLI works well after initial installation, however, after installing `parlai (1.6.0)` in the same environment (as required by some examples such as [parlai_chat_task_demo](https://github.com/facebookresearch/Mephisto/tree/main/examples/parlai_chat_task_demo)), the Mephisto CLI doesn't work. There looks like there might be a dependency clash between `mephisto` and `parlai`.

Steps to reproduce:

```
$ python3 -m venv venv
$ source venv/bin/activate
$ cd ~/Mephisto
$ pip3 install -e .
$ mephisto check
Mephisto seems to be set up correctly.

$ pip3 install parlai
$ mephisto check
Traceback (most recent call last):
  File "/home/nick/Downloads/ComplexQA-Mephisto/env/lib/python3.8/site-packages/pkg_resources/__init__.py", line 579, in _build_master
    ws.require(__requires__)
  File "/home/nick/Downloads/ComplexQA-Mephisto/env/lib/python3.8/site-packages/pkg_resources/__init__.py", line 897, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/home/nick/Downloads/ComplexQA-Mephisto/env/lib/python3.8/site-packages/pkg_resources/__init__.py", line 788, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (websockets 10.2 (/home/nick/Downloads/ComplexQA-Mephisto/env/lib/python3.8/site-packages), Requirement.parse('websockets==10.1'), {'mephisto'})
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/home/nick/Downloads/ComplexQA-Mephisto/env/bin/mephisto", line 33, in <module>
    sys.exit(load_entry_point('mephisto', 'console_scripts', 'mephisto')())
  File "/home/nick/Downloads/ComplexQA-Mephisto/env/bin/mephisto", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/usr/lib/python3.8/importlib/metadata.py", line 77, in load
    module = import_module(match.group('module'))
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
[...]

PR #779, opened 2022-05-23
Embracing Poetry: includes version + deps bump

### LINKED ISSUE #818: Errors when installing new dependencies
## Overview
Adding a new dependency into requirements.txt causes errors to occur in the github actions workflow.

For example, 
When I added the `rich` dependency to the requirements.txt file I got this output for the test:
https://github.com/facebookresearch/Mephisto/runs/7237323897?check_suite_focus=true

The issue is not exclusive to `rich` as I also got this error when trying to add the `detoxify` dependency.

JackUrb:
# Overview 
Preparing for the release of `1.0.2`, this PR moves us to fully adopt `poetry` as our python package manager. This is handled by moving all of the details that used to be in `setup.py`, `requirements.txt`, and `MANIFEST.in` into the `pyproject.toml` file. We're adapting this as we needed editable installations with `pip install -e .` as described in https://github.com/python-poetry/poetry/issues/34, and following https://github.com/python-poetry/poetry-core/pull/182 we could actually have it. A year late, but eventually is better than never!

After this it should be easier to modify and manage packages, and to deal with ParlAI dependencies (perhaps we can push them to adopt as well?). 

Should resolve #818 and #768.

Note - some of the `pyproject.toml` file is still commented out. This is because there's some nicer syntax coming out in poetry 1.2 that I'd like to adopt, but I can't get tests to work using it _just_ yet so I've left it commented out.


pringshia:
Seems like there are a bunch of test failures - possible that the dependencies in pyproject.toml aren't up to date?

pringshia:
Will this mean `pip install -e` won't work anymore btw? In that case we should update our Github Actions tests and Dockerfiles to use `poetry install` instead of `pip install -e`. 

https://github.com/facebookresearch/Mephisto/search?q=pip+install

JackUrb:
Debugging the tests, but `pip install -e .` still works perfectly :)

JackUrb:
Alright solved the issue by commenting out the fancy poetry 1.2 syntax, will return to that when the release is finally cut.

PR #782, opened 2022-05-27
✨ Input hides when annotation is correct for mnist example

Etesam913:
## Summary
In the mnist example clicking on the "Annotation Correct?" checkbox does not make it immediately clear to the user that they should type in the corrected annotation in the text input below. 

This is fixed by only showing the "Corrected Annotation" text input when the "Annotation Correct?" checkbox is not ticked. When it is ticked, the text input is hidden.
## Testing

https://user-images.githubusercontent.com/55665282/170790842-6508c645-6f82-4869-9178-9415be5f24aa.mov


Etesam913:
Am I not supposed to commit the package-lock.json because for some reason that seems to be adding many additions. I'm not sure why it's changing so much?

Etesam913:
I'll do `git revert`. I am using npm version 8.9.0.

pringshia (approved):

PR #785, opened 2022-06-02
Adding mephisto-worker-experience package (contains a tip component)

Etesam913:


Etesam913:
I'm thinking of setting up something like this: https://stackoverflow.com/questions/47610213/how-do-i-run-prettier-only-on-files-that-i-want-to-commit

I feel like it would be pretty helpful because on many commits I seem to forget to lint a certain file and I have to make an extra commit because of this.

Etesam913:
What are the prettier configuration settings that are supposed to be set to follow the code-style?

There might be some value in creating a .prettierrc file if there is not one already.

My prettier config is not recognizing the need for this change.
Screen Shot 2022-06-02 at 10 11 00 AM

pringshia:
> I'm thinking of setting up something like this: https://stackoverflow.com/questions/47610213/how-do-i-run-prettier-only-on-files-that-i-want-to-commit
> 
> I feel like it would be pretty helpful because on many commits I seem to forget to lint a certain file and I have to make an extra commit because of this.
> 
> Thoughts?

For note: after an in-person chat we decided to just rely on VSCode's format-on-save feature for now. Perhaps we can include this suggestion to the [front-end development README guidelines](https://github.com/facebookresearch/Mephisto/blob/main/docs/web/docs/guides/how_to_contribute/frontend_development.md).

@Etesam913 mentioned he will also look into committing a `.prettierrc` config file.

packages/mephisto-worker-experience/README.md
```diff
@@ -0,0 +1,70 @@
+# mephisto-worker-experience
+
+This package provides the Tips component to aid in the creation of user-interfaces for Mephisto
```
pringshia, line 3: "user-interfaces for Mephisto" is a bit vague. Feel free to borrow verbiage from the Project Plan or the tasks tool.

> The Tips component will allow task authors to set up a communication channel to solicit "tips" from workers to share with other workers, thus allowing for the "crowdsourcing" of the instructions for tasks as well. We find that workers sometimes will share these tips in third-party forums or via emails to the task authors. This feature creates a more vetted channel for such communication.

packages/mephisto-worker-experience/demo/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+This section has moved here: [https://facebook.github.io/create-react-app/docs/advanced-configuration](https://facebook.github.io/create-react-app/docs/advanced-configuration)
+
+### Deployment
+
+This section has moved here: [https://facebook.github.io/create-react-app/docs/deployment](https://facebook.github.io/create-react-app/docs/deployment)
+
+### `npm run build` fails to minify
+
+This section has moved here: [https://facebook.github.io/create-react-app/docs/troubleshooting#npm-run-build-fails-to-minify](https://facebook.github.io/create-react-app/docs/troubleshooting#npm-run-build-fails-to-minify)
```
pringshia, line 70: we don't need this default CRA README

packages/mephisto-worker-experience/src/index.css
```diff
@@ -0,0 +1,26 @@
+.tips-list {
+  list-style-type: none;
+  padding: 0;
+}
+.tip {
```
pringshia, line 5: we should definitely make these class names more specific.

having something as simple as `.tip` can cause unintended conflicts with other libraries that the user might be pulling in, or with css that they've written themselves.

I propose prefixing the class names here with some namespacing to avoid such collisions. For example:

`.tip` -> `.mephisto-worker-experience-tips__tip`
`.submit-button` -> `.mephisto-worker-experience-tips__submit-button`

packages/mephisto-worker-experience/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "mephisto-worker-experience",
+  "version": "1.0.0",
+  "description": "A package that provides ui components for Mephisto",
```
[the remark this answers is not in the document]
pringshia: see other comment in this PR

packages/mephisto-worker-experience/demo/README.md
pringshia, line 70: This README is not super useful.

You could just keep this part if you'd really like to:

> # Getting Started with Create React App
> 
> This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).

You should also make a note that this demo is just for inspecting the front-end portion of the tips tool, and doesn't persist the submitted data in any way.

In the future towards the completion of this milestone, we should add a new task to the `examples/` folder that is more fully featured and has the back-end support as well.

pringshia (changes requested):

packages/mephisto-worker-experience/demo/README.md
Etesam913: > (In the future towards the completion of this milestone, we'll add a new task to the examples/ folder that is more fully featured and will have the back-end support as well.)

Yeah, it doesn't make too much sense for the tips component to be implemented on the toxicity detection as I have now.

mephisto/abstractions/blueprints/remote_procedure/remote_procedure_agent_state.py
```diff
@@ -101,6 +101,8 @@ def get_data(self) -> Dict[str, Any]:
             "requests": [r.to_dict() for r in self.requests.values()],
             "start_time": self.start_time,
             "end_time": self.end_time,
+            # TODO: Add a metadata field here which handles tips that the agent submits
+            # TODO: Create a script that can review these tips by looping through all agents
```
pringshia, line 105: we can remove this TODO. we decided not to include this in `remote_procedure`...

pringshia:
Hey @Etesam913 - this is a great suggestion. I really appreciate the Pros/Cons breakdown and I think you're pretty on the money with your analysis there.

- One _compromise_ could be that we auto-call the `handleSubmitMetadata()` as default behavior (your idea), but also allow the ability to override that if the `handleSubmit` prop is explicitly provided. The general idea would look something like:
   ```jsx
   // within the Tips component:
   const { handleSubmitMetadata } = useMephistoTask();

   function submitTip = () => {
      if (props.handleSubmit) {
          props.handleSubmit(...)
          // if user is explicitly defining submit behavior, we just let them handle everything
      } else {
          handleSubmitMetadata(...)
          // otherwise we auto-submit
      }
   }
   // ...
   <button onClick={submitTip}>Submit Tip</button>
   ```

Etesam913:
> One potential concern (though may not) that comes to mind that we'd need to test out is: can we call the hook useMephistoTask from within the mephisto-worker-experience library while the parent project has it's own useMephistoTask? or even a different hook like useMephistoRemoteProcedureTask? I think this should be fine, but we'll need to test this out.

I tried this out where the webapp uses a useMephistoRemoteProcedureTask and the mephisto-worker-experience uses a useMephistoTask and there were no issues. I'll complete this in another pr as the handleSubmitMetadata() method is not available in this pr.

examples/remote_procedure/toxicity_detection/webapp/package.json
```diff
@@ -11,6 +11,7 @@
   "dependencies": {
     "bootstrap": "^4.6.0",
     "mephisto-task": "^2.0.1",
+    "mephisto-worker-experience": "file:../../../../packages/mephisto-worker-experience",
```
pringshia, line 14: Since we don't have `mephisto-worker-experience` published, this is the only way at the moment to link to the package. However, we probably want to change this to a version number (similar to how mephisto-task is in this folder). If we ship this as is, I think we should file a Github Issue mentioning to change this after the pkg is published so we don't lose track.

Another solution is to include these projects in the yarn workspace where yarn should link them up to the packages in the root project automatically.

examples/remote_procedure/toxicity_detection/webapp/src/app.jsx
```diff
@@ -62,6 +63,24 @@ function RemoteProcedureApp() {
             handleSubmit={handleSubmit}
             handleToxicityCalculation={handleToxicityCalculation}
           />
+          <Tips
+            handleSubmit={(tipData) => createTip(tipData.header, tipData.text)}
```
pringshia, line 67: isn't this essentially a no-op? createTip doesn't have any side effects right? should we `console.log` instead?

pringshia: where is `createTip` coming from? missing import?

packages/mephisto-worker-experience/README.md
```diff
@@ -0,0 +1,103 @@
...
+```
+{
+  header: header,
+  text: text,
+  type: "tips"
+}
+```
+
+### `createFeedback(text)`
+This method is meant to be used as a parameter to the `handleSubmitMetadata` function of the `mephisto-task` library. The `createFeedback(text)` takes one strings as parameters. If the parameter is not a string then a rejected promise gets returned that can be caught for error handling. If the parameter is a string then the following object gets returned
```
pringshia, line 97: grammar nit:

> takes one strings

packages/mephisto-worker-experience/src/Functions.js
```diff
@@ -0,0 +1,42 @@
...
+  if (!header || !typeof header === "string" || !header instanceof String) {
+    return new Promise(function (resolve, reject) {
+      reject("Tip header is not a string");
+    });
+  }
+  if (!text || !typeof text === "string" || !text instanceof String) {
+    return new Promise(function (resolve, reject) {
+      reject("Tip text is not a string");
+    });
+  }
```
pringshia, line 18: curious... what's the benefit of using a Promise here for validation/error handling? I'm a little unfamiliar with this pattern. Have you seen it used in other libraries?

It makes the function return signature is a little weird: `Promise | { header, text, type }`

examples/remote_procedure/toxicity_detection/webapp/package.json
Etesam913: Makes sense 👍

examples/remote_procedure/toxicity_detection/webapp/src/app.jsx
Etesam913: Yeah, it should be console.log

packages/mephisto-worker-experience/src/Functions.js
Etesam913: I thought there was some value in this because you could attach a .catch to the createTip() and do something with it for error handling. Not sure if this is too needed though because the next pr makes createTip a function that is not exported publicly.

packages/mephisto-worker-experience/README.md
```diff
@@ -0,0 +1,103 @@
...
+```
+{
+  header: header,
+  text: text,
+  type: "tips"
+}
+```
+
+### `createFeedback(text)`
+This method is meant to be used as a parameter to the `handleSubmitMetadata` function of the `mephisto-task` library. The `createFeedback(text)` takes one string as parameters. If the parameter is not a string then a rejected promise gets returned that can be caught for error handling. If the parameter is a string then the following object gets returned
```
pringshia, line 97: this will need to be updated now as well

packages/mephisto-worker-experience/README.md
```diff
@@ -0,0 +1,103 @@
...
+### `placement`
+This prop accepts a string that specifies the placement of the tips popup relative to the "Show/Hide Tips" button. 
+
+It accepts values of "top-start", "top", "top-end", "right-start", "right", "right-end", "bottom-start", "bottom", "bottom-end","left-start", "left", "left-end". 
+
+Only some of these values will be able to be applied in certain cases. For example, if the button is in the top right corner of the screen, then only the bottom and right values will be recognized as the left and top values would lead to the popup overflowing. 
+
+The default value for this prop is "top-start", but this may not always be applied as mentioned in the case above.
+### `createTip(header, text)`
+This method is meant to be used as a parameter to the `handleSubmitMetadata` function of the `mephisto-task` library. The `createTip(header, text)` takes two strings as parameters. If one or both of the parameters are not a string then a rejected promise gets returned that can be caught for error handling. If both the two parameters are strings then the following object gets returned
```
pringshia, line 87: this will need to be updated now as well

pringshia (approved):
Thanks for updating with all the feedback. Looks good to merge!

PR #791, opened 2022-06-11
[WIP] Add a tips example

Etesam913:
Supposed to add a new task that can allow for users to submit and view tips


mephisto/abstractions/architects/router/build_router.py
```diff
@@ -105,6 +108,22 @@ def build_router(
...
+
+        mephisto_data_browser = MephistoDataBrowser(db)
+        metadata = mephisto_data_browser.get_metadata_from_task_name(task_name)
+        accepted_tips = list(
+            filter(lambda tip: tip["accepted"] == True, metadata["tips"])
+        )
+
+        frontend_args["metadata"] = {
+            "tips": accepted_tips,
+            "feedback": metadata["feedback"],
```
Etesam913, line 125: Causing a problem with the python build test

mephisto/abstractions/architects/router/node/server.js
```diff
@@ -507,6 +507,24 @@ app.post("/submit_task", upload.any(), function (req, res) {
   }
 });
 
+app.post("/submit-metadata", function (req, res) {
```
Etesam913: submit-metadata should be changed to submit_metadata

mephisto/abstractions/blueprints/remote_procedure/remote_procedure_agent_state.py
```diff
@@ -92,6 +97,11 @@ def load_data(self) -> None:
             self.final_submission = state["final_submission"]
             self.start_time = state["start_time"]
             self.end_time = state["end_time"]
+            # Metadata field would not exist prior to this pr
+            if "metadata" not in state:
+                self.metadata = {"tips": [], "feedback": []}
```
Etesam913, line 102: Was running into some weird scenarios when switching branches and running tasks, in which I would get a key error with metadata when loading data. That's why I added this if block

Etesam913: Yeah, I just removed the if statement and I don't see any key errors. I'll just get rid of the check.

examples/remote_procedure/toxicity_detection/webapp/package.json
```diff
@@ -10,13 +10,13 @@
   "author": "",
   "dependencies": {
     "bootstrap": "^4.6.0",
-    "mephisto-task": "^2.0.1",
+    "mephisto-task": "file:../../../../packages/mephisto-task",
```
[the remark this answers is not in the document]
pringshia: if you're having trouble with yarn workspaces, we can look at [lerna](https://lerna.js.org/) as well

examples/static_react_task_with_tips/webapp/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "main": "webpack.config.js",
+  "scripts": {
+    "dev": "webpack --mode development",
+    "dev:watch": "webpack --mode development --watch"
+  },
+  "keywords": [],
+  "author": "",
+  "dependencies": {
+    "mephisto-task": "file:../../../packages/mephisto-task",
+    "mephisto-worker-experience": "file:../../../packages/mephisto-worker-experience",
```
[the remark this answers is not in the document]
Etesam913: #795 removes the need for these file:../../ statements. You would be able to link using an external script.

Using yarn workspaces may be a little complex due to the fact that the python run_task.py runs npm commands like npm install and npm run dev. These would have to be changed to their yarn equivalents.

mephisto/abstractions/_subcomponents/agent_state.py
```diff
@@ -127,7 +127,7 @@ def set_init_state(self, data: Any) -> bool:
         raise NotImplementedError()
 
     @abstractmethod
-    def get_init_state(self) -> Optional[Any]:
+    def get_init_state(self, get_all_state: Optional[bool] = False) -> Optional[Any]:
```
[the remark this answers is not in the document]
Etesam913: This is for the static_agent_state.py 's implementation of get_init_state(). For some reason the default version of this method only returned state["inputs"], which is not enough as state["metadata"] is also needed. Setting this property to true allows for the whole state object to be returned.

examples/remote_procedure/toxicity_detection/webapp/package.json
```diff
@@ -10,13 +10,13 @@
...
+    "pm": "^2.2.6",
```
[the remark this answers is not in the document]
Etesam913: Yeah, that's not supposed to be there 😅 

I probably mispelled npm or something

examples/remote_procedure/toxicity_detection/webapp/package.json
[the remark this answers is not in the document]
Etesam913: Yeah, I reverted this as toxicity detection does not even need the local version of mephisto-task anymore. 

I wrote a little bit about yarn workspaces here: https://github.com/facebookresearch/Mephisto/pull/791/files/6a8f43ec38efb01077fb9bbac304c1db42f5ca59#r904149774

packages/mephisto-task/src/index.js
```diff
@@ -79,7 +97,10 @@ const useMephistoTask = function () {
     if (taskConfig.block_mobile && isMobile()) {
       setState({ blockedReason: "no_mobile" });
     } else if (!state.isPreview) {
-      requestAgent().then((data) => afterAgentRegistration(data));
+      requestAgent().then((data) => {
+        console.log(data);
```
s0mya, line 101: We should remove console.log. Is there a logger that can be used instead?

mephisto/abstractions/blueprints/static_react_task/static_react_blueprint.py
```diff
@@ -125,3 +130,26 @@ def assert_task_args(
...
+        db = LocalMephistoDB()
+        mephisto_data_browser = MephistoDataBrowser(db)
+        metadata = mephisto_data_browser.get_metadata_from_task_name(task_name)
+        accepted_tips = list(
+            filter(lambda tip: tip["accepted"] == True, metadata["tips"])
+        )
+        frontend_task_config.update({"metadata": {"tips": accepted_tips}})
+        # Use overrides provided downstream
+        frontend_task_config.update(self.frontend_task_config)
+        return frontend_task_config
```
s0mya, line 155: Is this the same function as in mephisto/abstractions/blueprints/remote_procedure/remote_procedure_blueprint.py. It's hard tell if they are same or different. If same, does it make sense to refactor into common code.

s0mya (approved):
Looks good to me, thanks for taking this on end-to-end!

I had some enhancement suggestions for your consideration. 

1. Rather than a script, would some kind of a web admin interface make sense for tip approval ?
2. May be a good idea to think about splitting up the submit and view tips actions. With more that a few tips I suspect the ui will become cluttered. 
3. it might be worth reducing the font sizes on the tips also.

packages/mephisto-task/src/index.js
Etesam913: Yeah, I probably was just checking something and forgot to remove it.

mephisto/abstractions/blueprints/static_react_task/static_react_blueprint.py
Etesam913: Yeah, I think I thought that this method was implementing an abstract version, but this doesn't actually seem to be the case. I should be able to make it into common code.

Etesam913:
> Looks good to me, thanks for taking this on end-to-end!
> 
> I had some enhancement suggestions for your consideration.
> 
>     1. Rather than a script, would some kind of a web admin interface make sense for tip approval ?
> 
>     2. May be a good idea to think about splitting up the submit and view tips actions. With more that a few tips I suspect the ui will become cluttered.
> 
>     3. it might be worth reducing the font sizes on the tips also.

For the second enhancement suggestion I did the following:

https://user-images.githubusercontent.com/55665282/175383967-b49bbc94-2426-43e4-9545-90be12772f83.mov

The maxHeight of the tips list is the maxHeight of the popup/2, so it is dynamic.

packages/mephisto-task/README.md
```diff
@@ -102,6 +102,26 @@ Generally speaking, the value is the `InitializationData` object that the `TaskR
 
 A callback provided for the webapp to finalize and submit the worker's resulting work back to Mephisto.
 
+### `handleMetadataSubmit(payload)`
+A callback provided for the Tips component in the `mephisto-worker-experience` package to submit a tip to its agent's metadata. 
+Not recommended to be used by the user as mephisto-worker-experience should use this for you under the hood.
```
pringshia, line 107: The `handleMetadataSubmit` callback is generic and can be used for any future addon, not just Tips, Feedback or `mephisto-worker-addons` library. Though you developed it in conjunction with these, we should keep the documentation here vague to indicate to users at a more generic level what this method does: persists agent-level metadata in the AgentState.

packages/mephisto-task/README.md
```diff
@@ -102,6 +102,26 @@ Generally speaking, the value is the `InitializationData` object that the `TaskR
...
+    type: "tips"
+}
+```
+or
+```js
+{
+    text: string,
+    type: "feedback"
+}
+```
```
pringshia, line 123: ^this feedback and tips specific payload information should be in the README for `mephisto-worker-addons`, not here

packages/mephisto-task/src/index.js
```diff
@@ -67,6 +67,24 @@ const useMephistoTask = function () {
     },
     [state.agentId]
   );
+  const handleMetadataSubmit = React.useCallback(
+    (...args) => {
+      let malformed = false;
```
pringshia, line 72: Unused variable?

packages/mephisto-worker-experience/src/Tips.js
```diff
@@ -0,0 +1,188 @@
...
+  const tipsArr = (list ? list : []).concat(
+    taskConfig ? taskConfig["metadata"]["tips"] : []
+  );
+  const headlessPrefix = headless ? "headless-" : "";
+  const [tipData, setTipData] = useState({ header: "", text: "" });
+  const tipsComponents = tipsArr.map((tip, index) => {
+    return (
+      <li
+        key={`tip-${index + 1}`}
+        className={`${headlessPrefix}mephisto-worker-experience-tips__tip`}
```
pringshia, line 54: I would suggest you consider pulling the `mephisto-worker-experience` portion also into a prefix var,maybe just call it `stylePrefix` or something, so that:

```jsx
const headlessPrefix = headless ? "headless-" : "";
const stylePrefix = `${headlessPrefix}mephisto-worker-experience-tips__`

<div className={`${stylePrefix}tip`} />

// or:

const s = className => `${headlessPrefix}mephisto-worker-experience-tips__${className}`

<div className={s('tip')} />
```

makes the classnames in the html more readable, less verbose/repetitive

A less important reason is that that now that we'll have to rename `mephisto-worker-experience` to `mephisto-worker-addons` it's easier to make changes like that. Although this is not as important bc we won't be changing this name again, and with find/replace it's pretty easy to fix.

packages/mephisto-worker-experience/src/index.css
```diff
@@ -0,0 +1,207 @@
+/* Headless Classes */
+.headless-mephisto-worker-experience-tips__tips-list {
```
pringshia, line 2: unfortunately we'll have to rename these styles to `mephisto-worker-addons`. apologies for the thrash here, should be a simple find/replace

mephisto/abstractions/blueprint.py
```diff
@@ -264,6 +264,25 @@ def get_frontend_args(self) -> Dict[str, Any]:
...
+        frontend_task_config: Dict[str, Any],
+    ):
+        """
+        Updates the frontend args metadata with accepted tips
+        """
+        metadata = mephisto_data_browser.get_metadata_from_task_name(task_name)
+        accepted_tips = list(
+            filter(lambda tip: tip["accepted"] == True, metadata["tips"])
+        )
+        metadata["tips"] = accepted_tips
```
pringshia, line 280: As mentioned in other comments through out, I'm not sure how I feel about leaking the tips information logic into the base blueprint class. we should be thinking in vaguer terms on the level of "metadata" or "plugins/addons" here, not specific implementations. this is a code smell

pringshia: one solution could be to move the lambda to be an argument to the `update_task_config_with_metadata` function. that way the caller specifies the metadata-specific logic, and doesn't concern the blueprint class with the knowledge. lmk if this doesn't make sense, happy to clarify

examples/remote_procedure/toxicity_detection/webapp/src/app.jsx
```diff
@@ -54,14 +53,21 @@ function RemoteProcedureApp() {
...
             handleSubmit={handleSubmit}
             handleToxicityCalculation={handleToxicityCalculation}
           />
+          <div
+            style={{
+              marginTop: "20rem",
+              marginLeft: "auto",
+              width: "fit-content",
+            }}
+          ></div>
```
pringshia, line 70: Is this just a spacer div? Why do we need it?

examples/static_react_task_with_tips/README.md
```diff
@@ -0,0 +1,131 @@
+# Simple Static React Task With Tips
```
pringshia, line 1: Having a copy-pasted duplicated README here creates "two sources of truth" that we have to maintain on an on-going basis. I would much prefer linking to the other README from this README, and only keeping the additional "delta" information on Tips usage here.

mephisto/abstractions/_subcomponents/agent_state.py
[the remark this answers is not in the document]
pringshia: got it, thanks for the explanation, I understand the need here. I'd like to get @JackUrb 's thoughts on this as well.

The overloaded fn signature feels a little awkward to me, I feel as if we should have a separate abstract method instead so both `get_init_state` and `get_full_state` or something. We would then have to implement that abstract method though in all the derived classes. Would also have to understand if this approach would make the upgrade story complicated for users.

package.json
```diff
@@ -17,6 +17,7 @@
   "workspaces": [
     "packages/annotated/*",
     "packages/mephisto-task",
+    "packages/mephisto-worker-experience",
```
pringshia, line 20: we'll have to rename to `mephisto-worker-addons`. happy to accept that either in this PR or a quick follow up

examples/remote_procedure/toxicity_detection/webapp/src/app.jsx
Etesam913: Removed

examples/static_react_task_with_tips/README.md
Etesam913: Agreed

packages/mephisto-task/src/index.js
Etesam913: Removed it

mephisto/operations/operator.py
```diff
@@ -9,7 +9,6 @@
 import os
 import time
 import threading
-import traceback
```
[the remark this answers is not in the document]
Etesam913: It wasn't being used globally. The file did use traceback in exceptions, but it would import it in those exceptions.

I added it back and removed the other imports of traceback in this file

mephisto/abstractions/blueprint.py
Etesam913: It would still concern a blueprint class. The update_task_config_with_metadata method(allows for metadata to be exposed in the frontend through taskConfig) is called in the get_frontend_args() method of the static_react_blueprint.py, remote_procedure_blueprint.py, and the parlai_chat_blueprint.py files. 

I guess it is a good idea though, because the base blueprint file will not be using the databrowser

mephisto/operations/operator.py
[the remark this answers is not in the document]
pringshia: ah gotcha

package.json
Etesam913: Yeah, I did it in this pr

Rebecca-Qian (approved):
@Etesam913 this is awesome work! Really looking forward to using this Mephisto feature.

Some UX nits: Can you make the font size for tips a bit smaller, and/or use grey text? In the beginning when there are no tips, you can also make the text a bit smaller and/or lighter color so it's less jarring.

Also, for future PRs I'd try to keep them smaller and more modular if possible :) it's easier to review that way. For large PRs like this it would also be helpful to indicate which files need more review in the PR description, or include a brief description of the files committed.

Etesam913:
Yeah, I definitely agree with that

> @Etesam913 this is awesome work! Really looking forward to using this Mephisto feature.
> 
> Some UX nits: Can you make the font size for tips a bit smaller, and/or use grey text? In the beginning when there are no tips, you can also make the text a bit smaller and/or lighter color so it's less jarring.
> 
> Also, for future PRs I'd try to keep them smaller and more modular if possible :) it's easier to review that way. For large PRs like this it would also be helpful to indicate which files need more review in the PR description, or include a brief description of the files committed.

Here is the updated version with some of your suggestions:

Screen Shot 2022-06-30 at 12 23 38 PM

examples/remote_procedure/mnist/run_task.py
```diff
@@ -65,7 +65,7 @@ def handle_with_model(
     )
 
     task_dir = cfg.task_dir
-    build_custom_bundle(task_dir)
+    build_custom_bundle(task_dir, cfg.mephisto)
```
pringshia, line 68: We decided to turn this into an overloaded method with default values, so as to avoid introducing breaking changes. I'm happy to use Python's kwargs here for the remaining two params, `post_build_script` (default `""`) and `force_rebuild` (default: `False`)

**Example usage:**
```
build_custom_bundle(task_dir) # still workds
build_custom_bundle(task_dir, force_rebuild=True)
build_custom_bundle(task_dir, force_rebuild=cfg.mephisto.task.force_rebuild) # from config
```

@JackUrb @Etesam913

mephisto/abstractions/_subcomponents/agent_state.py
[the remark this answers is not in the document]
JackUrb: Oof this is a bit hairy and points to some of my pending refactors here. `StaticAgentState` for instance should be returning the whole state object as well, and, in general, `AgentState` should have some top-level attributes that are shared and known.

I'd consider this low-hanging fruit for a code quality refactor (in 2 stages). The base `AgentState` class should define the attributes we expect to be in _every_ task, and then in the subclasses that have inherited and perhaps used different names we can alias to the correct values with the `@property` decorator. Then we can mark those as deprecated and remove them in 1.2 or later.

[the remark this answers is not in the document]
JackUrb: I don't want to block you on this for too long @Etesam913 - won't get a chance today but I can throw together the `AgentState` refactor tomorrow morning, then this would just need a rebase.

mephisto/abstractions/blueprint.py
JackUrb: This method is likely to get refactored as we'll be pulling tips from a directory instead.

mephisto/tools/scripts.py
```diff
@@ -234,8 +237,21 @@ def build_custom_bundle(custom_src_dir, run_config: DictConfig):
...
+        warnings.warn(
+            "Consider adding the force_rebuild property to your hydra config and passing it into build_custom_bundle()\nSteps on how to do this: \nhttps://github.com/facebookresearch/Mephisto/issues/811",
+            stacklevel=2,
+        )
+
+    if "post_install_script" not in kwargs:
+        warnings.warn(
+            "Consider adding the post_install_script property to your hydra config and passing it into build_custom_bundle()\nSteps on how to do this: \nhttps://github.com/facebookresearch/Mephisto/issues/811",
+            stacklevel=2,
+        )
```
pringshia, line 251: I think this is too noisy, these flags would only be really used with very limited use cases (developers happening to be working on a library package) yet we would be promoting this warning to all users of mephisto on ever run. I know we discussed adding this during the meeting, but I suggest we just remove it altogether.

JackUrb: Agreed on retrospect. Would be worth linking the issue in a comment at the top of the script though.

mephisto/abstractions/blueprint.py
pringshia: although the final UI data will be from the directory, I think we still need an approach for the helper scripts to access the metadata when the reviewer has to review the tips - so wouldn't we still need to figure this out?

mephisto/tools/scripts.py
Etesam913: I agree.
I just removed the warnings.
I did add a comment to #811  in case the programmer is looking at the code and is interested in local package development

examples/static_react_task_with_tips/webapp/src/css/style.css
```diff
@@ -0,0 +1,51 @@
...
+  margin-right: 1rem;
+  font-size: 1.7em;
+  cursor: pointer;
+}
+.good {
+  background-color: #48c774;
+}
+.bad {
+  background-color: #f14668;
+}
```
Etesam913, line 51: I wrote custom styles for this task as the inherit styles from bulma & bootstrap were causing the tips popup to look weird.

packages/mephisto-worker-addons/webpack.config.js
```diff
@@ -0,0 +1,57 @@
...
+  output: {
+    path: __dirname,
+    filename: "lib/index.js",
+    library: "mephisto-worker-addons",
+    libraryTarget: "umd",
+  },
+  target: "web",
+  externals: {
+    react: "react",
+  },
```
Etesam913, line 21: For some reason adding `react-dom` to this externals object will cause the task that is linking to this package to receive a cannot find `react-dom` error. 

May be worthwhile to investigate this at a different time

Etesam913 (commented):
Did a cleanup of the files.

* Added tips field to the _AgentStateMetadata
* Added tips field to the return dict of the data browser's get_data_from_unit() method:
`"tips": agent.state.get_tips(),`
* Removed many unused imports

mephisto/abstractions/_subcomponents/agent_state.py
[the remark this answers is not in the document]
Etesam913: It should be fully merged with your refactor now

mephisto/abstractions/_subcomponents/agent_state.py
```diff
@@ -268,3 +268,20 @@ def get_task_end(self) -> Optional[float]:
...
+    def update_metadata(self, property_name: str, property_value: Any) -> None:
+        if self.metadata is not None:
+            assert (
+                hasattr(self.metadata, property_name) == True
+            ), "The {property_name} field must exist in _AgentStateMetadata. Go into _AgentStateMetadata and add the {property_name} field".format(
+                property_name=property_name
+            )
+            replaced = {property_name: property_value}
+            self.metadata = replace(self.metadata, **replaced)
+            self.save_data()
```
Etesam913, line 287: This is how you replace a property of a dataclass at runtime

mephisto/scripts/local_db/remove_accepted_tip.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+    print("")
+    while task_name not in task_names:
+        print("That task name is not valid\n")
+        task_name = input(
+            "Enter the name of the task that you want to look at the tips of: \n"
+        )
+        print("")
+    units = mephisto_data_browser.get_all_units_for_task_name(task_name)
+    if len(units) == 0:
```
Etesam913, line 26: This command line interface of this scripts gets improved in #813

Etesam913 (commented):
Added simple comments

examples/static_react_task_with_tips/webapp/src/css/style.css
pringshia: this is a larger issue though, we must not rely on or mandate users to have to update their styles to avoid conflicts. Ideally their is scoping and separation. Perhaps we should use the Shadow DOM? https://web.dev/shadowdom-v1/ See the scoped CSS section.

Etesam913: So I got the shadow dom to work, but I am not sure that this is fully desired. I believe the point of the headless Tips component is to remove most of the styles and to allow the user to modify the styles so that they can style it themselves. This isn’t possible with the shadow dom as you can’t apply styles

pringshia: how about we toggle it based on the headless prop? if they specify headless, it doesn't use ShadowDOM, if they don't then it uses ShadowDOM?

Etesam913: Yeah, I'll definitely do that, but it doesn't change the underlying reason why I added these styles. The style conflicts with bulma were making the headless version of the component look really ugly. I removed bulma and added these styles to make the task look as similar to the static_react_task example

Etesam913: This is how the headless component looks when the shadow dom is off:
Screen Shot 2022-07-12 at 10 16 29 AM

The styles from bulma removes bold from headers & keeps their size the same as regular text. I added the above styling and removed bulma so that the rest of the task looked the same as the static_react_task and so that the headless version of tips looked normal.

JackUrb (commented):
Thanks for your patience on this one - I think there's one last round of changes to consolidate and clean up here, but afterwards it should be good to ship!

mephisto/abstractions/blueprints/parlai_chat/parlai_chat_blueprint.py
```diff
@@ -306,8 +307,19 @@ def get_frontend_args(self) -> Dict[str, Any]:
...
+
+        task_name = self.task_run.to_dict()["task_name"]
+        db = LocalMephistoDB()
+        mephisto_data_browser = MephistoDataBrowser(db)
+        tips = mephisto_data_browser.get_metadata_property_from_task_name(
+            task_name=task_name, property_name="tips"
+        )
+        front_end_task_config_with_metadata = super().update_task_config_with_tips(
+            tips, frontend_task_config
+        )
```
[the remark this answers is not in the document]
Etesam913: I think I did do this in the past, but reverted it because there was some concerns about using the DataBrowser and LocalMephistoDB() in the base blueprint class.

mephisto/abstractions/_subcomponents/agent_state.py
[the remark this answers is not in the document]
Etesam913: No it is not, just removed it

mephisto/tools/data_browser.py
```diff
@@ -48,6 +48,26 @@ def _get_units_for_task_runs(self, task_runs: List[TaskRun]) -> List[Unit]:
...
+                for unit in found_units:
+                    if unit.get_status() in [
+                        AssignmentState.COMPLETED,
+                        AssignmentState.ACCEPTED,
+                        AssignmentState.REJECTED,
+                        AssignmentState.SOFT_REJECTED,
+                        AssignmentState.EXPIRED,
+                    ]:
+                        units.append(unit)
+        return units
```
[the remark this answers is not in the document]
Etesam913: Makes sense, done

mephisto/tools/data_browser.py
```diff
@@ -28,45 +25,47 @@ def __init__(self, db=None):
...
+        Return a list of all Units in a terminal completed state from all
+        the provided TaskRuns.
+        """
+        statuses = [
+            AssignmentState.COMPLETED,
+            AssignmentState.ACCEPTED,
+            AssignmentState.REJECTED,
+            AssignmentState.SOFT_REJECTED,
+        ]
+        return self.collect_matching_units_from_task_runs(task_runs, statuses)
```
pringshia, line 55: i think passing in the negation (filtering) would feel better here, e.g. all tasks except those that are EXPIRED, but I'm just nit-picking, this is fine

Etesam913: I feel like that would just make the code harder to understand as you would have to go to assignment_state.py to figure out which AssignmentState's are being accepted. 

It wouldn't really make it more concise either as the code would become

```python
filtered_statuses=[
  AssignmentState.CREATED,
  AssignmentState.LAUNCHED,
  AssignmentState.ASSIGNED,
  AssignmentState.MIXED,
  AssignmentState.EXPIRED  
]
```

JackUrb: Generally for legibility, it could also be OK to put these collections inside of the `AssignmentState` class as class functions. `AssignmentState.completed()` and `AssignmentState.final_agent()` come to mind as relevant for these two lists at the moment.

mephisto/abstractions/blueprints/parlai_chat/parlai_chat_blueprint.py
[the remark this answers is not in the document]
JackUrb: Ah okay I recall now what I had considered - Ideally we'd be pulling the `tips` from the task folder on launch, which will be part of the hydra `DictConfig` default arguments, rather than from a query by task name. To this end it may be necessary to implement the "storing tips in an accessible file" here. 

The implementation here would instead look something like:
```
tips = self._load_tips_from_dir(self.args.blueprint.tips_directory)
```
and that would be safe in the `super`.

For the tips saving method, I imagine something like:
```
save_dir = unit.get_task_run().args.blueprint.tips_directory
# write tips out to file
```

And then the default value for `tips_directory` should be interpolatable from `${task_dir}/tips`.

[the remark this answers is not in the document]
Etesam913: I am planning on keeping the tips file in the task's folder.

For example, the static_react_task_with_tips task should have its tips in the `examples/static_react_task_with_tips` folder. The run_task.py file seems to have a property called cfg.task_dir that gets the appropriate field. 

Is there anyway to get this cfg.task_dir in the blueprint.py file?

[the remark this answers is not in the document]
JackUrb: `cfg.task_dir` is a configuration value that is only available from a `TaskConfig` used from a run script. Adding the `tips_directory` argument (or similar) to the blueprint and then autofilling with `${task_dir}/...` locks in that value at runtime and would ensure that future loads of the blueprint have the correct directory filled.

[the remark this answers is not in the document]
Etesam913: What do you mean by autofilling? Not exactly sure what that means.

mephisto/abstractions/blueprint.py
```diff
@@ -262,8 +264,47 @@ def get_frontend_args(self) -> Dict[str, Any]:
...
         """
+        self.update_task_config_with_tips(self.frontend_task_config)
         return self.frontend_task_config.copy()
 
+    def get_tips_directory(self) -> Optional[str]:
+        """
+        Returns the tips directory as set in hydra config
+        By default this is '${task_dir}/outputs/tips'
+        """
+        path_to_tips = self.args["blueprint"]["tips_location"]
```
JackUrb: `self.args.blueprint.tips_location` is the better path to access here (treating as a `DictConfig`) rather than a `Dict`

JackUrb (changes requested):
I know this has gotten a lot of rounds of feedback, but I think this should be the last one! Thanks for your patience here as we've been working through a new design.

mephisto/abstractions/blueprints/parlai_chat/parlai_chat_blueprint.py
```diff
@@ -127,6 +128,13 @@ class ParlAIChatBlueprintArgs(OnboardingRequiredArgs, BlueprintArgs):
             "help": "Optional count of conversations to have if no context provided"
         },
     )
+    tips_location: str = field(
+        default=MISSING,
```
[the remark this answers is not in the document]
Etesam913: That's what I originally kept it as, but I was getting the test error so I thought that setting it to MISSING would solve the error. It did not, I had to do something else. I will revert it back

mephisto/abstractions/blueprints/parlai_chat/parlai_chat_blueprint.py
```diff
@@ -149,7 +157,7 @@ def __init__(
     ):
         super().__init__(task_run, args, shared_state)
         self._initialization_data_dicts: List[Dict[str, Any]] = []
-
+        self.task_run = task_run
```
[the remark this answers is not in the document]
Etesam913: yeah

mephisto/abstractions/blueprints/parlai_chat/parlai_chat_blueprint.py
```diff
@@ -306,8 +314,7 @@ def get_frontend_args(self) -> Dict[str, Any]:
                 "frontend_task_opts": shared_state.frontend_task_opts,
             }
         )
-        # Use overrides provided downstream
-        frontend_task_config.update(self.frontend_task_config)
```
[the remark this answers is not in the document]
Etesam913: I don't really know what it does. It looks completely redundant, but I'll add it back though.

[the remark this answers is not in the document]
JackUrb: `self.frontend_task_config` may not actually be captured by `super().get_frontend_task_config()`, and importantly `self.frontend_task_config` is something we can set values for directly from the command line. This means, if the user wants to, by doing this at the end we can always ensure they have a path to force override something set in here.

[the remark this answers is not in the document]
pringshia: agree that the code looks a bit confusing. ideally if there's a way to capture the above explanation in source (whether via comments, more explicit variable names etc., or other) that would be good

examples/static_react_task_with_tips/hydra_configs/conf/example.yaml
```diff
@@ -7,6 +7,7 @@ mephisto:
   blueprint:
     task_source: ${task_dir}/webapp/build/bundle.js
     link_task_source: true
+    tips_location: ${task_dir}/outputs/tips.csv
```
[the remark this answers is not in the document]
Etesam913: Yeah, I am adding it to the BlueprintArgs class instead to avoid the need to add the tips_location property itself to each blueprint.

This is causing a github actions testing error that I have to solve though.

mephisto/abstractions/blueprint.py
```diff
@@ -229,7 +231,7 @@ def __init__(
         self.args = args
         self.shared_state = shared_state
         self.frontend_task_config = shared_state.task_config
-
+        self.task_run = task_run
```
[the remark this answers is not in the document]
JackUrb: Nit: this is still here, but I don't see why it's added. Please remove.

Issue #787, opened 2022-06-06
`build_custom_bundle` can cause dev confusion for package authors


pringshia:
## Overview

Let's say you are working on changes in the `mephisto-task` package.

You'd like to test out these changes in a project from the `examples/` folder. You add a `console.log` line to `mephisto-task`, but despite re-running `python run_task.py` you don't see the changes reflected in the browser.

The issue is that the [`build_custom_bundle`](https://github.com/facebookresearch/Mephisto/blob/afed2955ba54f4dea80c27cea3e734f3528bbaa5/mephisto/tools/scripts.py#L241-L261) method includes a build-time optimization for tasks so that they are only rebuilt when changes are detected in the immediate source code for the webapp, not in a linked library package (or anything else in node_modules for that matter).

## Current workaround (not ideal)

Every time you make a change in a linked library file, e.g. `mephisto-task`, make a modification to the webapp task source as well. Could be a dummy line like `console.log(1);` that you then change to 2, 3, 4, etc. The app will rebuild every time the file changes, so we are manually forcing a change for development purposes.

## Proposed workarounds

- Give `build_custom_bundle` another param that allows rebuilds to always occur, bypassing the optimization.
- Allow node_modules to also be part of the optimization check (could be very slow though! and we'd need to make sure we follow symlinks)
- Create a "development mode" local setting for the repo that automatically links most things live and avoids a lot of these foot guns. Right now there are [too many steps and gotchas](https://mephisto.ai/docs/guides/how_to_contribute/frontend_development/) to set up the local development experience.

## References

https://github.com/facebookresearch/Mephisto/blob/afed2955ba54f4dea80c27cea3e734f3528bbaa5/mephisto/tools/scripts.py#L241-L261

The optimization is that the build is only run if no other source files were modified after the build artifact. On the surface this seems sensible. However, when sometimes developing on a Mephisto package locally, e.g. `mephisto-task`, this can lead to confusion because the library update is under `node_modules` which the `build_custom_build` help ignores.

Issue #788, opened 2022-06-07
Include additional instructions to docs for local front-end development


pringshia:
Update https://github.com/facebookresearch/Mephisto/blob/main/docs/web/docs/guides/how_to_contribute/frontend_development.md

Potential items to include:
- Link example projects to locally library packages by editing the `webapp/package.json` file as such:
```js
// package.json
"mephisto-task": "file:../../../../packages/mephisto-task"
```

- Make a note of the issue and workaround in #787 
- Troubleshooting suggestions for yarn workspace management, e.g. clearing the yarn cache, etc.

Etesam913:
This gets resolved once #795 gets merged in.
* You should not be using the file syntax to link a package once #795 is merged in. 
* To link you would
  * create a .sh script in your webapp folder that runs `npm link mephisto-task`
  * set force_rebuild: true and post_install_script: <name of .sh script> in the task property of your hydra config file

PR #795, opened 2022-06-16
Add tests for static_react_task example

Etesam913:
### Tests
Configured cypress with the static_react_task webapp. 
Added tests that test if:
* a request is made to the "/request_agent" endpoint
* the correct react elements are displayed
* a request is made to the "/submit_task" endpoint after clicking the green button
* a request is made to the "/submit_task" endpoint after clicking the red button
* an alert shows up with the correct text after clicking the green button
* an alert shows up with the correct text after clicking the red button

### GitHub Actions
Added a github action workflow named, cypress-e2e-tests which runs the cypress tests as described above.

### Linking (RESOLVED)
To make sure that tests on a task adapt when doing local package development, the example task's webapp should be linked to the package.  

This can easily be done by writing something like this in the package.json :

`"mephisto-task": "file: ../../../packages/mephisto-task",`

However, this is not ideal as an individual would have to go to mephisto-task run `npm link`, and then go to the examples task's webapp folder and run `npm link mephisto-task`, and then cd out and type python run_task. This is clearly many extra steps.

The best of both worlds would allow someone who is just running the task without doing any local package development work to avoid having to do all the npm linking. This intuitively should work by the following steps:

1. Setting the following in the package.json: `"mephisto-task": "^2.0.1",`

2. running `npm link mephisto-task` in the webapp folder
3. Running python run_task.py in the root folder for the task.
  
However, this is not possible as running python run_task.py runs `npm install` in the webapp folder (done under the hood), which undoes the link and installs mephisto-task 2.01 from the package.json. 

To resolve this issue for local development a new command line arg is defined.

#### Example:
Step 1:
``` bash
cd packages/mephisto-task
npm link
```
Step 2:
```bash
cd examples/static_react_task
python run_task.py mephisto.task.package_to_link_to=mephisto-task
```

This works as this command line arg is read by the program and runs npm link mephisto-task after the npm install to get the local version of the package.
 
Removing the **mephisto.task.package_to_link_to** command line arg skips the npm link command. This means that the version of the package as specified in the package.json is used.

This achieves the best of both worlds, easy running of a task when doing no local development, and local development without having to tinker with the package.json by adding a weird file path.

#### Downsides
https://github.com/facebookresearch/Mephisto/pull/795/files/ac59bbdcde1c861f892d32485aaee87b2c7db8bc#r901965814

## Testing

https://user-images.githubusercontent.com/55665282/174145018-11851bdf-52cb-419d-b1d0-a4e265a7dd0e.mov

To see the [GitHub action output](https://github.com/facebookresearch/Mephisto/runs/6941057173?check_suite_focus=true) you can look at the "cypress-e2e-tests" test and click on "details" in the bottom of this pr.

## TODO:
- [x] Planning on adding the information to run these tests in the CONTRIBUTING.md file
- [x] Configure this with GitHub actions
- [x] Look at more Cypress file/test structures to understand if my file structure makes sense

This pr will fix the issues: #787 #788


mephisto/tools/scripts.py
```diff
@@ -239,6 +239,7 @@ def build_custom_bundle(custom_src_dir):
     build_path = os.path.join(prebuild_path, "build", "bundle.js")
 
     # see if we need to rebuild
```
Etesam913, line 241: The code below is an optimization that is commented out. 

**I probably need some feedback if this is the correct way to go.** 

What this code does is that it forgoes the `npm build` and `npm link` commands when it detects that nothing about the webapp changes. However, switching from 

```bash
python run_task.py
```

to 

```bash
python run_task.py mephisto.task.package_to_link_to=mephisto-task
```
requires the npm link command to gets executed. As nothing changed about the webapp the optimization is ran and the npm link command is skipped.

The optimization is pretty detrimental to local package development as you are forced to do quirky things like adding random text to the webapp to force the build and link. This is really unintuitive and personally has made me waste a lot of time in trying to fix.

The optimization itself only really seems to save 1-2 seconds for each time python run_task.py is ran.

Outside of commenting it out, another solution could be to create a new command line arg named, "always_build", which would skip the optimization when set to true. 

Thoughts 🤔?

Etesam913, line 241: The code below is an optimization that is commented out. 

**I probably need some feedback if this is the correct way to go.** 

What this code does is that it forgoes the `npm build` and `npm link` commands when it detects that nothing about the webapp changes. However, switching from 

```bash
python run_task.py
```

to 

```bash
python run_task.py mephisto.task.package_to_link_to=mephisto-task
```
requires the npm link command to gets executed. As nothing changed about the webapp the optimization is ran and the npm link command is skipped.

The optimization is pretty detrimental to local package development as you are forced to do quirky things like adding random text to the webapp to force the build and link. This is really unintuitive and personally has made me waste a lot of time in trying to fix.

The optimization itself only really seems to save 1-2 seconds for each time python run_task.py is ran.

Outside of commenting it out, another solution could be to create a new command line arg named, "always_build", which would skip the optimization when set to true. 

Thoughts 🤔?

pringshia: left a comment below talking about a suggestion for adding code to avoid the optimization. see ~L260

pringshia:
Thanks for describing the issue in the Linking section above. I think I understand the problem here, the Mephisto builder does indeed call `npm install` and I can see how that would blow out the `npm link` set up for local development.

I think that adding a config such as `mephisto.task.package_to_link_to` could work here, though I wonder if we could do something even more generic that could be reusable for more cases in the future and enable flexibility. The current solution seems extremely overfit to the current problem we are facing.

My alternative suggestion here would be something like **`mephisto.task.post_build_script`** instead.

---

The usage would then be something like this:

```sh
python run_task.py mephisto.task.post_build_script=./link_mephisto_task.sh
```

and then you have a file `link_mephisto_task.sh`:

```sh
npm link mephisto-task
```

This way you could run [*any* arbitrary code](https://stackoverflow.com/questions/3777301/how-to-call-a-shell-script-from-python-code) after the build step. For example, in other dialog tasks we have another local package  we use called `bootstrap-chat`. For that, the post build script would look like:

```sh
npm link mephisto-task bootstrap-chat
# here we link two packages instead of just one. running a script allows us this flexibility
```

In sum, I propose introducing "lifecycle hooks"-style config params to allow for more flexible configuration.

Also [see my comment above](https://github.com/facebookresearch/Mephisto/pull/795#discussion_r902110605) for how to solve for the optimization issue. We would detect when `post_build_script` is set and could potentially bypass the optimization in that case?

Etesam913:
The post_build flag definitely seems like a good suggestion. It allows for much more freedom in choosing what to run.

I'm not sure that checking if post_build_script is not none can be used to sufficiently skip the optimization, however. What I mean is that it doesn't work in both directions.

If the flag does exist then you can skip the optimization and build it. What if you want to go back to the package.json package version and remove your link? Well then you would remove the flag. However, if the webapp is not touched, then no rebuild will be triggered and the linked version will still be used. You would have to make some arbitrary change to the webapp to force the rebuild.

I think the simplest way to do this would be to add a param called `force_rebuild` that when set to true would skip over the optimization. It would be good for local development and to prevent these weird situations with building.

pringshia:
> I think the simplest way to do this would be to add a param called `force_rebuild` that when set to true would skip over the optimization. It would be good for local development and to prevent these weird situations with building.

Let's do it. Explicit is always better.

Etesam913:
The **Cypress Testing** and the **Local Package Development** sections of https://github.com/facebookresearch/Mephisto/blob/d5e846d23619da190ab54d84b291d7980cb48b74/CONTRIBUTING.md go into detail about the new changes

mephisto/data_model/task_run.py
```diff
@@ -113,6 +113,27 @@ class TaskRunArgs:
...
+        default=False,
+        metadata={
+            "help": (
+                "Determines if npm run build should be ran every time the task is ran."
+                "By default there is an optimization that only builds the webapp when there is a change in its contents."
+                "It would make sense to turn this on when doing local package development as you want to force a rebuild after linking"
+            )
+        },
+    )
+
```
Etesam913, line 136: These are the two new properties that were added

mephisto/tools/scripts.py
```diff
@@ -237,38 +239,45 @@ def build_custom_bundle(custom_src_dir):
...
         raise Exception(
             "please make sure npm is installed, otherwise view "
             "the above error for more info."
         )
+    print(run_config.task.post_build_script)
+    if (
+        run_config.task.post_build_script is not None
+        and len(run_config.task.post_build_script) > 0
+    ):
+        subprocess.call(["bash", run_config.task.post_build_script])
```
Etesam913, line 280: This is where the post_build_script is ran. It is ran after npm install to prevent the links from getting overridden by the package.json.

mephisto/data_model/task_run.py
```diff
@@ -113,6 +113,27 @@ class TaskRunArgs:
         },
     )
 
+    post_build_script: str = field(
+        default="",
+        metadata={
+            "help": (
+                "The path to a shell script that will run right before npm build."
```
pringshia, line 120: This is confusing. We say right _before_ build in the description text, but the field is called _post_ build script. Should we have called it post-install script or pre-build script instead?

examples/remote_procedure/mnist/hydra_configs/conf/launch_with_local.yaml
```diff
@@ -22,3 +22,5 @@ mephisto:
     task_tags: "mnist,drawing,models,correction"
     # We expect to handle 25 people using the MNIST model at once
     max_num_concurrent_units: 25
+    post_build_script: ""
```
pringshia, line 25: now that you renamed to post install, you'll have to update here as well

examples/remote_procedure/template/hydra_configs/conf/launch_with_local.yaml
```diff
@@ -23,3 +23,5 @@ mephisto:
     # NOTE Model-in-the-loop tasks need to be careful to configure only as many concurrent
     # connections as their model can handle at once
     max_num_concurrent_units: 40
+    post_build_script: ""
```
pringshia, line 26: now that you renamed to post install, you'll have to update here as well

examples/static_react_task/README.md
```diff
@@ -116,4 +116,16 @@ In order to get started on your own task, it is a good idea to copy this `static
...
 8. Iterate more.
-9. Collect some good data.
\ No newline at end of file
+9. Collect some good data.
+
+## Testing
+To run tests locally you should first launch the task as follows:
+
+```bash
+python run_task.py mephisto.task.post_build_script=link_mephisto_task.sh mephisto.task.force_rebuild=true
```
pringshia, line 125: now that you renamed to post install, you'll have to update here as well

mephisto/tools/scripts.py
```diff
@@ -237,39 +234,46 @@ def build_custom_bundle(custom_src_dir):
...
+
     packages_installed = subprocess.call(["npm", "install"])
     if packages_installed != 0:
         raise Exception(
             "please make sure npm is installed, otherwise view "
             "the above error for more info."
         )
 
+    if (
+        run_config.task.post_build_script is not None
```
pringshia, line 272: now that you renamed to post install, you'll have to update here as well

pringshia (commented):
there a bunch of places we need to update to post install script now

Etesam913:
Yeah, for some reason I forgot to update it

Rebecca-Qian (approved):
Looks good - thanks for adding tests!!

CONTRIBUTING.md
```diff
@@ -14,7 +14,49 @@ We actively welcome your pull requests.
...
+If you are modifying either the `mephisto-task` or `mephisto-worker-experience` packages you probably want to see your changes propagate to the task that you are working on.
+
+The easiest way to do this is to run a task by doing:
+```bash
+python run_task.py mephisto.task.post_install_script=link_mephisto_task.sh mephisto.task.force_rebuild=true
+```
+
+Setting `mephisto.task.force_rebuild=true` runs `npm build` before running your task. By default the task is only rebuilt if a file is changed in the webapp, not if a linked package is changed.
+
+Setting `mephisto.task.post_install_script=link_mephisto_task.sh` runs the `link_mephisto_task.sh` script after `npm install` is ran and before the task is started. This script should be located in the webapp folder of the task. While this script can do many things, for local package development the primary purpose of it is to link to a local package.
```
JackUrb, line 51: As a note - we can probably get these to automatically trigger as part of the test suite when any file in `packages` is updated, such that we get this 'for free' on PRs.

examples/remote_procedure/mnist/hydra_configs/conf/launch_with_local.yaml
```diff
@@ -22,3 +22,5 @@ mephisto:
     task_tags: "mnist,drawing,models,correction"
     # We expect to handle 25 people using the MNIST model at once
     max_num_concurrent_units: 25
+    post_install_script: ""
+    force_rebuild: false
```
JackUrb, line 26: Given these have a default hydra value (introduced in `TaskRun`'s file) is it required that we include these defaults in all of these files? 

I imagine the tradeoff between discoverability and boilerplate is an important one to think about if not, but I'd lean towards these more landing in the camp of a heavy user or contributor.

JackUrb (commented):
Will need to execute the update to a 3-param `build_custom_bundle` as discussed, but overall I'm really excited for this PR. Thanks for getting this together! 

We'll definitely want to take the learnings from here to eventually have a test for all of the example tasks, as well as a workflow for being able to create new tests for new tasks (with different configuration options as well like onboarding for instance). Will leave as future work for now though!

mephisto/tools/scripts.py
```diff
@@ -234,8 +237,21 @@ def build_custom_bundle(custom_src_dir, run_config: DictConfig):
...
+        warnings.warn(
+            "Consider adding the force_rebuild property to your hydra config and passing it into build_custom_bundle()\nSteps on how to do this: \nhttps://github.com/facebookresearch/Mephisto/issues/811",
+            stacklevel=2,
+        )
+
+    if "post_install_script" not in kwargs:
+        warnings.warn(
+            "Consider adding the post_install_script property to your hydra config and passing it into build_custom_bundle()\nSteps on how to do this: \nhttps://github.com/facebookresearch/Mephisto/issues/811",
+            stacklevel=2,
+        )
```
Etesam913, line 251: New warning messages:
They link to a closed issue that describes how to fix this

examples/remote_procedure/mnist/hydra_configs/conf/launch_with_local.yaml
Etesam913: As most people will not do local package development, I would think it makes sense to remove them from the hydra config.

I just tested it, and it works without defining them in the hydra config(uses the TaskRun default value).

examples/remote_procedure/toxicity_detection/run_task.py
```diff
@@ -63,7 +63,11 @@ def handle_with_model(
     )
 
     task_dir = cfg.task_dir
-    build_custom_bundle(task_dir, cfg.mephisto)
+    build_custom_bundle(
+        task_dir,
+        force_rebuild=cfg.mephisto.task.force_rebuild,
+        post_install_script=cfg.mephisto.task.post_install_script,
```
Etesam913, line 69: An example of passing in the new properties

mephisto/tools/scripts.py
```diff
@@ -221,7 +224,7 @@ def augment_config_from_db(script_cfg: DictConfig, db: "MephistoDB") -> DictConf
     return script_cfg
 
 
-def build_custom_bundle(custom_src_dir, run_config: DictConfig):
+def build_custom_bundle(custom_src_dir, **kwargs: Union[str, bool]):
```
Etesam913, line 227: kwargs is type Union[str, bool] because the parameter value can be a string(post_install_script) or bool(force_rebuild) currently.

JackUrb: Shouldn't you be able to just write this as:
```
def build_custom_bundle(custom_src_dir, force_rebuild: Optional[bool] = None, post_install_script: Optional[str] = None)
```

Etesam913: Yeah, I can do it like that, I was just following Pratik's suggestion here: https://github.com/facebookresearch/Mephisto/pull/791#discussion_r914119245

pringshia: @Etesam913, you'd still be using kwargs with @JackUrb's suggestion above by naming the arguments in the list

more info: https://treyhunner.com/2018/04/keyword-arguments-in-python/

Etesam913: Oh so kwargs just refers to passing the arguments with keywords as opposed to using the **kwargs parameter. I got it 👍

.github/workflows/cypress-end-to-end-tests.yml
```diff
@@ -0,0 +1,51 @@
...
+          yarn build
+          npm link
+
+      - name: ⌛️ Running cypress tests
+        uses: cypress-io/github-action@v3.1.0
+        with:
+          install: false
+          project: ./examples/static_react_task/webapp
+          config-file: ./cypress.config.js
+          start: python examples/static_react_task/run_task.py mephisto.task.post_install_script=link_mephisto_task.sh
```
Etesam913, line 48: > As a note - we can probably get these to automatically trigger as part of the test suite when any file in `packages` is updated, such that we get this 'for free' on PRs.

I think I am already doing that in this github action

pringshia:
Thanks for getting e2e testing finally added to the codebase here

PR #798, opened 2022-06-17
Fixes bug with remoteFunction() not existing in remote_procedure template

Etesam913:
## Summary

In the app.jsx a function named remoteFunction was being extracted from mephistoProps (mephistoProps comes from the mephisto-task package). This function does not seem to exist anymore as seen by the below screenshot:
template-bug

I'm pretty sure this function got renamed to remoteProcedure()

## Testing
You can clearly see from the video below, that the template example works now

https://user-images.githubusercontent.com/55665282/174390966-ef8371be-f252-4375-a299-db60924437ad.mov


pringshia (approved):
Thanks for finding and fixing!

PR #800, opened 2022-06-18
Add tests for remote examples

Etesam913:
## Summary

Adds Cypress tests for the remote_procedure/template, remote_procedure/mnist, and remote_procedure/toxicity_detection examples.

Three GitHub action jobs are added. One for each of the remote_procedure examples. They just install the necessary dependencies and run the cypress tests.

### Toxicity Detection
https://user-images.githubusercontent.com/55665282/175695870-9e7d5f60-8e1f-4c9d-8a4a-c982421d7fc8.mov
### Template
https://user-images.githubusercontent.com/55665282/175690344-db97fdf7-078a-4fc5-b40c-08642b9dfe29.mov
### Mnist(Updated with drawing capability)

https://user-images.githubusercontent.com/55665282/179091752-9fcfccb0-b9c4-40f3-b8db-5be704d39f07.mov


examples/remote_procedure/mnist/webapp/cypress/e2e/remote_procedure_mnist.cy.js
```diff
@@ -0,0 +1,63 @@
...
+    cy.get('[data-cy="correct-text-input-2"]').should("be.disabled");
+
+    cy.get("@clearButton0").should("be.visible");
+    cy.get("@clearButton0").click({ force: true });
+
+    cy.get("@clearButton1").should("be.visible");
+    cy.get("@clearButton1").click({ force: true });
+
+    cy.get("@clearButton2").should("be.visible");
+    cy.get("@clearButton2").click({ force: true });
```
Etesam913, line 61: Going to re investigate this to see if it can be done cleaner

examples/remote_procedure/toxicity_detection/webapp/cypress/e2e/remote_procedure_toxicity_detection.cy.js
```diff
@@ -0,0 +1,54 @@
...
+    cy.get("@submitButton").click();
+    cy.get('[data-cy="loading-spinner"]');
+
+    cy.get('[data-cy="toxicity-alert"]', { timeout: 25000 }).as(
+      "toxicityAlert"
+    );
+    cy.get("@toxicityAlert").contains(
+      'The statement, "I hate bob!," has a toxicity of:'
+    );
+  });
```
Etesam913, line 33: A long timeout is required for this as the detoxify model takes a long time to run on the GitHub action (they have slow computers).

examples/remote_procedure/mnist/webapp/cypress/e2e/remote_procedure_mnist.cy.js
```diff
@@ -0,0 +1,87 @@
...
+    cy.get('[data-cy="current-annotation-0"]').should("contain.text", "4");
+    cy.get('[data-cy="correct-checkbox-0"]').check();
+    cy.get('[data-cy="correct-text-input-0"]').should("not.exist");
+
+    // draw 1
+    cy.get('[data-cy="canvas-mouse-down-container-1"]')
+      .trigger("mouseover")
+      .trigger("mousedown", 100, 20)
+      .trigger("mousedown", 120, 200)
+      .trigger("mouseup", 120, 200);
```
Etesam913, line 60: This draws a 1, but in the GitHub action it seems to sometimes think it is a 8 leading to the test failing about 1 out of every 4 times. Might make more sense to draw a number that is clearer to the model.

examples/remote_procedure/mnist/webapp/cypress/e2e/remote_procedure_mnist.cy.js
Etesam913: Yeah, the force: true was not needed

examples/remote_procedure/mnist/webapp/cypress/e2e/remote_procedure_mnist.cy.js
Etesam913: It feels kinda weird though because given the exact same mouse down values I feel like the model should give me the exact same result, but I guess that is not the case. AI stuff I guess 🤷

Etesam913: Changed the 1 to a 3, hopefully this is more consistent

JackUrb (commented):
Overall these tests make sense to me!

Etesam913:
> > I don't really know if there is a way to alleviate this, but in some of the tests I click the submit button. This makes locally running the tests a second time fail. The task understandably isn't expected to work after it has been submitted. One workaround is to just comment out the line where you click the submit button in the test. This would allow you to run the test on the task many times without it failing because of submission.
> 
> Another possible approach is to launch multiple tasks, and have a "submit" be the last step of a test, then other tests can use a different `worker_id` in the URL.

It may not even be related to what I said to be honest,  not exactly sure

Here's a video of the problem.


https://user-images.githubusercontent.com/55665282/179279144-fc172f09-948f-4e4f-89a8-2b1c9db23774.mov


I think if you submit the task enough times it will fail or something like that.

JackUrb:
Ah okay so the backend server ends up turning off as it thinks it's "finished" collecting tasks. I don't think we have a great way of dealing with this outside of just launching a "reasonable" amount to cover all of the testing, and then note that too many submissions will eventually lead to tests failing once the server shuts down.

JackUrb (approved):
Implementation looks good to me, excited to have these tests in!

Issue #388, opened 2021-02-03
Better CLI output


pringshia:
It would be great to think of Mephisto's output from a user-first perspective. Here's a mockup of what a more friendly output stream could look like. In this case, I move away from just using logging as output as mentioned in #386 and focus on a more first-class support for CLI output.

### GIF:

![rich_print](https://user-images.githubusercontent.com/425059/106812180-4d3fd800-663d-11eb-9fb8-3e03951d28fe.gif)

### Still image:

<img width="1325" alt="Screen Shot 2021-02-03 at 4 32 52 PM" src="https://user-images.githubusercontent.com/425059/106812353-8b3cfc00-663d-11eb-8f75-6e7600ff6434.png">

---

### Mockup source:

Requires the `rich` package, `pip install rich`:

```python
from time import sleep
from rich.console import Console
from rich.prompt import Confirm

console = Console()
DELAY_OFF = 0

def print_summary(task):
    from rich.table import Table, Column
    from rich import box
    import locale
    locale.setlocale(locale.LC_ALL, 'en_US.UTF-8') 

    table = Table("property", Column("value", style="magenta bold"), "description", title="Task Launch Preview", box=box.ROUNDED, expand=True, show_lines=True)
    # table = Table("task_name", "blueprint", "architect", "crowd provider", "requester", title="Task Launch Preview")
    # table.add_row("my_custom_task", "simple_static_task", "local", "mturk_sanbox", "demo")
    table.add_row("task_name", "my_custom_task", "ID used to later query all HITs launched in this run.\nConfigure via: task.task_name.")
    table.add_row("blueprint", "simple_static_task", "The base template for your task. Can be found in the folder: mephisto/abstractions/blueprints.")
    table.add_row("architect", "local", "Where this task will be hosted/launched from. Configure via: mephisto/architect=...")
    table.add_row("crowd provider", "mturk_sandbox", "The underlying platform that supplies the crowdworkers (e.g. mturk, mturk_sandbox).")
    table.add_row("requester", "demo", "The account/budget to run the task through. Configure via: mephisto.provider.requester_name=... Or use the mephisto CLI for more info: 'mephisto requesters'")


    console.print(table)

    billing_table = Table("num HITs", "price", "estimated total", title="Estimated Costs")
    billing_table.add_row(str(task['hits']), locale.currency(0.02), f"[bold green]{locale.currency(task['hits'] * 0.02)}[/bold green]")
    console.print(billing_table)

def log(category, message):
[...]

PR #813, opened 2022-07-06
Improve cli for tips review

Etesam913:
## Summary
Uses the information in #388 to improve the cli for the review_tips_for_task.py and remove_accepted_tip.py scripts.

I tested using `rich`(cli formatting python package) out on these two scripts first because they are of low importance. Once this is reviewed  then more important commands like `mephisto wut` can potentially be reformatted using `rich`. 

### New Video:
https://github.com/facebookresearch/Mephisto/pull/813#issuecomment-1177801561

### Old Video:
https://user-images.githubusercontent.com/55665282/177645781-01f45c21-723a-4843-9e34-0639f5b3b5e7.mov


mephisto/scripts/local_db/remove_accepted_tip.py
```diff
@@ -1,30 +1,43 @@
-from typing import Set
 from mephisto.abstractions.databases.local_database import LocalMephistoDB
 from mephisto.scripts.local_db.review_tips_for_task import remove_tip_from_metadata
 from mephisto.tools.data_browser import DataBrowser as MephistoDataBrowser
+from mephisto.utils.rich import console
+from rich import print
+from rich.markdown import Markdown
+from rich.prompt import Confirm
+from rich.table import Table, Column
+from rich import box
```
pringshia, line 9: in the root folder, we would also need to add `rich` to `pyproject.toml` (running `poetry add rich` should [automatically do this](https://python-poetry.org/docs/basic-usage/#specifying-dependencies)) and also update `poetry.lock` I think.

I think we also would need to add this package then manually to `requirements.txt` (or would it be requirements_dev.txt?) - @JackUrb might know more here

This process of adding Python packages should also be added to contributing.md I think

JackUrb:
Oy this is tough, I'm thinking the best approach _right now_ would be to drop the `rich` requirement (considering it's only used in these scripts) for this merge, and for me to fix the `poetry` + `requirements.txt` issues more stably (perhaps addressing #768 in the process.

pringshia:
> I'm not exactly sure how I would make the reason field optional. Should I just add an extra question asking if you want to write a reason? I feel like there might already be too many steps for accepting a tip.

Agreed that there are already too many steps, I would suggest using smart defaults to streamline/optionalize:

```
> Would you like to (a)ccept, (r)eject, or (s)kip this tip? (Default: s):
> How much would you like to bonus the tip submitter? (Default: 0.0):

# if not default/0 show:
> What reason would you like to give the worker for this tip? NOTE: This will be shared with the worker.
   (Default: Thank you for submitting a tip!)
```

## Benefits
- This way you can just hit Enter, Enter, Enter as you browse through all tasks, and doesn't require any state-changing actions to take place requiring money and budget
- Removes need to ask if you want to give a bonus or not. You just need one question, if the answer is 0, then no bonus is implied. We also set a default value here.
- Reason is only shown if bonus is non-zero
- Default reason is provided
- Emphasize that the reason is shared with the worker. This wasn't clear with the current design and is important to emphasize

Etesam913:
> Oy this is tough, I'm thinking the best approach _right now_ would be to drop the `rich` requirement (considering it's only used in these scripts) for this merge, and for me to fix the `poetry` + `requirements.txt` issues more stably (perhaps addressing #768 in the process.

Yeah I just making it so that you need the package installed to run the scripts:
Screen Shot 2022-07-08 at 4 58 11 PM

Etesam913:
> 

Makes sense.

### Here is the video that shows some of the defaults

https://user-images.githubusercontent.com/55665282/178070414-7d11c911-6ec7-4eb4-9227-b45e48f20485.mov

JackUrb (approved):

Issue #567, opened 2021-09-20
[Refactor] Move all data file manipulation to MephistoDB


JackUrb:
# Overview
At the moment a number of the classes we use in Mephisto for storing data or preparing task information are managed by classes outside of the MephistoDB. This prevents us from reaching our ideal state of fully abstracting out the storage backend from the data being collected. While not necessarily high-priority to resolve at the moment, this PR documents the locations where this abstraction is leaking.

Ultimately, many individuals' tasks may find it better to continue doing direct writes in `AgentState`'s for instance, but we should strive for clean abstractions internally.

# Implementation plan
- [ ] Create file writing abstractions in the MephistoDB
- [ ] Update all callsites with this TODO tag to use the MephistoDB abstract writes instead.

PR #814, opened 2022-07-06
Consolidating AgentState metadata

JackUrb:
# Overview
To unblock @Etesam913 on #791, this PR standardizes a flow for reading and writing metadata from the top-level `AgentState` class

After much (internal) deliberation on what should go in the main `AgentState` class and what should still be fragmented out into the subclasses, I decided that _only_ metadata would be considered a first-class feature of `AgentState`'s (rather than all data types) but that the save+load flow would be standardized and updated towards the direction of #567. Happy to revisit this in discussions, but I couldn't find any other clean way to allow new agent types to be flexible with how they handle their data.

# Implementation
New functionality:
- Some `AgentState` abstract methods have been moved behind wrappers such that I can put the save+load lifecycle functions inside
- `AgentState.save_metadata` and `AgentState.load_metadata` have been created as ways to handle IO for the metadata, while direct edits can be made to the `AgentState.metadata` object (a dataclass of type `_AgentStateMetadata`). 

Towards future design, this allows the `update_metadata` method to exist inside of the base `AgentState` class rather than needing implementations throughout the similar classes.

### DIFF
--- a/mephisto/abstractions/_subcomponents/agent_state.py
+++ b/mephisto/abstractions/_subcomponents/agent_state.py
@@ -37,7 +37,7 @@ class _AgentStateMetadata:
     Class to track the first-class feature fields of info about an AgentState.
 
     AgentState subclasses may choose to track additional metadata, but should
-    attach to the agent state subclass directly.
+    put these as attributes of the agent state subclass directly.
     """
 
     task_start: Optional[float] = None
--- a/mephisto/abstractions/_subcomponents/agent_state.py
+++ b/mephisto/abstractions/_subcomponents/agent_state.py
@@ -14,6 +14,10 @@ from typing import (
     Union,
     TYPE_CHECKING,
 )
+from dataclasses import dataclass
+import time
+import weakref
+import os.path
 
 if TYPE_CHECKING:
     from mephisto.data_model.agent import Agent, OnboardingAgent
@@ -23,6 +27,22 @@ if TYPE_CHECKING:
 from mephisto.utils.logger_core import get_logger
 
 logger = get_logger(name=__name__)
+
+METADATA_FILE = "agent_meta.json"
+
+
+@dataclass
+class _AgentStateMetadata:
+    """
+    Class to track the first-class feature fields of info about an AgentState.
+
+    AgentState subclasses may choose to track additional metadata, but should
+    attach to the agent state subclass directly.
+    """
+
+    task_start: Optional[float] = None
+    task_end: Optional[float] = None
+    # TODO other metadata fields can be initialized
 
 
 # TODO(#567) File manipulations should ultimately be handled by the MephistoDB, rather than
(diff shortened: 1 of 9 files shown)


mephisto/abstractions/_subcomponents/agent_state.py
```diff
@@ -24,6 +28,22 @@
...
+    """
+    Class to track the first-class feature fields of info about an AgentState.
+
+    AgentState subclasses may choose to track additional metadata, but should
+    attach to the agent state subclass directly.
+    """
+
+    task_start: Optional[float] = None
+    task_end: Optional[float] = None
+    # TODO other metadata fields can be initialized
```
pringshia, line 45: I was imagining that adding the tips plugin would be as simple as doing `yarn add mephisto-worker-addons` and using it without having to modify the back-end code much. Would this require end users to update the python code to use different subclasses for AgentStateMetadata?

Also what if they wanted to compose different plugins? For example, let's say in the future we introduce a plugin for gamification or leaderboards. It would be nice if adding those plugins just worked with the metadata storage, without having to define and compose the dataclasses.

Not sure if this is the right solution, but just wanted to articulate the use case.

mephisto/abstractions/blueprints/abstract/static_task/static_agent_state.py
```diff
@@ -65,27 +43,29 @@ def get_init_state(self) -> Optional[Dict[str, Any]]:
...
-        if os.path.exists(data_path):
-            with open(data_path, "r") as data_file:
-                self.state = json.load(data_file)
+        if self.agent.db.key_exists(data_path):
+            self.state = self.agent.db.read_dict(data_path)
+            # Old compatibility with saved times
+            if "times" in self.state:
+                assert isinstance(self.state["times"], dict)
+                self.metadata.task_start = self.state["times"]["task_start"]
+                self.metadata.task_end = self.state["times"]["task_end"]
```
pringshia, line 56: tbh I'm happy keeping the times within the "times" top level field, as library managed metadata... and the new metadata field for plugins and extensions to use. it might make some of the compatibility stuff here easier to manage though i see you've already handled it. I don't have a strong opinion on this though

mephisto/abstractions/_subcomponents/agent_state.py
```diff
@@ -24,6 +28,22 @@
...
+METADATA_FILE = "agent_meta.json"
+
+
+@dataclass
+class _AgentStateMetadata:
+    """
+    Class to track the first-class feature fields of info about an AgentState.
+
+    AgentState subclasses may choose to track additional metadata, but should
+    attach to the agent state subclass directly.
```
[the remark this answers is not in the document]
JackUrb: Ah I just mean use their own attributes, this is leftover commentary from when I intended to standardize more. I'll update

mephisto/abstractions/_subcomponents/agent_state.py
JackUrb: Valid concern, but I'm wary of projecting too far here. Additional plugins could piggyback on metadata by adding directly to this data class on functionality we feel is standard or general enough. Anything additional or in development should probably just store on an AgentState+Blueprint if the functionality is too complex to simply be in this metadata dataclass.

mephisto/abstractions/blueprints/abstract/static_task/static_agent_state.py
JackUrb: I think the cost of standardization here for these simple fields is overall worth it. I likely should have done this before.

pringshia (approved):

Etesam913 (approved):
Looks good, I don't see any problems and it ran as expected.

One thing that would make sense to think about (I could do this when I merge into add-tips-example):
There isn't a way to update the metadata generally. As of now this isn't a problem because the metadata only has a task_start and task_end field. There is no reason to update these outside of init and submit.

However, if there is a feature like Tips or something else, it might not make too much sense to create an update method for each new field in metadata.

Something like below can be created in agent_state.py:

```python
def update_metadata(self, property_name: str, property_value):
    if property_name not in self.metadata:
        """
        put some message here to tell user that the metadata field does not exist 
        (could help user know that they have to add a field to the _AgentStateMetadata) 
        """
        return

    else:
        self.metadata[property_name] = property_value
```

PR #817, opened 2022-07-11
[WIP] Dynamically update baseURL for cypress testing

Etesam913:


examples/static_react_task/webapp/cypress/e2e/static_react_task.cy.js
```diff
@@ -1,32 +1,38 @@
-describe("Loads static_react_task", () => {
+const pathToAssignmentId = "cypress/assignment_id.txt";
+
+describe("Requests", () => {
   it("Makes request for agent", () => {
     cy.intercept({ pathname: "/request_agent" }).as("agentRequest");
-    cy.visit("/");
+    cy.readFile(pathToAssignmentId).then((assignmentId) => {
+      cy.visit("/?worker_id=x&assignment_id=" + assignmentId);
+    });
```
Etesam913, line 8: For requests, consist test results seem to occur when the intercept is before the visit statement. This is why this test has to be in its own describe block that does not have the beforeEach function in it.

examples/static_react_task/webapp/package.json
```diff
@@ -26,6 +26,7 @@
     "css-loader": "^3.5.1",
     "cypress": "^10.1.0",
     "file-loader": "^6.0.0",
+    "fs": "^0.0.1-security",
```
Etesam913, line 29: Should be removed, not used

mephisto/operations/operator.py
```diff
@@ -595,4 +606,6 @@ def wait_for_runs_then_shutdown(
                 exc_info=False,
             )
         finally:
+            if assignment_id_file_path is not None:
+                os.remove(assignment_id_file_path)
```
Etesam913, line 610: Delete the file with the assignmentId when the server is shut down as this file should be temporary to prevent confusion.

mephisto/operations/task_launcher.py
```diff
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 
+from locale import strcoll
```
Etesam913, line 8: Delete this import, not used

mephisto/operations/task_launcher.py
```diff
@@ -180,7 +181,7 @@ def generate_units(self):
             if not self.unlaunched_units:
                 break
 
-    def _launch_limited_units(self, url: str) -> None:
+    def _launch_limited_units(self, url: str, assignment_id_file_path: str) -> None:
```
Etesam913, line 184: Should set the parameter to None by default

examples/static_react_task/webapp/cypress.config.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
   video: false,
 
   e2e: {
-    baseUrl: "http://localhost:3000/?worker_id=x&assignment_id=1",
+    baseUrl: "http://localhost:3000/",
```
Etesam913, line 5: The baseUrl is updated

JackUrb (changes requested):
Sorry - I should've noted sooner the behavior of `assignment_id`s as they relate to the `MockArchitect` and `LocalArchitect` vs `Assignments`. In this case, they are considered a **Provider-specific** ID, which is entirely decoupled from our backend `Assignment`'s IDs. As such, you can provide _any_ ID as the URL, but once a worker_id and (provider) assignment_id pair is created, Mephisto links these two for reconnection purposes.

For example:
A task launches assignment IDs 304, 305, 306
Connection with Worker ID X and assignment ID 1 -> assigned 304
Connection with Worker ID Y and assignment ID 1 -> assigned 305
Connection with worker ID X and assignment ID 304 -> assigned 306
Connection with worker ID X and assignment ID 1 -> reconnect to 304

This additional machinery of saving assignment IDs for later use shouldn't be necessary given this. Please feel free to follow up and further disambiguate if you think it's required.

Etesam913:
Yeah, that makes sense. I remember some time ago when switching between tasks I would have some issue with cypress testing and I would manually have to update the baseUrl. I tried it again after what you said and the baseUrl with assignmentId=1 worked. 

I don't really know what the issue was before, but it is clear that this pr is probably not needed. I will close this and if I ever encounter that issue again I will take a look at this pr.

PR #820, opened 2022-07-12
Improve cli for mephisto wut

Etesam913:
## Summary

Improved the cli of the `mephisto wut` command and general cli commands like `mephisto`. 

It should be easier to see options for commands and errors should be more clear.

Tables should be much easier to read.

This is done with the [rich-click](https://github.com/ewels/rich-click) and `rich` packages.

The `rich-click` package gives rich formatting to cli commands. 

```python
import rich_click as click
```
this replaces the default click object to get basic rich formatting. 

The `rich` package is still needed for printing things to the console, such as tables.

## Video
https://user-images.githubusercontent.com/55665282/178544113-1397707d-6b4c-45bd-a598-3d0cbd6cb1b7.mov

PR #822, opened 2022-07-12
Improve cli for other mephisto commands

Etesam913:
Just building off of #820 

Moves away from the tabulate method in favor of the Table import in `rich`. 

### `mephisto requesters`
mephisto-requesters

### `mephisto register`
mephisto-register
Added red color & provided list of valid providers in case the user forgot

### `mephisto register <provider_name>`
mephisto-register-provider-name

### `mephisto metrics`
mephisto-metrics


JackUrb (approved):
LGTM - thanks for these QoL improvements across the codebase!

PR #832, opened 2022-07-17
Adds http:// before launch message url

### LINKED ISSUE #828: Make URLs in terminal output clickable on launch
CleanShot 2022-07-15 at 10 43 57@2x

It's annoying to have to copy and paste the URLs from the terminal output into a browesr on task launch.

```
# change:

localhost:3000/?worker_id=x&assignment_id=1

# to:

http://localhost:3000/?worker_id=x&assignment_id=1
```

Adding a protocol (http: / https://) before the URL will make it easy to simply click on them to open (in some terminals). We should add this.

---

Note: I'm not sure if we need to decide between http or https based on certain scenarios

Etesam913:
### Resolves #828
* Adds http:// before the url that appears when a mock unit is launched. 
* This allows for links to be visited in an easier fashion.

## Video
https://user-images.githubusercontent.com/55665282/179418063-26efe9ca-07b5-4574-a363-0f1416a3ff36.mov

### DIFF
--- a/mephisto/abstractions/providers/mock/mock_unit.py
+++ b/mephisto/abstractions/providers/mock/mock_unit.py
@@ -49,12 +49,12 @@ class MockUnit(Unit):
         port = task_url.split(":")[1].split("/")[0]
         print(task_url)
         print(
-            f"Mock task launched: localhost:{port} for preview, "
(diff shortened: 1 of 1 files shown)


JackUrb (approved):
LGTM for `MockUnit`s. Could also add something similar with `https://` for `MTurkUnit`s.

PR #833, opened 2022-07-17
Add tests to tips example

Etesam913:
## Overview

Adds three cypress end to end tests for the `static-react-task-with-tips` task example.
1. 0-static_react_task_with_tips_pre_tip_submission.cy
2. 1-static_react_task_with_tips_feedback_submission.cy
3. static_react_task_with_tips_post_tip_submission.cy

### Pre-tip submission
This tests if the tips popup can be opened. It does error checking by typing messages that are too long. It also submits three tips and checks if the server response is correct.

#### Video:

https://user-images.githubusercontent.com/55665282/179538283-922f90e3-8f13-4ee0-8c81-16b749038cef.mp4

### Feedback submission
Checks if the correct elements are present and submits answers to the two feedback questions in the example.

#### Video:
https://user-images.githubusercontent.com/55665282/179541492-9788612d-bc60-431f-82ec-069619db7e64.mov

### Post-tip submission
After the pre-tip submission test there are three tips that need to be reviewed. Prior to running the post-tip submission test make sure to run `python review_tips_for_task.py` and accept the first tip and reject the other two tips.

The test checks to see if the first submitted tip from the pre-tip submission test is present. 

#### Video:
https://user-images.githubusercontent.com/55665282/179540956-24869fa1-c98f-46b9-bbdd-0692713182a2.mov

### GitHub action file
Adds a new job to run the cypress test above.
1. Installs dependencies and uninstalls detoxify(it is too slow for the GitHub action)
2. Link to local version of `mephisto-task` and `mephisto-worker-addons` packages.
3. Runs the pre-tip submission cypress test and the feedback_submission cypress test
    * The numbers in the file names of these two tests are set the order that the tests are ran in.
4. Kills the python and node server
    * This is supposed to happen automatically after cypress finishes, but this does not happen so I have to do it manually.
5. The first two units that are launched are expired by running the `expire_all_units.py` script
    * Again, this should happen automatically from shutting down the server, but it doesn't seem to happen in the GitHub action.
    * Expiring the units is important because this state is required for the reviewing steps later on.
6. First and second tips are accepted, while the third tip is rejected. 
    * This is done by running the `review_tips_for_task.py` script.
7. Second tip is removed by running `remove_accepted_tip.py`.
8. Reviews the feedback by running `review_feedback_for_task.py`
9. The post-tip submission test is ran


mephisto/scripts/local_db/expire_all_units.py
```diff
@@ -0,0 +1,24 @@
+from mephisto.abstractions.databases.local_database import LocalMephistoDB
```
JackUrb, line 1: I feel like this should go in a `gh_actions` folder

JackUrb (commented):
Overall the test here seems great, thanks for adding it! One note on the added script

JackUrb (approved):

mephisto/scripts/local_db/gh_actions/expire_all_units.py
```diff
@@ -0,0 +1,24 @@
+from mephisto.abstractions.databases.local_database import LocalMephistoDB
+
+"""
+This script should not be used locally.
+It is only used in a GitHub action as it does not automatically expire units.
```
pringshia, line 5: this script seems very "one-off"-y. is there no other case we may want to intentionally force expire units? I assume it could be useful? I would consider parametizing the task_name below and having this be a general script?

pringshia (approved):
a few comments, but overall LGTM

mephisto/scripts/local_db/gh_actions/expire_all_units.py
Etesam913: I'm not aware of any situation outside of the GitHub action where you would want to run a script to expire units. Usually you would expire units by just doing ctrl-c two or three times on your mephisto task.

JackUrb: "Force expiring" units is generally done with a `cleanup` script, which is provider-associated. This only updates the local status, and is a particular anti-pattern outside of the context of eval scripts.

examples/static_react_task_with_tips/webapp/cypress/e2e/pre_submission_tests/0-static_react_task_with_tips_pre_tip_submission.cy.js
```diff
@@ -0,0 +1,163 @@
...
+      expect(interception.response.statusCode).to.eq(200);
+    });
+  });
+
+  it("Loads correct task react elements", () => {
+    cy.visit("/");
+    cy.get('[data-cy="directions-container"]');
+    cy.get('[data-cy="task-data-text"]');
+    cy.get('[data-cy="good-button"]');
+    cy.get('[data-cy="bad-button"]');
```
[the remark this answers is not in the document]
Etesam913: Yeah, I don't think it is necessary. I mainly use `cy.should('not.exist')` to check for content that is not on the page, but I don't really know the point of using `cy.should('exist')`

mephisto/scripts/local_db/gh_actions/expire_all_units.py
pringshia: got it... if this script is really needed for a test, I guess I'm okay with keeping it then

examples/static_react_task_with_tips/webapp/cypress/e2e/pre_submission_tests/0-static_react_task_with_tips_pre_tip_submission.cy.js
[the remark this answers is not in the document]
pringshia: yeah it seems it's redundant, it maybe just indicates intention that an assertion is taking place. i'm fine with keeping it as is

PR #834, opened 2022-07-18
Improves performance of cypress tests

Etesam913:
## Overview
There were additional `cy.visit("/")` statements that were present in the cypress tests because I thought that every "it" block had to have `cy.visit("/")` in it. 

This can be a costly command, so I removed the additional `cy.visit("/")` statements in all the tests.

I also increased the timeout to 40,000 ms for the toxicity detection as 25,000 ms was not always long enough for the github action. It would not pass consistently enough. This issue should be fixed now.


JackUrb (approved):
LGTM, thanks for the fix!

PR #837, opened 2022-07-18
Improves clarity of setting data folder in docs

### LINKED ISSUE #835: Improve docs on setting up mephisto data directory
Screen Shot 2022-07-18 at 1 18 00 PM

I think the docs are kind of confusing when setting up the data directory. As of now it assumes that you cloned your repo in your home directory. For me I cloned the repo in my Documents folder, so the command won't work for me.

It probably makes sense to mention that this command is assuming that you cloned the repo in your home directory.

Me and @pringshia both seemed to have an external data directory in our home directory named mephisto, but I believe this is discouraged as this folder exists in the repo:
https://github.com/facebookresearch/Mephisto/tree/main/data

Etesam913:
### Resolves #835 
* Deletes the data folder in the repo as the data directory should be external.
* Adds an extra sentence to give more clarity on what should be done before running the command.

## Screenshot

data-dir-update

### DIFF
--- a/data/README.md
+++ b/data/README.md
@@ -1,4 +1,0 @@
-## Data
-The data folder is populated with any data that the Mephisto system is using to keep state and store results.
-
-This is the default location for such storage, however the location is configured in `~/.mephisto/config.yml`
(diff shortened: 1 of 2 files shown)


pringshia:
Slight nuance from the PR description:

> should be external

is wrong... it *could be*

PR #841, opened 2022-07-19
Adds a new cli command, 'scripts'

Etesam913:
## Overview
After submitting a tip and closing the server it it was annoying to have to go to a separate directory to find the script needed to review the tips. 

I would often do this navigation in a separate terminal window as I didn't want to lose my current task directory. This would be annoying as I would often have to setup the virtual env in the new terminal window as well.

This problem is gone with the new `mephisto scripts` cli command. It allows you to review tips/feedback and run other scripts no matter your current directory.

### Updated Video:
https://github.com/facebookresearch/Mephisto/pull/841#issuecomment-1190495428

### Old Video:
https://user-images.githubusercontent.com/55665282/179857805-2f48a928-bbca-4afa-9e44-2e4c184f46f7.mov

In the video I first shut down my task after I got my tip/task submissions. Then in the same directory I review the tips by running `mephisto scripts review_tips`


JackUrb (commented):
I really like the initiative, however I'm not sure why you're restricting to only the `local_db` scripts. I could see a benefit in being able to run `mephisto scripts mturk cleanup` or something as well.

Understandably knowing how to fill through the options becomes more complicated in that case.

Etesam913:
Yeah, I just forgot to do that. I will add them

Etesam913:
Updated version of the script. Added an additional script_type parameter

https://user-images.githubusercontent.com/55665282/180034249-f62260bd-3aa9-470f-8bb3-d059366b10ad.mp4

JackUrb (approved):
Happy to have this improvement!

I imagine there's a nicer way to construct the script list and mapping, but there's no need to go for that right now, this is much better than what we had before.

PR #854, opened 2022-07-20
Upgrade to yarn 3.2.1

pringshia:
Upgrading to yarn 3.2.1 (latest). Fixes some issues I was getting around changing node versions and nvm: https://github.com/yarnpkg/berry/issues/4581#issuecomment-1166958170

The issue was this on `yarn install`:
```
Internal Error: Unable to deserialize cloned data due to invalid or unsupported version.
    at Object.deserialize (v8.js:343:7)
```

I also had to upgrade `@yarnpkg/plugin-version` as part of the upgrade: https://github.com/yarnpkg/berry/issues/3615


Etesam913 (approved):
Tried it locally, works with no problems 👍

PR #857, opened 2022-07-21
Add path filter

Etesam913:
## Overview
* Adds path filtering on the job level to allow for jobs to be skipped if certain files are not changed.

github-actions-architecture

The base branch is looked at(in this case `add-tests-to-tips-example`) to determine what files are changed.

As commit [62faad8](https://github.com/facebookresearch/Mephisto/pull/857/commits/62faad800a4a68dfbdecc124f369138217877331) updates a file in the static_react_task folder, the static_react_task test gets ran and the others don't.
tests-being-ran

### DIFF
--- a/.github/workflows/cypress-end-to-end-tests.yml
+++ b/.github/workflows/cypress-end-to-end-tests.yml
@@ -5,7 +5,56 @@ on:
     branches: [main]
 
 jobs:
+  changes:
+    runs-on: ubuntu-latest
+    # Set job outputs to values from filter step
+    outputs:
+      static_react_task: ${{ steps.filter.outputs.static_react_task }}
+      static_react_task_with_tips: ${{ steps.filter.outputs.static_react_task_with_tips }}
+      mnist: ${{ steps.filter.outputs.mnist }}
+      template: ${{ steps.filter.outputs.template }}
+      toxicity_detection: ${{ steps.filter.outputs.toxicity_detection }}
+      abstractions: ${{ steps.filter.outputs.abstractions }}
+      data_model: ${{ steps.filter.outputs.data_model }}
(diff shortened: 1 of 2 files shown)


.github/workflows/cypress-end-to-end-tests.yml
```diff
@@ -5,7 +5,56 @@ on:
...
+      operations: ${{ steps.filter.outputs.operations }}
+      tools: ${{ steps.filter.outputs.tools }}
+      mephisto-task: ${{ steps.filter.outputs.mephisto-task }}
+      mephisto-worker-addons: ${{ steps.filter.outputs.mephisto-worker-addons }}
+    steps:
+      # For pull requests it's not necessary to checkout the code
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 2
+      - uses: dorny/paths-filter@v2
```
Etesam913, line 28: This GitHub action is used to allow for path filtering on the job level. 

GitHub actions natively supports path filtering on the action level. This is too broad for this use-case as it would look something like examples/**.

This would mean that all the tests would run every time a file is changed in static_react_task.

This solution allows only for the static_react_task test to run when a file is changed in static_react_task

JackUrb (approved):
This is awesome. Filtering will definitely help overall!

One _small_ change I'd make though, we'd expect that changes to the underlying architecture that could break `mnist` and `toxicity_detection` (our slowest tests here) would _also_ break `template`, so we can drop the 
```
(needs.changes.outputs.abstractions == 'true') || (needs.changes.outputs.data_model == 'true') || (needs.changes.outputs.operations == 'true') || (needs.changes.outputs.tools == 'true')
```
triggers from these cases. Otherwise this is a solid DevX improvement!

JackUrb (approved):

PR #861, opened 2022-07-25
Make `self.metadata` backward-compatible

chiehminwei:
This patch makes `self.metadata` in `parlai_chat_agent_state` backward-compatible by adding fallback options when the `metadata` field is not present in the state.


Etesam913:
You might want to run these three commands to make sure that the code that you added is formatted properly. 
Screen Shot 2022-07-26 at 10 06 26 AM
 
This will make sure that your code passes the pre-commit / code-style (pull_request) test.

JackUrb (approved):
LGTM thanks for the fix!

PR #864, opened 2022-07-26
fix breaks in create_validation_function for screening units

macota:
Per offline discussions, adding additional checks to use max_screening_units=0 as a way to denote screening on real data

### DIFF
--- a/mephisto/abstractions/blueprints/mixins/screen_task_required.py
+++ b/mephisto/abstractions/blueprints/mixins/screen_task_required.py
@@ -217,7 +217,7 @@ class ScreenTaskRequired(BlueprintMixin):
                 args.blueprint.max_screening_units == 0
                 and agent.get_worker().is_qualified(passed_qualification_name)
             ):
-                return
+                return  # Do not run validation if running with no screening_data_factory and worker is qualified
             validation_result = screen_unit(unit)
             if validation_result is True:
                 agent.get_worker().grant_qualification(passed_qualification_name)
--- a/mephisto/abstractions/blueprints/mixins/screen_task_required.py
+++ b/mephisto/abstractions/blueprints/mixins/screen_task_required.py
@@ -205,13 +205,16 @@ class ScreenTaskRequired(BlueprintMixin):
         failed_qualification_name = args.blueprint.block_qualification
 
         def _wrapped_validate(unit):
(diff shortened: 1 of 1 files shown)


mephisto/abstractions/blueprints/mixins/screen_task_required.py
```diff
@@ -205,13 +205,16 @@ def create_validation_function(
...
+            if args.blueprint.max_screening_units > 0 and unit.unit_index >= 0:
                 return  # We only run validation on the validatable units
-
             agent = unit.get_assigned_agent()
             if agent is None:
                 return  # Cannot validate a unit with no agent
-
+            if (
+                args.blueprint.max_screening_units == 0
+                and agent.get_worker().is_qualified(passed_qualification_name)
```
macota, line 215: Was slightly confused here if there was a significant difference between checking if it had been granted passing qualification vs checking if it _hadn't_ been granted failing qualification

i.e 
`agent.get_worker().is_qualified(passed_qualification_name)` vs `not agent.get_worker().is_qualified(failed_qualification_name)`

JackUrb: Good question - if they've already failed, then they shouldn't be able to get this far, so this is the accurate check.

mephisto/abstractions/blueprints/mixins/screen_task_required.py
```diff
@@ -205,13 +205,16 @@ def create_validation_function(
...
+            ):
+                return
```
JackUrb, line 217: Only thing I'd request here is a comment about what this break is considering. Should capture not needing to re-validate.

JackUrb (changes requested):
Fix seems great to me! For the purpose of completeness and ease of use, however, we should update the [`ScreenTaskRequired`](https://github.com/facebookresearch/Mephisto/blob/c780e4ab167e3c7b56b60ce70b967476a4cff3e4/mephisto/abstractions/blueprints/mixins/screen_task_required.py#L91) class's `assert_mixin_args` function to ensure that `max_screening_units` is 0 if `screening_data_factory` is false, and nonzero otherwise.

JackUrb (approved):
Thanks for these fixes! Good to have both specific usage types of screening units properly implemented and checked.

PR #865, opened 2022-07-26
Fix mephisto web cli command

### LINKED ISSUE #856: mephisto web command is broken
Running `mephisto web` in main will give you a url which you can go to, but the page is blank once visited. 

When opening the browser console, this is the error message that I get:
Screen Shot 2022-07-21 at 10 44 24 AM

Clicking on the link in the console gives you:
Screen Shot 2022-07-25 at 10 35 59 AM

Etesam913:
## Overview
There were a bunch of bugs that were preventing the server from rendering any content.
* First I fixed the build bugs
  * I had to update `react-router-dom` to the most up to date version
  * There was some axios type issue preventing building

* Then I had to fix the rendering bugs
  * There was a bug where an object was being put right into jsx
    * It was something like this <span>{thisIsAnObject}</span>

## Video
https://user-images.githubusercontent.com/55665282/181117866-ba626a16-2619-4cb8-89f0-aecd0c24bd62.mov

## There may still be some bugs
**For example**: clicking the buttons in the action column (on the review page) doesn't seem to do anything.

## Resolves #856


pringshia:
Thanks for getting this to run! Feels like seeing an old friend again ha.

Btw, you should do an `npm run build` and check in the contents of the build folder. Checking the build in to source was an intentional decision so that client users can just run `mephisto web` without having to trigger the build manually on their end first.

pringshia (approved):

PR #869, opened 2022-07-28
Enhance logging messages with rich

### LINKED ISSUE #386: Nicer looking logging
It would be nice if Mephisto could have some better formatted logging. 

I came across this library, which has a drop-in logging handler for formatting:

https://github.com/willmcgugan/rich#logging-handler

image

I tried to implement this in `logger_core.py` and was able to get it to work to some extent, but had issues like certain log outputs were getting left out, and other log outputs were getting duplicated (printed both richly and plainly), so maybe someone else can take a look into this. Jack mentioned that the 1.1 release of Hydra should also make logging a bit easier.

### LINKED ISSUE #388: Better CLI output

Etesam913:
## Summary
Many of the scripts are beautified using rich, why not add this to the logs.

Added a `RichHandler` object to the logging config to allow for clearer logging messages.

## Updated Video:
https://github.com/facebookresearch/Mephisto/pull/869#issuecomment-1198627769

## Video

https://user-images.githubusercontent.com/55665282/181590554-238b7ace-26c5-4fc1-9141-61873c3180aa.mov

## Duplicated Log Messages Issue (RESOLVED) https://github.com/facebookresearch/Mephisto/pull/869#issuecomment-1198627769
One thing that I want to fix is the duplicated log messages. You can see that each colorful log message has a uncolored version. 

I believe this is because the main logger is also logging the messages, which leads to duplicated logging. One way to fix this is to set 
```python
logger.propagate = False
```
This does work, however, it means that the messages won't get saved in the `outputs/<date>/scripts.log` file. This is a pretty big issue, so I didn't apply it. Removing the duplicated messages does make the logs much easier to read though.

## Resolves #386

### DIFF
--- a/mephisto/operations/hydra_config.py
+++ b/mephisto/operations/hydra_config.py
@@ -17,7 +17,6 @@ from typing import List, Type, Dict, Any, TYPE_CHECKING
 
 
 if TYPE_CHECKING:
-    from mephisto.abstractions.architect import Architect
     from mephisto.abstractions.blueprint import Blueprint
 
 
@@ -46,7 +45,7 @@ class TaskConfig:
     mephisto: MephistoConfig = MephistoConfig()
     task_dir: str = get_run_file_dir()
     num_tasks: int = 5
-    defaults = ["_self_", LOGGING_OVERRIDE]
+    defaults: List[Any] = field(default_factory=lambda: ["_self_", LOGGING_OVERRIDE])
 
 
 def register_abstraction_config(name: str, node: Any, abstraction_type: str):
--- a/mephisto/operations/hydra_config.py
+++ b/mephisto/operations/hydra_config.py
@@ -22,7 +22,7 @@ if TYPE_CHECKING:
 
 
 logger = get_logger(name=__name__)
-LOGGING_OVERRIDE = {"override /hydra/hydra_logging": "mephisto_default"}
+LOGGING_OVERRIDE = {"override /hydra/job_logging": "mephisto_default"}
 config = ConfigStoreWithProvider("mephisto")
 
 
--- a/mephisto/operations/hydra_config.py
+++ b/mephisto/operations/hydra_config.py
@@ -22,7 +22,7 @@ if TYPE_CHECKING:
 
 
 logger = get_logger(name=__name__)
-LOGGING_OVERRIDE = {"override /hydra/job_logging": "mephisto_default"}
+LOGGING_OVERRIDE = {"override /hydra/hydra_logging": "mephisto_default"}
 config = ConfigStoreWithProvider("mephisto")
 
 
--- a/mephisto/operations/hydra_config.py
+++ b/mephisto/operations/hydra_config.py
@@ -22,7 +22,7 @@ if TYPE_CHECKING:
 
 
 logger = get_logger(name=__name__)
-
+LOGGING_OVERRIDE = {"override /hydra/job_logging": "mephisto_default"}
 config = ConfigStoreWithProvider("mephisto")
 
 
@@ -46,6 +46,7 @@ class TaskConfig:
     mephisto: MephistoConfig = MephistoConfig()
     task_dir: str = get_run_file_dir()
     num_tasks: int = 5
+    defaults = ["_self_", LOGGING_OVERRIDE]
 
 
 def register_abstraction_config(name: str, node: Any, abstraction_type: str):
@@ -57,7 +58,7 @@ def register_abstraction_config(name: str, node: Any, abstraction_type: str):
 
 
 def build_default_task_config(conf_name: str) -> Type[TaskConfig]:
-    default_list = ["_self_", {"conf": conf_name}]
+    default_list = ["_self_", {"conf": conf_name}, LOGGING_OVERRIDE]
 
     @dataclass
     class DefaultTaskConfig(TaskConfig):
--- a/test/core/test_live_runs.py
+++ b/test/core/test_live_runs.py
@@ -124,6 +124,16 @@ class BaseTestLiveRuns:
         self.worker_pool.register_run(live_run)
         return live_run
 
+    def assert_server_subbed_in_time(self, server, timeout: int = 5) -> None:
+        start_time = time.time()
+        while (len(server.subs) == 0) and time.time() - start_time < timeout:
+            time.sleep(0.3)
+        self.assertEqual(
+            len(self.architect.server.subs),
+            1,
+            "MockServer doesn't see registered channel",
+        )
+
     def get_mock_assignment_data_array(self) -> List[InitializationData]:
         mock_data = MockTaskRunner.get_mock_assignment_data()
         return [mock_data, mock_data]
@@ -240,11 +250,7 @@ class BaseTestLiveRuns:
         self.assertIsNotNone(channel)
         self.assertTrue(channel.is_alive())
         task_runner = live_run.task_runner
-        self.assertEqual(
-            len(self.architect.server.subs),
-            1,
-            "MockServer doesn't see registered channel",
-        )
+        self.assert_server_subbed_in_time(self.architect.server)
         self.assertIsNotNone(
             self.architect.server.last_alive_packet,
             "No alive packet received by server",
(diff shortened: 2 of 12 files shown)


JackUrb (commented):
The duplicate logging is a symptom of [hydra's default logging](https://hydra.cc/docs/tutorials/basic/running_your_app/logging/). This _should_ be [configurable](https://hydra.cc/docs/configure_hydra/logging/).

With a bit of digging I'd suggest
1. Editing `hydra_config.py` to include a `LOGGING_OVERRIDE = {'override hydra/job_logging': 'mephisto_default'}` override constant.
2. Creating a file called `mephisto_default.yaml` in the specific folder `hydra_configs/hydra/job_logging`.
3. Include the `LOGGING_OVERRIDE` as the last value in a defaults list provided both to the standard `TaskConfig` and the version produced by `build_default_task_config`.

Now as far as what to do when you've gotten this far, I'm not 100% sure. Either the choice is going to be to make the `hydra_log_file.yaml` use the rich logger instead of the simple one (more likely), or to make it so that the `hydra_log_file` resembles the [disabled](https://github.com/facebookresearch/hydra/blob/1.2_branch/hydra/conf/hydra/hydra_logging/disabled.yaml) setup to block hydra's logs to console if that doesn't work.

Etesam913:
Got it working 👍 I had to make sure that the hydra job_logging did not write results to console, but did write to file.

## New Video

https://user-images.githubusercontent.com/55665282/181636653-4c9528a9-0f90-415e-b463-b4027c9a4f06.mov

Etesam913:
This is in mock_unit.py:
mock_unit py
Do we have to print this out given that it is already logged?

It seems a little extra.

JackUrb:
> This is in `mock_unit.py`: mock_unit py Do we have to print this out given that it is already logged?
> 
> It seems a little extra.

My concern here is if someone has logging entirely disabled, and then doesn't get to see this message

Etesam913:
Makes sense

mephisto/operations/hydra_config.py
```diff
@@ -46,6 +46,7 @@ class TaskConfig:
     mephisto: MephistoConfig = MephistoConfig()
     task_dir: str = get_run_file_dir()
     num_tasks: int = 5
+    defaults = ["_self_", LOGGING_OVERRIDE]
```
JackUrb, line 49: You'll need to use:
```python
defaults: List[Any] = field(default_factory=lambda: ["_self_", LOGGING_OVERRIDE])
``` 
here or Hydra will throw for base `TaskConfig`s.

JackUrb (approved):
Thanks for this change! One small Hydra gotcha but otherwise I'm really excited to have improved, readable logs! Thanks for taking this on :)

test/core/test_live_runs.py
```diff
@@ -124,6 +124,16 @@ def get_mock_run(self, blueprint, task_runner) -> LiveTaskRun:
...
 
+    def assert_server_subbed_in_time(self, server, timeout: int = 5) -> None:
+        start_time = time.time()
+        while (len(server.subs) == 0) and time.time() - start_time < timeout:
+            time.sleep(0.3)
+        self.assertEqual(
+            len(self.architect.server.subs),
+            1,
+            "MockServer doesn't see registered channel",
+        )
```
JackUrb, line 135: ❤️

JackUrb (approved):

pringshia:
Side thought - The duplication of logs here could be an opportunity for clean-up?

CleanShot 2022-08-01 at 11 37 41@2x

JackUrb:
> Side thought - The duplication of logs here could be an opportunity for clean-up?
> 
> CleanShot 2022-08-01 at 11 37 41@2x

Etesam asked about this [above](https://github.com/facebookresearch/Mephisto/pull/869#issuecomment-1199378831). The difficulty here I suppose is that I wanted the link to _always_ appear, even if logging is disabled. A solution would be one that forces the print only if logging isn't active, but I don't know if there's any really elegant way to do that.

PR #870, opened 2022-07-29
Auto Generate Blueprint Documentation

### LINKED ISSUE #696: Auto-generate Blueprint docs
## Context

Currently the best way to get the full list of configuration options for a Blueprint is to use the `mephisto wut` CLI command (see example below). Looking at the source code alone isn't convenient/sufficient as due to inheritance you'd have to walk up the base classes and see all possible options.

## Proposal

It would be nice to autogenerate docs (perhaps this can be a Github Action that monitors the blueprints/ folder) so that we can view a similar output to the `mephisto wut` CLI too, but from within the [API Reference](https://mephisto.ai/docs/reference/overview) section of the docs.

---

#### `mephisto wut` CLI usage:
```
$ mephisto wut blueprint=static_react_task static_task_data

                Tasks launched from static blueprints need
                a prebuilt javascript bundle containing the task. We suggest building
                with our provided useMephistoTask hook.
            

Additional SharedTaskState args from SharedStaticTaskState, which may be configured in your run script
dest              type                      default    help                                     choices    required
----------------  ------------------------  ---------  ---------------------------------------  ---------  ----------
static_task_data  Iterable[Dict[str, Any]]  []         List or generator that returns dicts of  None       False
                                                       task data. Generators can be used for
                                                       tasks with lengths that aren't known at
                                                       the start of a run, or are otherwise
                                                       determined during the run.

```

Etesam913:
## Overview
This pull request adds a script named `auto_generate_blueprint.py` which creates a `Blueprint.md` file with a table for each blueprint value.

Uses the `mdutils` library to create the markdown file

This can be extended to architects, requesters, and providers

## Video
https://user-images.githubusercontent.com/55665282/181821598-e03f7532-e645-4c71-a6fd-97c689244c7b.mov

## Resolves #696


mephisto/scripts/local_db/gh_actions/auto_generate_blueprint.py
```diff
@@ -0,0 +1,78 @@
...
+        rows=len(arg_keys) + 1,
+        columns=len(first_arg_keys),
+        text=list_of_strings,
+        text_align="left",
+    )
+
+
+def main():
+    blueprint_file = MdUtils(
+        file_name="../../../../docs/web/docs/reference/Blueprints.md",
```
Etesam913, line 42: Any way to make this path cleaner?

mephisto/client/cli_commands.py
```diff
@@ -0,0 +1,184 @@
...
+        abstractions_table.add_row(
+            "requester",
+            f"The requester is an account for a crowd provider. \nValid requester types types are [b]{get_valid_provider_types()}[/b]. \n"
+            "\nUse `mephisto requesters` to see registered requesters, and `mephisto register <requester type>` to register.",
+        )
+        abstractions_table.add_row(
+            "provider",
+            f"The crowd provider determines the source of the crowd workers. \nValid provider types are [b]{get_valid_provider_types()}[/b]",
+        )
+        console.print(abstractions_table)
```
pringshia, line 41: stray console statement?

mephisto/scripts/local_db/gh_actions/auto_generate_blueprint.py
```diff
@@ -0,0 +1,77 @@
...
+    args = get_wut_arguments(("blueprint=parlai_chat",))
+    arg_dict = args[0]
+    create_blueprint_info(blueprint_file, arg_dict)
+
+    blueprint_file.new_header(level=2, title="Mock")
+
+    args = get_wut_arguments(("blueprint=mock",))
+    arg_dict = args[0]
+    create_blueprint_info(blueprint_file, arg_dict)
+    blueprint_file.create_md_file()
```
pringshia, line 71: wonder if there's a way to make the above dynamic instead of having to define all of the blueprints manually

mephisto/scripts/local_db/gh_actions/auto_generate_blueprint.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+# ('blueprint=static_react_task',)
```
pringshia, line 74: delete?

docs/web/docs/reference/Blueprints.md
```diff
@@ -0,0 +1,101 @@
...
+
+
+Tasks launched from static blueprints need
+                a prebuilt javascript bundle containing the task. We suggest building
+                with our provided useMephistoTask hook.  
+
+|dest|type|default|help|choices|required|
+| :--- | :--- | :--- | :--- | :--- | :--- |
+|block_qualification|str|???|Specify the name of a qualification used to soft block workers.|None|False|
+|tips_location|str|/Users/etesam1/Documents/Mephisto/mephisto/scripts/local_db/gh_actions/assets/tips.csv|Path to csv file containing tips|None|False|
```
pringshia, line 17: your full local file path is being output and committed to the docs here:

> /Users/etesam1/Documents/Mephisto/mephisto/scripts/local_db/gh_actions/assets/tips.csv

Ideally it just says:

> mephisto/scripts/local_db/gh_actions/assets/tips.csv

I think you can figure out the root folder where the Mephisto dir is located and "subtract that out" from here to get the more relative URL

mephisto/client/cli_commands.py
pringshia: I guess console.print and print are different yet similar?

Etesam913: console.print prints content to your terminal. So this prints out your abstractions_table when you type `mephisto wut`

docs/web/docs/reference/Blueprints.md
Etesam913: I could do that, my idea was that this script gets ran on push to main, so it would get replaced with the GIthub runner, but maybe I should filter it out

mephisto/scripts/local_db/gh_actions/auto_generate_blueprint.py
Etesam913: There is a `get_valid_blueprint_types()` method that I can use in a for loop

mephisto/scripts/local_db/gh_actions/auto_generate_blueprint.py
```diff
@@ -20,6 +22,14 @@ def create_blueprint_info(blueprint_file, arg_dict):
                 if isinstance(item_content, str)
                 else item_content
             )
+            if (
+                isinstance(item_to_append, str)
+                and item_to_append.find("mephisto/") != -1
+            ):
+                item_to_append = item_to_append[
+                    item_to_append.find("mephisto/") : len(item_to_append)
```
pringshia, line 30: this is not very robust. this is obviously a contrived example, but a user can have their code located somewhere like this:

> /User/mephisto/my documents/mephisto/private secret folder/mephisto

The result of this would yield back:

> my documents/mephisto/private secret folder/mephisto

This is why I suggested "subtracting out the root folder" before

mephisto/scripts/local_db/gh_actions/auto_generate_blueprint.py
pringshia: I'm not sure if there's an importable python constant/method floating around the codebase that could get you the root dir. @JackUrb might know

mephisto/client/cli_commands.py
```diff
@@ -154,8 +154,11 @@ def get_wut_arguments(args):
                 checking_args[arg]["required"] = "[b]{requiredVal}[/b]".format(
                     requiredVal=checking_args[arg]["required"]
                 )
+            if arg == "tips_location":
+                checking_args[arg]["default"] = "path_to_task/assets/tips.csv"
```
pringshia, line 158: would be nice to leave a brief comment here explaining why we had to do this, for people down the road looking at this

mephisto/client/cli_commands.py
```diff
@@ -154,8 +154,11 @@ def get_wut_arguments(args):
                 checking_args[arg]["required"] = "[b]{requiredVal}[/b]".format(
                     requiredVal=checking_args[arg]["required"]
                 )
+            if arg == "tips_location":
```
pringshia, line 157: should we also add check against the blueprint name too? trying to think in the off-chance someone else names a field in a separate blueprint the same name as this

Etesam913: It would make it much more verbose, because every blueprint uses a tips_location. It also kind of goes against the idea of the docs updating by themselves as you would have to go and add a blueprint name.

pringshia: Fair point

pyproject.toml
```diff
@@ -67,6 +67,7 @@ click_default_group = "^1.2"
 tabulate = "^0.8.7"
 rich = "^12.5.1"
 rich-click = "^1.5.1"
+mdutils = "^1.4.0"
```
JackUrb, line 70: Nit: this is a `dev` dependency, not a `cli` dependency.

JackUrb (approved):
This is great! It'll be really useful for the docs site to be a definitive location for all of this info :)

The abstraction for `get_wut_arguments` in particular is a great decision, and certainly makes the rest of these scripts easier to produce.

PR #875, opened 2022-08-01
Fixing Worker Pool race condition for assignments

JackUrb:
# Overview
An awkward race condition existed that allowed Mephisto to attempt to launch an `Assignment` before the `WorkerPool` had actually registered all of the `Agent`s for that `Assignment` into its local `self.agents` dictionary. This could occur if the last `_assign_unit_to_agent` thread for an Assignment ran during the `await`-wrapped `get_init_data_for_agent` call of the second to last `_assign_unit_to_agent` for the same `Assignment`. Then the second to last thread would resume during the last call's `get_init_data_for_agent`. The outcome is that `loop.run_in_executor(None, assignment.get_agents)` would return all the `Agent`s, but one wouldn't yet be added to the `WorkerPool`'s tracking.

# Fix Details
Rather than just using the `None` check for the `Assignment.get_agents()` call, we do a `get` check for the `WorkerPool`'s agents as well, then bail early if any haven't been set.

# Testing:
Automated testing

Would be nice if there was an `asyncio` testing library that intentionally took unusual routes (like by adding `asyncio.sleep` of random durations to all `await` calls) to catch this kind of thing sooner...


chiehminwei (approved):
Thanks so much for the fix! LGTM

PR #878, opened 2022-08-01
ParlAI Assignment Fixes pt 2

JackUrb:
Following #875, there's now the condition where too many assignments are being launched. While usually harmless to have this redundancy, it introduced a new race conditions where two threads could be launched for a single assignment. Having two assignment threads for the same assignment, being not ideal, is now patched.

Also bumps hydra in-line with the new expectations as of #869.


mephisto/operations/worker_pool.py
```diff
@@ -265,26 +265,30 @@ async def _assign_unit_to_agent(
...
+
+                # See if the concurrent assignment is ready to launch
                 logger.debug(f"Attempting to launch {assignment}.")
                 agents = await loop.run_in_executor(None, assignment.get_agents)
                 if None in agents:
-                    agent.update_status(AgentState.STATUS_WAITING)
                     return  # need to wait for all agents to be here to launch
 
+                for queried_agent in agents:
+                    if queried_agent.get_status() != AgentState.STATUS_WAITING:
```
chiehminwei, line 280: Under what condition would this not be true? When things haven't synced to the worker threads?

chiehminwei (approved):

mephisto/operations/worker_pool.py
JackUrb: When two workers connect at the same time, and both have had an `_assign_unit_to_agent` thread called, and have thus both had `Agent`s created, but only one of those threads has made it to the `update_status` call.

PR #879, opened 2022-08-01
Added ability to customize `html-static-task-example` submit button

### LINKED ISSUE #692: Make the Submit button for the static_html_task blueprint customizable
The Submit button for the static html task is rendered from within the blueprint. This makes it difficult for users to control, e.g. in cases where they may want to disable the button.

Although we do recommend folks to use React, I think to help transition users with existing projects into using Mephisto, we should support use cases such as this.

https://github.com/facebookresearch/Mephisto/blob/07791f2afed3d9eebc5624785a01be911370d484/mephisto/abstractions/blueprints/static_html_task/source/dev/app.jsx#L122

### Recommendations:
- Adding an in-line comment to the demo onboarding file in the examples folder explaining why the Submit button isn't shown there and how to "access" it for customization. This is a source of confusion initially for users.
  - One workaround at the moment is to do something like `document.querySelector("button[type=submit]").disabled = true;`. At the least, we should add an HTML ID to the submit button to make it easier to target
- Allow the default submit button to be "hidden" so users can provide their own implementation. Perhaps configurable via setting a global variable, a la `window.HIDE_MEPHISTO_SUBMIT_BUTTON = true`. Or perhaps via Hydra config. @JackUrb any recommendations/preferences?

cc/ @edwardguo61

Etesam913:
## Overview
#692 mentions giving the user the ability to disable the submit button in the `html-static-task-example`

This is now possible by writing the following in the `demo_task.html`
```html
<script>
  window._MEPHISTO_CONFIG_.set("HIDE_SUBMIT_BUTTON", true)
</script>
```

The `set(property, value)` method sets the property to the value in the `window._MEPHISTO_CONFIG` object and also uses the newly created event emitter to emit an event of name `property` and argument `value`.

This is the event emitter(non-transpiled and non-minified version) that is defined in the `wrap_crowd_source.js` files

```js
let eventEmitter = () => ({
  events: {},
  emit(event, ...args) {
    let callbacks = this.events[event] || [];
    for (let i = 0, length = callbacks.length; i < length; i++) {
      callbacks[i](...args);
    }
  },
  on(event, cb) {
    this.events[event]?.push(cb) || (this.events[event] = [cb]);
    return () => {
      this.events[event] = this.events[event]?.filter((i) => cb !== i);
    };
  },
});
```

Events can be listened to and make a state update using the `useMephistoGlobalConfig` hook. 

For example, in the simple_static_task's `app.jsx`:

```jsx
  const [
    isSubmitButtonHidden,
    setIsSubmitButtonHidden,
  ] = useMephistoGlobalConfig(
    "HIDE_SUBMIT_BUTTON",
    false,
    (val) => typeof val === "boolean"
  );
```

This code snippet updates the `isSubmitButtonHidden` state to whatever `HIDE_SUBMIT_BUTTON` was set to in `demo_task.html`. 

The second param(optional) value, `false`, is the default value of `isSubmitButtonHidden`. This could be useful when the event is emitted after some action in the future. In this case it is emitted instantly at the beginning so it is not that useful.

The last param(optional) value, is a function that has to return true for the state to be updated. In this case, the function ensures that  `HIDE_SUBMIT_BUTTON` was set to a boolean in `demo_task.html` before updating the state.

## Resolves #692

https://user-images.githubusercontent.com/55665282/182468561-7c66abfa-4f9a-478a-9d6e-7c66161fbf70.mov


mephisto/abstractions/providers/mock/wrap_crowd_source.js
```diff
@@ -45,6 +45,12 @@ function handleSubmitToProvider(task_data) {
   return true;
 }
 
+window.HIDE_SUBMIT_BUTTON = (isHidden) => {
```
pringshia, line 48: Can you add this property under the `_MEPHISTO_CONFIG_` variable? We added that to avoid polluting the global namespace over time

mephisto/abstractions/providers/mock/wrap_crowd_source.js
```diff
@@ -45,6 +45,12 @@ function handleSubmitToProvider(task_data) {
...
+  const submitButton = document.getElementById("html-task-submit-button");
+  if (isHidden) submitButton.style.display = "none";
+  else submitButton.style.display = "block";
```
pringshia, line 51: Hmm will this actually work? If we modify the element like this outside of React, then if React ever re-renders the above component, there's no guarantee that this style stays, and will actually most likely be wiped out. Sounds like this could lead to a really confusing bug down the road.

pringshia (commented):
I'm not sure this solution actually works. I explained why [here](https://github.com/facebookresearch/Mephisto/pull/879#discussion_r935536457).

What I imagine would work instead:

1. We expose setter and getter methods for modifying the global `_MEPHISTO_CONFIG_` object, e.g. `_MEPHISTO_CONFIG_.get(x)` and `_MEPHISTO_CONFIG_.set(x, y)`
2. We expose an event emitter type object, that we can subscribe to events. `_MEPHISTO_CONFIG_.subscribe(handler)` and `_MEPHISTO_CONFIG_.unsubscribe(handle)` where handler looks something like `(propName, val) => {...}`
3. The base blueprint React component subscribes to the handler to update the disabled state:

```jsx
const [isHidden, setIsHidden] = React.useState(false);

const handler = React.useCallback((propName, val) => {
  if (propName === "HIDE_SUBMIT_BUTTON")
     setIsHidden(val);
}, [setIsHidden]);

React.useEffect(() => {
  window._MEPHISTO_CONFIG_.subscribe(handler)
  return () => window._MEPHISTO_CONFIG_.unsubscribe(handler);
}, [handler])

// later...

 {isHidden ? null : <Button
    type="submit"
    disabled={disabled}
  >}
```

Etesam913:
That definitely seems like a better solution, I forgot that there was a `app.jsx` file for some reason

mephisto/abstractions/blueprints/static_html_task/source/dev/app.jsx
```diff
@@ -98,11 +98,23 @@ function MainApp() {
...
 
   React.useEffect(() => {
     // Reset submitting when switching from onboarding
     setSubmitting(false);
-  }, [currentTask]);
+  }, []);
+
+  React.useEffect(() => {
+    window._MEPHISTO_CONFIG_.EVENT_EMITTER.on("HIDE_SUBMIT_BUTTON", handler);
+  }, [setIsHidden]);
```
Etesam913, line 117: Runs the hideSubmitButton handler method when the HIDE_SUBMIT_BUTTON event is emitted.

mephisto/abstractions/providers/mock/wrap_crowd_source.js
```diff
@@ -15,6 +15,22 @@ Returning None for the assignment_id means that the task is being
...
+    }
+  },
+  on(event, cb) {
+    this.events[event]?.push(cb) || (this.events[event] = [cb]);
+    return () => {
+      this.events[event] = this.events[event]?.filter((i) => cb !== i);
+    };
+  },
+});
+
```
Etesam913, line 33: The small event emitter class

mephisto/abstractions/providers/mock/wrap_crowd_source.js
```diff
@@ -45,8 +61,29 @@ function handleSubmitToProvider(task_data) {
...
+window._MEPHISTO_CONFIG_.get = (property) => {
+  if (!(property in window._MEPHISTO_CONFIG_))
+    throw new Error(`${property} does not exist in window.MEPHISTO_CONFIG`);
+  else return window._MEPHISTO_CONFIG_[property];
+};
+
+window._MEPHISTO_CONFIG_.set = (property, value) => {
+  window._MEPHISTO_CONFIG_[property] = value;
+  events.emit(property, value);
+};
```
Etesam913, line 84: getter and setter methods for the window._MEPHISTO_CONFIG_ object

mephisto/abstractions/providers/mturk/wrap_crowd_source.js
```diff
@@ -15,6 +15,22 @@ Returning None for the assignment_id means that the task is being
...
+let eventEmitter = () => ({
+  events: {},
+  emit(event, ...args) {
+    let callbacks = this.events[event] || [];
+    for (let i = 0, length = callbacks.length; i < length; i++) {
+      callbacks[i](...args);
+    }
+  },
+  on(event, cb) {
+    this.events[event]?.push(cb) || (this.events[event] = [cb]);
```
pringshia, line 27: just realized something. this file gets loaded in directly into the HTML page that runs on the client's browser: https://github.com/facebookresearch/Mephisto/blob/210e407e0683ce4cef273325746d07cc82d62fc4/examples/static_react_task_with_tips/webapp/src/static/index.html#L22

As such, we maaaay run into some slight compat issues with "newer" ES6 syntax such as the spread syntax and optional chaining "?."

- https://caniuse.com/mdn-javascript_operators_optional_chaining
- https://caniuse.com/mdn-javascript_operators_spread

I would suggest using the [Babel transpiler](https://babeljs.io/repl#?browsers=defaults&build=&builtIns=false&corejs=false&spec=false&loose=false&code_lz=DYUwLgBCBuIHZgKIFsCWYwgE4QLwQAoBKPAPkIG8AoCKWBAZwC4IKBfAGhqjTAJnhgOEAHRiAhlgDmDEtVq1QkAMbjgwAEbjlAawZ4IYABaoGIgYwDaFsAF0IAHwcRLtgNzdaAMwD2OAkoQqAYADMKgcFLGBqrqWrpmEVFGbkEQADwQScapqADUeXKeCrGa2nqWqLYEYiKSMkQeChBs3JzcPnD89EIQyhpFzcam5j0M1j22APwiAA4ArgxGBP0kToTDZjbjNvb4lv22jcVY4PNYcIQkuOTyzYYmW2MTgnsPI9svCNMiXqjAmCwBAIqGu5H6EAAhLh8KCmgo2PD2mxGkA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=env%2Creact%2Cstage-3&prettier=false&targets=Node-10.2&version=7.18.10&externalPlugins=&assumptions=%7B%7D) to make sure we use a `target` with more coverage. I suggest playing around with the `targets` value in the sidebar here:

CleanShot 2022-08-03 at 10 35 48@2x

pringshia: Let's also then minify the transpiled output above into maybe a single line of compressed output and then add that to the file. We can then leave a comment above it explaining our process for generating it nanoevents lib -> transpiler -> minifer

mephisto/abstractions/blueprints/static_html_task/source/dev/app.jsx
```diff
@@ -98,11 +98,26 @@ function MainApp() {
...
     setSubmitting(false);
-  }, [currentTask]);
+  }, []);
+
+  React.useEffect(() => {
+    window._MEPHISTO_CONFIG_.EVENT_EMITTER.on(
+      "HIDE_SUBMIT_BUTTON",
+      handleHideSubmitButtonEvent
+    );
+  }, [setIsSubmitButtonHidden]);
```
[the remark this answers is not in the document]
Etesam913: I presume you mean the React.useState hook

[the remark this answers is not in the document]
pringshia: oops yes, updated

mephisto/abstractions/blueprints/static_html_task/source/dev/hooks.js
```diff
@@ -0,0 +1,33 @@
+import { useState, useCallback, useEffect } from "react";
+
+/**
+ * This hook is to be used with events that are supposed to update state when they are consumed.
+ * @param {string} eventName
```
pringshia, line 5: small nit: I would rename `eventName` to `configName`.

I consider `HIDE_SUBMIT_BUTTON` to be a config property, not an event

mephisto/abstractions/blueprints/static_html_task/source/dev/hooks.js
```diff
@@ -0,0 +1,33 @@
...
+ * @param {any} defaultValue
+ * @param {function} validatorFunction
+ * @returns [any any]
+ */
+export function useMephistoGlobalConfig(
```
pringshia, line 10: this method looks great implementation-wise! Thanks for the docs as well :)

mephisto/abstractions/blueprints/static_html_task/source/dev/app.jsx
```diff
@@ -110,11 +111,19 @@ function MainApp() {
...
+    "HIDE_SUBMIT_BUTTON",
+    false,
+    (val) => typeof val === "boolean"
+  );
 
   React.useEffect(() => {
     // Reset submitting when switching from onboarding
     setSubmitting(false);
-  }, [currentTask]);
+  }, []);
```
pringshia, line 126: what was the rationale behind removing this?

Etesam913: that's a mistake, I don't even know how that got updated.

mephisto/abstractions/providers/mock/wrap_crowd_source.js
```diff
@@ -15,6 +15,11 @@ Returning None for the assignment_id means that the task is being
 previewed by the given worker.
 \------------------------------------------*/
 
+// The following is the nanoevents npm library (link) manually processed as such:
+// 1. transpiled to support more browser targets using the Babel transpiler (link), and
+// 2. minified using a JS minifier (link)
```
pringshia, line 20: `(link)` from my comment meant to actually replace that with the link (you can grab the links from my comment as well.

The babel one you might want to clean up a bit and remove the params otherwise it's really long and also includes the source code in the url

mephisto/abstractions/providers/mturk/wrap_crowd_source.js
```diff
@@ -15,6 +15,11 @@ Returning None for the assignment_id means that the task is being
 previewed by the given worker.
 \------------------------------------------*/
 
+// The following is the nanoevents npm library (link) manually processed as such:
+// 1. transpiled to support more browser targets using the Babel transpiler (link), and
+// 2. minified using a JS minifier (link)
```
pringshia, line 20: replace link

mephisto/abstractions/blueprints/static_html_task/source/dev/hooks.js
```diff
@@ -0,0 +1,33 @@
...
+) {
+  const [configState, setConfigState] = useState(defaultValue ?? false);
+
+  const handleEvent = useCallback(
+    (eventValue) => {
+      if (validatorFunction) {
+        if (validatorFunction(eventValue)) {
+          setConfigState(eventValue);
+        }
+      } else setConfigState(eventValue);
```
pringshia, line 23: both conditional paths are calling `setConfigState` here...?

pringshia: nvm, i see one is for the outer branch

pringshia (approved):
Looking excellent! Thanks for the updated video in the PR description as well, that demoes this.

It would be good to add documentation for this feature here: https://github.com/facebookresearch/Mephisto/tree/main/mephisto/abstractions/blueprints/static_html_task

In the future it would be nice to add a README to the folder here:  `https://github.com/facebookresearch/Mephisto/tree/main/examples/simple_static_task`

where we can also mention how the blueprint supports this functionality of hiding the button, and sample code to use this feature.

PR #881, opened 2022-08-03
Add cypress test to simple static task

Etesam913:
## Overview
Recently there was a bug with this task that made it into the `main` branch, so that motivates the need for these tests.

The test checks if 
* a request to register the agent is successfully made.
* The correct DOM elements are present on the page
* If the starting modal can be closed
* If the character name can be matched to a description using the select tag
* If a file can be chosen from the file input(the test uses the file named bliss.png in the `cypress/fixtures/bliss.png` directory
* If the submitted task shows an alert with the correct info
  * There was a bug where the alert that shows after submission would have `file1: {}` even if a file was submitted.
  * This happens because `JSON.stringify(File)` doesn't do what you expect it to. 
  * I just created a custom obj that included the file fields to bypass this.

This Cypress test is added to the GitHub action with the appropriate path filtering as seen here: https://github.com/facebookresearch/Mephisto/runs/7655824377?check_suite_focus=true.

## Video:
https://user-images.githubusercontent.com/55665282/182653810-1b519cc0-a671-47c6-9a99-2d9a45a7197d.mov


JackUrb (approved):
Really happy to have something in the test flow that also checks our file upload system. Thanks for adding this!

.github/workflows/cypress-end-to-end-tests.yml
```diff
@@ -52,6 +55,51 @@ jobs:
             mephisto-worker-addons:
               - 'packages/mephisto-worker-addons/src/**'
 
+  # Learn more about this test here: https://github.com/facebookresearch/Mephisto/pull/881
```
pringshia, line 58: love this inline pointer, thanks

mephisto/abstractions/blueprints/static_html_task/source/cypress/e2e/simple_static_task.cy.js
```diff
@@ -0,0 +1,53 @@
...
+        mimeType: "image/png",
+      });
+    });
+  });
+  it("Submit the task", () => {
+    cy.on("window:alert", (txt) => {
+      expect(txt).to.contains(
+        'The task has been submitted! Data: {"rating":"bad"'
+      );
+      expect(txt).to.contains('"name":"bliss.png"');
```
pringshia, line 43: this picture brings back memories

PR #883, opened 2022-08-03
Add Screening Example

Etesam913:
### This pr configures the option to run screening units on `static_react_task`. 
* Created a new `screening_example.yaml` configuration
* Fixed many bugs with adding screening units to `SharedStaticTaskState`
* Added video and extra information to the documentation page on Screening Units

## Mnist Video (this is the showcase in the documentation page below)
https://user-images.githubusercontent.com/55665282/183139879-d252d899-454c-4c15-afaa-474e6f02812b.mp4

## Static React Task Video
https://user-images.githubusercontent.com/55665282/183100782-4c6610f5-85c8-44e2-b056-db9ec9a80fa5.mp4

## New Documentation Page for Screening Unit
new_screening_docs_page


Etesam913:
What are your thoughts on putting a video in a documentation page? I compressed down the video, so it is about 9.4mb right now.

I do think that it would help a newcomer quickly understand how screening units work.

Etesam913:
I updated the video

docs/web/package.json
```diff
@@ -25,6 +25,7 @@
     "prism-react-renderer": "^1.2.1",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
+    "react-player": "^2.10.1",
```
Etesam913, line 28: This is for being able to play videos in mdx file.

examples/static_react_task/run_task.py
```diff
@@ -4,27 +4,76 @@
...
+            validate_onboarding=onboarding_always_valid,
+            on_unit_submitted=ScreenTaskRequired.create_validation_function(
+                cfg.mephisto,
+                validate_screening_unit,
+            ),
+            screening_data_factory=my_screening_unit_generator(),
+        )
+        shared_state.qualifications += ScreenTaskRequired.get_mixin_qualifications(
+            cfg.mephisto, shared_state
+        )
```
Etesam913, line 76: If you are using the screening unit conf they are more properties that you need to add to your `SharedStaticTaskState`.

examples/static_react_task/webapp/src/components/core_components.jsx
```diff
@@ -16,12 +16,21 @@ function OnboardingComponent({ onSubmit }) {
...
       >
-        Move to main task.
-      </button>
+        <button
+          className="button is-link"
+          onClick={() => onSubmit({ success: true })}
+        >
+          Move to Main Task
+        </button>
+      </div>
```
Etesam913, line 33: I was peeved by the fact that the onboarding button was not centered on the page.

examples/static_react_task/run_task.py
JackUrb: You can add these dynamically to the `SharedStaticTaskState` after the case:
```
share_state = ...
if is_using_screening_units:
    shared_state.on_unit_submitted = ...
    ...

```
To reduce duplication a bit

examples/static_react_task/webapp/src/components/core_components.jsx
JackUrb: Yeeeeaaah this could've been adjusted long ago hahaha

mephisto/abstractions/blueprint.py
```diff
@@ -8,6 +8,7 @@
 import csv
 from genericpath import exists
 import os
+from rich import print
```
JackUrb, line 11: Unneeded import

mephisto/operations/worker_pool.py
```diff
@@ -470,6 +470,7 @@ async def register_agent(
             function="get_valid_units_for_worker"
         ).time():
             units = task_run.get_valid_units_for_worker(worker)
+
```
JackUrb, line 473: Odd whitespace changes in this file

JackUrb (approved):
I love the new docs page and examples! Only a tiny nit on code duplication, but otherwise I'm excited to have solid examples of this workflow now, and for it to be wrapped into the two most commonly used blueprints.

pringshia:
> Fixed many bugs with adding screening units to SharedStaticTaskState

Curious to hear more about this, but thanks for fixing!

examples/remote_procedure/mnist/run_task.py
```diff
@@ -30,12 +34,29 @@
...
+        annotation = data["final_submission"]["annotations"][0]
+        if annotation["isCorrect"] and annotation["currentAnnotation"] == 3:
+            return True
+    return False
 
 
 @task_script(default_config_file="launch_with_local")
 def main(operator: Operator, cfg: DictConfig) -> None:
-    tasks: List[Dict[str, Any]] = [{}] * cfg.num_tasks
+    tasks: List[Dict[str, Any]] = [{"isScreeningUnit": False}] * cfg.num_tasks
```
pringshia, line 57: this could also be reverted back to what it was and the frontend would still work since it's doing a boolean check, right?

Etesam913:
There were some bugs with hydra conf variables not existing. The changes made to `mephisto/abstractions/blueprint.py` fixed this. @JackUrb knows more about this.

JackUrb:
> > Fixed many bugs with adding screening units to SharedStaticTaskState
> 
> Curious to hear more about this, but thanks for fixing!

This was my bad - I had made a mistake in the Mephisto magic that extracts the `BlueprintMixin`s from `Blueprint`s that inherit them to run initialization and other methods.

The issue: I wanted to be able to only take the _terminal_ class of a `BlueprintMixin`. Say someone defines `X` and someone else makes `YExtendsX` and your `Blueprint` uses `YExtendsX`, when we collect subclasses via `mro()` we want to call the init of `YExtendsX` and _not_ that of `X`. The way I did this before was amass the set of all subclasses of the `Blueprint`, then **drop** any class from that set that was a superclass of any other class in there. Unfortunately, when I amassed this list before, it didn't distinguish between `Blueprint`s and `BlueprintMixin`s explicitly. Thus, since static blueprints extend the abstract `StaticBlueprint`, the resolver would do the following (abridged):
1. `mro()` -> `[StaticReactBlueprint, StaticBlueprint, OnboardingRequired, ScreenTaskRequired, Blueprint, BlueprintMixin]`
2. Remove self and base classes -> `[StaticBlueprint, OnboardingRequired, ScreenTaskRequired]`
3. Remove subclasses of any other class in the set -> `[StaticBlueprint]`
The fix was to do a step 2.5 that drops any subclasses of `Blueprint` from the set first.

examples/remote_procedure/mnist/run_task.py
```diff
@@ -58,12 +79,24 @@ def handle_with_model(
...
+            function_registry=function_registry,
+            on_unit_submitted=ScreenTaskRequired.create_validation_function(
+                cfg.mephisto,
+                validate_screening_unit,
+            ),
+            screening_data_factory=my_screening_unit_generator(),
+        )
+        shared_state.qualifications += ScreenTaskRequired.get_mixin_qualifications(
+            cfg.mephisto, shared_state
+        )
```
pringshia, line 94: this is unrelated to this PR, but it seems that `ScreenTaskRequired.get_mixin_qualifications` is not even using `shared_state` despite requiring it in the signature.

examples/remote_procedure/mnist/run_task.py
```diff
@@ -58,12 +79,24 @@ def handle_with_model(
...
+    else:
+        shared_state = SharedRemoteProcedureTaskState(
+            static_task_data=tasks,
+            function_registry=function_registry,
+        )
```
pringshia, line 99: this could also be condensed based on Jack's earlier comment

pringshia (approved):
Pretty nice! Thanks for improving the screening developer experience. It's an often-used feature of Mephisto and helps greatly in assuring data quality.

examples/remote_procedure/mnist/run_task.py
JackUrb: `get_mixin_requirements` has a shared signature for all `BlueprintMixin`s, as ideally this will be replaced with `<>Blueprint.get_mixin_qualifications` automatically pulling required qualifications into the state _under the hood_.

examples/remote_procedure/mnist/run_task.py
Etesam913: It's being used like this: `const NUM_ANNOTATIONS = taskData.isScreeningUnit ? 1 : 3; `

PR #885, opened 2022-08-08
Add info on `link_task_source` to docs

Etesam913:
## Overview
For local development using the `link_task_source` property can significantly improve productivity. However, on the docs, there is not much information available on it. 

When searching "link_task_source" on the docs I only get two results, which both come from the blueprint table.

search-results

Ideally there should be some info on this property in the tutorial.

### Changes:
* Updated tutorial to include information on `link_task_source`
* Added a logging message that prints to the console if you have `link_task_source` set to false.

## Updated documentation section in "working on a custom task" page

launching_the_task

## Logging message that shows when `link_task_source` is set to false

link_task_source_logging

### DIFF
--- a/mephisto/operations/task_launcher.py
+++ b/mephisto/operations/task_launcher.py
@@ -26,7 +26,7 @@ import enum
 if TYPE_CHECKING:
     from mephisto.data_model.task_run import TaskRun
     from mephisto.abstractions.database import MephistoDB
-
+from omegaconf import DictConfig, OmegaConf
 import threading
 from mephisto.utils.logger_core import get_logger
 import types
@@ -180,7 +180,7 @@ class TaskLauncher:
             if not self.unlaunched_units:
                 break
 
-    def _launch_limited_units(self, url: str) -> None:
+    def _launch_limited_units(self, url: str, run_config: Optional[DictConfig]) -> None:
         """use units' generator to launch limited number of units according to (max_num_concurrent_units)"""
         # Continue launching if we haven't pulled the plug, so long as there are currently
         # units to launch, or more may come in the future.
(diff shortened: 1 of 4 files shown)


mephisto/operations/task_launcher.py
```diff
@@ -191,15 +191,24 @@ def _launch_limited_units(self, url: str) -> None:
                 if unit is None:
                     break
                 unit.launch(url)
+            if (
+                run_config is not None
+                and run_config.blueprint.link_task_source == False
+            ):
+                logger.info(
+                    "If you want your server to update on reload whenever you make changes to your webapp, then make sure to set \n\nlink_task_source: true\n\nin your task's hydra configuration and run \n\ncd webapp && npm run dev:watch\n\nin a separate terminal window. For more information check out:\nhttps://mephisto.ai/docs/guides/tutorials/custom_react/#12-launching-the-task"
+                )
```
pringshia, line 200: What is the rationale for putting the checking code here as opposed to somewhere else? Not challenging, just curious. Is there any other place this could be placed up the function call stack?

I find that adding a new param to the function just for this functionality to be a bit of a code smell, though I'm willing to be talked out of it.

Etesam913: It seems to be messing up the python build test so I will probably have to find a new place to put it. 

I put it there so that I could get access to the run_config variables. I need to check if link_task is set to false before showing anything.

pringshia: e.g. why not put it in `launch_task_run_or_die`?

Etesam913: Yeah, this seems like a way better solution. This will also prevent the tests from failing.

JackUrb (approved):
LGTM!

Issue #790, opened 2022-06-10
Utility to reset worker's onboarding status


xksteven:
I tried to find the answer but the link in [architecture overview] is broken(https://github.com/facebookresearch/Mephisto/blob/main/packages/mephisto-task/README.md#agentid).  Not sure where that info would be in the documentation.

I've been debugging locally and I've been testing out that I can fail the onboarding task but I don't know how to reset the default user. 

Issue #748 is a slight duplicate but refers to another repo namely [ParlAI](https://github.com/facebookresearch/ParlAI) for their instructions.  It's not clear how to do this in general in Mephisto.

JackUrb:
Hi @xksteven - thanks for the question! This would likely be a good candidate to add as a util to the `OnboardingRequired` mixin (`OnboardingRequired.clear_onboarding(worker, qualification_name)`), and then adding to the [onboarding docs](https://mephisto.ai/docs/guides/how_to_use/worker_quality/using_onboarding/). I'm currently on vacation, but in the meantime you can see the qualifications associated with onboarding here:
https://github.com/facebookresearch/Mephisto/blob/a224a180142604985c559f27663cb191601ec273/mephisto/operations/worker_pool.py#L316-L323

With this info you can use `revoke_qualification` on both of the related qualifications to clear an individual worker. 

For other possible solutions, if you're just clearing for yourself in local for testing, you can simply change your own worker id. If you want to 'reset' all workers for a task, you can just change the `onboarding_qualification`.

PR #886, opened 2022-08-08
Adds ability to clear onboarding for a worker

Etesam913:
## Overview
For development purposes it can be convenient to reset a worker's onboarding qualifications. Now it is possible!

### Changes:
* Added `clear_onboarding(self, worker: "Worker", qualification_name: str):` to `onboarding_required.py` as a class method.
* Added a `clear_worker_onboarding.py` script that can used to clear a worker's onboarding qualifications.
  * This script can be ran by typing `mephisto scripts local_db clear_worker_onboarding` in the console.

## Video:

https://user-images.githubusercontent.com/55665282/183500440-669f8f3c-3dd2-44a1-aa4b-fdd60275e542.mp4

This video shows how the script can be used to clear a worker's onboarding. Worker x gets the onboarding qualification granted when he presses the button. After the script is ran, the worker is taken back to the onboarding page as the script revoked the onboarding qualification.

## Resolves https://github.com/facebookresearch/Mephisto/issues/790

### DIFF
--- a/mephisto/abstractions/blueprints/mixins/onboarding_required.py
+++ b/mephisto/abstractions/blueprints/mixins/onboarding_required.py
@@ -147,6 +147,12 @@ class OnboardingRequired(BlueprintMixin):
             db, self.onboarding_failed_name
         )
 
+    @classmethod
+    def clear_onboarding(self, worker: "Worker", qualification_name: str):
+        worker.revoke_qualification(qualification_name)
+        if qualification_name is not None:
+            worker.revoke_crowd_qualification(self.get_failed_qual(qualification_name))
+
     def get_onboarding_data(self, worker_id: str) -> Dict[str, Any]:
         """
         If the onboarding task on the frontend requires any specialized data, the blueprint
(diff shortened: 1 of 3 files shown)


JackUrb (approved):
I love this feature!

The one thing I'd suggest though is that people using this may also want to be able to change/query by `worker_name` rather than just by the id, as they may know the former and not the latter. Still, this is a great feature to have either way.

Etesam913:
I changed it to query by worker name.

I added a large comment in the script as what I am doing may be quite confusing.

pringshia (approved):

pringshia:
Looks fine to me, will let Jack chime in on the worker_name comment/change

JackUrb (approved):
LGTM! Agreed that the distinction between mephisto worker ids and worker names (which are usually the external IDs for a worker) is confusing. Unsure how to improve this at the moment.

Etesam913:
The simplest solution to fix this would probably to rename "worker_id" to "worker_name" in the url, but that can be done in a separate pr.

Issue #829, opened 2022-07-15
Properly time `mephisto-task` out on "Initializing...."


JackUrb:
There are many reports of workers connecting to a task only to be stuck on "Initializing" forever. This occurs when the Mephisto routing server is set up, but the backend is not connected. 

We should provide better error messaging to the workers when this occurs, and ideally get a message to the task owner as it indicates improper shutdown or a connectivity issue.

pringshia:
On the initial handshake, when getting `agentId`, the handshake goes to the router, but doesn't reach the router, we hang forever for a response.

Two error cases:
- Router is aware that the server is down (no hearbeat after X mins), all subsequent messages result in 500x-esque error
- If there actually was a timeout issue but the server and router are still up (either the front-end sends another request or user refreshes)

To repro:
- Start up a server, kill Mephisto with a pipe (Ctrl+\) and not with a Ctrl+C (this skips all the shutdown processes but the router is still running)

PR #888, opened 2022-08-09
Add feedback component to loading page

Etesam913:
## Overview
A very common circumstance where feedback might want to be received is during loading.

There are times where  a task my be stuck in loading forever. The researcher should be able to know directly if a large chunk of workers are not able to see the task.

This was brought up to me by @EricMichaelSmith 

### Changes:
* Feedback component shows after loading for 5 seconds (should this number be adjusted? not exactly sure about the average loading times).

## Video

https://user-images.githubusercontent.com/55665282/183749626-174a09e7-5a55-4f79-926a-6ae33cb5681f.mp4

### DIFF
--- a/examples/remote_procedure/mnist/webapp/package-lock.json
+++ b/examples/remote_procedure/mnist/webapp/package-lock.json
@@ -10,6 +10,7 @@
       "dependencies": {
         "bootstrap": "^4.6.0",
         "mephisto-task": "2.0.2",
+        "mephisto-worker-addons": "^1.0.0",
         "rc-slider": "^8.6.3",
         "react": "16.13.1",
         "react-bootstrap": "^1.6.0",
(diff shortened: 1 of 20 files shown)


Etesam913:
@EricMichaelSmith Is this kind of what you expected?

JackUrb:
This is cool! Only issue is that the _most common_ reason this happens is because the mephisto router is alive, but the Mephisto server that was associated with it no longer is connected. After #829 however then this would cover the actual loading use cases.

EricMichaelSmith:
> @EricMichaelSmith Is this kind of what you expected?

Yeah! This is really cool to see, and it'll let us get feedback about broken HITs. I feel like, in my experience, emails about broken HITs are at least as common as all other email topics put together...

pringshia:
This is a clever use case for the new Feedback component! I would suggest a 10 second delay, 5 seems a little too quick to me. It would also be nice to have the timeout value as a constant, and perhaps we can expose that at the top of the `main.jsx` file. (in the future other configurable constants could also be colocated there)

PR #892, opened 2022-08-30
Finding OSTYPE is not exists

mojtaba-komeili:
While trying to install mephisto metrics, ran into issues with empty `OSTYPE` env variable. This fixes that by checking it exists.


JackUrb (approved):
Thanks for getting to the bottom of this @mojtaba-komeili!

PR #893, opened 2022-08-30
Adding more general function with better name

mojtaba-komeili:
# Description
We have the `direct_soft_block_mturk_workers` function which is great to assign any qualification to workers, including the soft blocking qualification. The convenience of this often tempts people to use it for any general qualification ([example](https://github.com/fairinternal/ParlAI-Internal/blob/765a8b11678c9b900d73d888046b327a557d0c44/crowdsourcing/projects/light_multiparty/refresh_allowlist.py#L54)). This sort of use causes confusion and may end up with logical bugs. This patch renames `direct_soft_block_mturk_workers` to `direct_assign_qual_mturk_workers` (with minimal changes). To keep the backward compatibility, we still have `direct_soft_block_mturk_workers`; but it is just a wrapper around the call to `direct_assign_qual_mturk_workers`.

# Test
Used it with assigning allow list qualification in [this PR](https://github.com/fairinternal/ParlAI-Internal/pull/3646/files), and it worked well. Checking the final results proved that it assigned the qualification correctly.

### DIFF
--- a/.github/workflows/cypress-end-to-end-tests.yml
+++ b/.github/workflows/cypress-end-to-end-tests.yml
@@ -71,9 +71,9 @@ jobs:
         uses: actions/setup-python@v2
 
       - name: 🪨 Setup Node
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v3
         with:
-          node-version: 16
+          node-version: 16.16.0
 
       - name: 🤖 Install Mephisto
         run: pip install -e .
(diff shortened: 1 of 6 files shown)


mojtaba-komeili:
@JackUrb not sure what they cypress one is about. Could you give me a hint please?

JackUrb:
Hm something strange is happening with the build flow. I'll see if I can reproduce on the other dependency branch I'm working on.

JackUrb (approved):
Quality change!

As far as the testing goes, it seems that cypress tests are failing on main at the moment. Need to debug the why.

mojtaba-komeili:
> Quality change!
> 
> As far as the testing goes, it seems that cypress tests are failing on main at the moment. Need to debug the why.

I wait for the new changes you had to land and retry after rebasing with that.

PR #896, opened 2022-08-31
Add max_assignments_per_hit; MTurk's  MaxAssignments

thammegowda:
The default values set by the current version does not allow multiple workers to work on same HIT.
The proposed pull request adds `task.max_assignments_per_hit` config  (with default=1), and allows configuring the maximum assignments per HIT.

Related: 
* https://github.com/facebookresearch/ParlAI/issues/3932 
* [mturk API  `create_hit_with_hit_type`](https://boto3.amazonaws.com/v1/documentation/api/1.11.9/reference/services/mturk.html#MTurk.Client.create_hit_with_hit_type)

### DIFF
--- a/mephisto/abstractions/providers/mturk/mturk_unit.py
+++ b/mephisto/abstractions/providers/mturk/mturk_unit.py
@@ -274,10 +274,12 @@ class MTurkUnit(Unit):
     def launch(self, task_url: str) -> None:
         """Create this HIT on MTurk (making it available) and register the ids in the local db"""
         task_run = self.get_assignment().get_task_run()
-        duration = task_run.get_task_args().assignment_duration_in_seconds
+        task_args = task_run.get_task_args()
+        duration = task_args.assignment_duration_in_seconds
+        max_assignments_per_hit = task_args.max_assignments_per_hit
         task_lifetime_in_seconds = (
-            task_run.get_task_args().task_lifetime_in_seconds
-            if task_run.get_task_args().task_lifetime_in_seconds
+           task_args.task_lifetime_in_seconds
+            if task_args.task_lifetime_in_seconds
             else 60 * 60 * 24 * 31
         )
         run_id = task_run.db_id
@@ -292,9 +294,10 @@ class MTurkUnit(Unit):
             task_url,
             hit_type_id,
             lifetime_in_seconds=task_lifetime_in_seconds,
+            num_assignments=max_assignments_per_hit
         )
         # TODO(OWN) get this link to the mephisto frontend
-        print(hit_link)
+        print(f'{hit_link} \tHITId:{hit_id}')
 
         # We create a hit for this unit, but note that this unit may not
         # necessarily match with the same HIT that was launched for it.
--- a/mephisto/data_model/task_run.py
+++ b/mephisto/data_model/task_run.py
@@ -98,12 +98,22 @@ class TaskRunArgs:
             )
         },
     )
+
     max_num_concurrent_units: int = field(
         default=0,
         metadata={
             "help": (
                 "Maximum units that will be released simultaneously, setting a limit "
                 "on concurrent connections to Mephisto overall. (0 is infinite)"
+            )
+        },
+    )
+    max_assignments_per_hit: int = field(
+        default=1,
+        metadata={
+            "help": (
+                "Maximum assignments per task. The number of times the task"
+                "can be accepted and completed"
             )
         },
     )


JackUrb:
Hi @thammegowda - thanks for reaching out with this fix. This functionality already exists on a `Blueprint`-specific basis though, as in Mephisto we do hit content assignment on our end rather than on the MTurk side. This allows us to get by without having to define MTurk-specific arguments (as MTurk isn't our only intended crowd provider).

Here's a [section of the tutorial](https://mephisto.ai/docs/guides/tutorials/first_task/#34-unit-creation-explained) that explains this functionality.

thammegowda:
@JackUrb Thanks for the reply.  
I agree we should try avoid args specific to mturk crowdsourcing backend. 

BTW, that tutorial is for static blueprint (`units_per_assignment`  is defined in `StaticBlueprintArgs`).  But my system is based on parlai blueprint (see [examples/parlai_chat_task_demo](https://github.com/facebookresearch/Mephisto/tree/main/examples/parlai_chat_task_demo))  
and `units_per_assignment` support is not working for Parlai.  
Even after setting MaxAssignments to higher value on MTurk API, mephisto is showing `There is currently no available work, please try again later` to mturk workers. Any hints/suggestions on how to support more than one units for parlai_chat_task_demo? thanks

JackUrb:
Interesting question here - if you're using the `ParlAIBlueprint` then inter-annotator agreement isn't what you're looking for. Is the goal to just make it so workers can do more than one task? This is controlled by `maximum_units_per_worker`. See `mephisto wut blueprint=parlai` and `mephisto wut task` for the complete list of controllable arguments. Let me know if I'm not understanding correctly what you're asking for here!

thammegowda:
@JackUrb Thanks for the reply. 
> Is the goal to just make it so workers can do more than one task? 

Our setup: We have a chatbot, we want to let (multiple) workers chat with the bot for a while (say $n \in [5,10]$ turns). We collect human evaluation at the end of $n$ turns (e.g., how coherent is chatbot). Multiple workers help in our case to obtain more reliable evaluation (say, average of scores).

Currently, the first worker to open is able to finish the assignment as expected. Then the other workers are getting `There is currently no available work, please try again later`. 

---
I am trying to debug this, this error seems to get triggered at this line
https://github.com/facebookresearch/Mephisto/blob/d8f5c3b0a8185d2f451cce2f31e6135b7fbc7615/mephisto/operations/worker_pool.py#L474

JackUrb:
If your goal is to collect multiple conversations, you'll need to tell Mephisto to launch multiple units (`mephisto.blueprint.num_converations`). This is [part of the `ParlAIChatBlueprint`](https://github.com/facebookresearch/Mephisto/blob/d8f5c3b0a8185d2f451cce2f31e6135b7fbc7615/mephisto/abstractions/blueprints/parlai_chat/parlai_chat_blueprint.py#L123-L128).

Perhaps I'm not following because we're not on the same page with what you mean by "work on same HIT". What makes the HIT the same that you want two workers to work on, the model you're connecting them to chat with?

thammegowda:
@JackUrb I will try `num_conversations` and see if it will solve my problem.  Thank you again for the reply 👍 
By same HIT, I mean the same model. We bootstrap/seed conversation with a few turns, so to be 100% precise, HIT is (model, seed).

jonmay:
@JackUrb as we understand it a HIT loads a seed conversation. We want to have several different people have a conversation using the same seed conversation. In MTurk terms that's several assignments per HIT. That's the value we seem to be unable to change.

JackUrb:
Hi @jonmay, sorry for the long delay here! This isn't fully supported in the `ParlAIChatBlueprint`, though admittedly it's not a far leap to implement. 

You can already provide your individual bits of context in [`SharedParlAITaskState.context_jsonl`](https://github.com/facebookresearch/Mephisto/blob/48c0c23f79f4ea0d94d500831b44b39c7d168b17/mephisto/abstractions/blueprints/parlai_chat/parlai_chat_blueprint.py#L119). Then, we determine whether we're launching `Unit`s (individual tasks per worker) or `Assignment`s based on how many humans you're specifying need to be present in a world: https://github.com/facebookresearch/Mephisto/blob/48c0c23f79f4ea0d94d500831b44b39c7d168b17/mephisto/abstractions/blueprints/parlai_chat/parlai_chat_task_runner.py#L132

So, for the fix, as long as `agent_count` is 1, then the goal is to have Mephisto launch more units for the same non-concurrent assignment here: https://github.com/facebookresearch/Mephisto/blob/48c0c23f79f4ea0d94d500831b44b39c7d168b17/mephisto/abstractions/blueprints/parlai_chat/parlai_chat_blueprint.py#L316

The difficulty is that we'd need to un-tether `agent_count` for the launch right here in the *specific* case of single-agent conversations that you want to have multiple copies of.

My suggestion is to add a the same `units_per_assignment` argument to the `ParlAIChatBlueprintArgs`, however **require** it be 1 if `get_world_params()['agent_count'] != 1` here: https://github.com/facebookresearch/Mephisto/blob/48c0c23f79f4ea0d94d500831b44b39c7d168b17/mephisto/abstractions/blueprints/parlai_chat/parlai_chat_blueprint.py#L231-L233. Then in `__init__`, `self.units_per_assignment` should be set to the max of it's parameter value and the value of `get_world_params()['agent_count']`. Then, in `get_initialization_data(self)` you can use `self.units_per_assignment` instead and everything should work as planned. 

Does this implementation make sense?

PR #897, opened 2022-08-31
Removing threadsafety as we already lock connections behind threads

JackUrb:
# Overview
Under rare circumstances, sqlite DB access was occurring across threads from where the table was originally created. Mephisto however wraps all DB access with conditions, so rather than having `sqlite` throw for us, we can actually safely process these requests.

# Testing
We had a long running job that had issues before with an access across threads that killed the run. Now that job appears stable with no DB issues.


mojtaba-komeili (approved):
LGTM

PR #899, opened 2022-09-07
updated react and react-dom

0x00b1:
Hi, @JackUrb! I needed to update react and react-dom for the compression [task](https://github.com/fairinternal/mephisto-internal/pull/21). Let me know if this works. 

cc: @mmuckley

PR #905, opened 2022-10-03
Update to React 18

### LINKED ISSUE #964: Cypress Testing is Broken
Seems frontend testing via Cypress is broken right now. he GH action is loading cypress 12.x.x, but the setup seems to work with only 10.x.x. Unsure how to pin the version as we already specify ^10.0.0 across most packages.

### LINKED ISSUE #991: Update react version in mephisto-task
Hi @JackUrb , hope you are doing well.

I'm trying to run a custom react task with Mephisto which is using https://github.com/react-dnd/react-dnd/ as a depency for drag and drop feature. That package requires React version `^16.14.0`. However, mephisto-task package is fixed with React `16.13.1` and I had react version conflict when I try to npm install `react-dnd`.

Therefore, I just wonder can we update `mephisto-task` package to accept newer react versions? I'm afraid that in the future we will use other packages that require latest react versions as well. Happy to hear if we have any alternative approaches 😄.

pringshia:
Fixes #964, fixes #991

## Got all tests passing! 🥳

Resuming work from #899 as a base and trying to get tests to pass.

Took almost all of an entire day but I think I got it. Key updates:

- **Published mephisto-task@2.0.4** which includes relaxed react peer deps
- **Published mephisto-worker-addons@1.0.2** which includes relaxed react peer deps
- **Published bootstrap-chat@2.0.2** which includes relaxed react peer deps
- Ran `./sync_package_version.sh mephisto-task 2.0.4`
- **Disabled storybook and \@annotated/\* packages from the workspace for now**, they're overcomplicating the build and I think it makes sense to move those types of UI components into their own repo anyway
- **Fixed failing Cypress tests**. After much hunting it seems that Cypress builds are failing due to v12 changes. Migration guide isn't clear as to why though. This PR pins CI version to v11.

---

Other notes:

- Downgraded \@mdx-js/react, causing issues with Docusaurus, see https://github.com/redhat-developer/odo/pull/6227
- Downgrade `react-bootstrap` since the components we use are missing in the newer version: https://stackoverflow.com/a/60101003
- Note: Several package.json files named `parlai-mturk-task-compiler` include an old version of the `react-table` lib @ version 6, which has react 16 max as the peer dep. Upgrading isn't straightforward I think since the library has undergone major changes in v7 (and is now at v8)


JackUrb (approved):
Exciting to be able to unblock! And also looking forward to the future annotated repo 👀 

> Several package.json files named parlai-mturk-task-compiler include an old version of the react-table lib @ version 6, which has react 16 max as the peer dep.

The only usage in Mephisto and the ParlAI libraries is in [this file](https://github.com/facebookresearch/Mephisto/blob/aa60850bfed9e1d5cb2cc3deadbc7b080728c3c3/mephisto/client/full/webapp/src/widgets/GridReview.js).

pringshia:
I'm inclined to deprecate that folder for now and in doing so also sunset the `react-table` dependency. That client UI was exploratory and hasn't been used for any meaningful projects since.

PR #909, opened 2022-10-04
Some fixes made to run a live demo of Mephisto functionality

JackUrb:
# Overview
As titled. Does the following 3 things:
- Introduces a frontend-based review flow for the MNIST demo, which looks pretty cool since you get to see the drawn images.
- Fixes preview for the `mnist` task, which was asking for `isScreeningUnit` during preview and breaking
- Updates the `examine_results.py` script to pull times from the correct location.

## Discussion:
It may be heavy-handed to have to commit the full react app when we know the components being replaced (the viewers). While having a flow that lets you create one from template is nice, it would also be nice from a "visible contribution" standpoint to have a Mephisto Review server that can pull in just these components. Unsure that kind of flow is at all attainable @pringshia.


pringshia:
@JackUrb - Totally understand your point here. It's somewhat possible using a batteries-loaded script as `react-scripts` is for create-react-app. However, maintaining/ supporting that comes with logistical overhead and release management costs. Also what makes `react-scripts` feasible is the `eject` functionality which lets users exit out when they run into limitations with the abstraction. Unfortunately that's not trivial to set up.

For now I tried my best to design the scaffolding so that all custom code is co-located, in this case in the `components/custom/` folder. You _could_ just check that in, with instruction on how to pull down the latest template and then drop-in the folder custom folder, updated the import link. It trades-off some setup work for a smaller repo-size... i'm undecided on what's a better option. I think just what you have right now is probably sufficient.

pringshia (approved):

pringshia:
> Introduces a frontend-based review flow for the MNIST demo, which looks pretty cool since you get to see the drawn images.

A screenshot or video of the review flow would be nice!

PR #910, opened 2022-10-04
Experimental functions for resuming a task run

JackUrb:
# Overview
This PR acts as a starting point for implementing the path of resuming incomplete or interrupted `TaskRun`s. It creates two new experimental functions (denoted with an `EXP_` prefix) that power the functionality. The basic flow is to:
1) Query for all assignments of that task run, and find ones with missing `Unit`s.
2) Load up all of the `Assignment`s, then force-reset their incomplete units from `EXPIRED` to `CREATED`, 
3) Clone configuration from the previous `TaskRun`, and use it to initialize a `CrowdProvider`, `Architect`, and `Blueprint`
4) Push the assignments through the `launch_unit`s flow (which would need to be updated to only launch units in a `CREATED` state and ignore all others).

# Implementation
The first portion is a version of `Operator.launch_task_run_or_die`, which instead of creating a new task run from a given config, loads a task run from ID and steals its config.  It then establishes the same type of local state that an operator would have normally and tries to execute the run.
The second part is `TaskLauncher.EXP_resume_assignments`, which replaces the `create_assignments` step and instead looks for existing assignments that have incomplete units. It moves those incomplete units (either by expiration or (+soft) rejection) to the CREATED state, and stores them locally so that the `launch_units` call from the `Operator` only relaunches these particular units.

# Testing
No testing yet

# TODO
- [ ] Do some testing in sandbox/local, discover some bugs, resolve those bugs
- [ ] Commit this to `main`: it's experimental. Some people can use it, but I'd like to standardize pre-releasing functionality that may unblock people (or lead them to contribute the remaining parts to bring to core functionality)

# TODO to move to core
- Add validation that a job _can_ be resumed. Perhaps we need metadata? Any task run with non-standard `SharedTaskState` for instance is going to struggle to resume properly, but it's possible we have a script wrapper that restores screening unit validator, onboarding functions, etc.
- Create tests
  - Create a simple automated test that actually runs this functionality
  - Ensure that nothing is broken with regards to shutting down after the new tasks are compeleted
  - Ensure that there's no weird linking if we go back and re-action other workers
  - Ensure that our review scripts still work on task runs that go through this flow
- Fix anything that was surfaced writing new tests
- Remove experimental prefix

### DIFF
--- a/mephisto/operations/operator.py
+++ b/mephisto/operations/operator.py
@@ -595,3 +595,82 @@ class Operator:
             )
         finally:
             self.shutdown()
+
+    def EXP_launch_from_incomplete_run(
+        self, task_run_id: str, shared_state: Optional[SharedTaskState] = None
+    ) -> str:
+        """
+        Attempt to resume a task run with incomplete assignments/units
+        """
+        task_run = TaskRun.get(db, task_run_id)
+        run_config = task_run.args
+
+        set_mephisto_log_level(level=run_config.get("log_level", "info"))
+
+        requester, provider_type = self._get_requester_and_provider_from_config(
+            run_config
+        )
+
+        # Next get the abstraction classes, and run validation
+        # before anything is actually created in the database
+        blueprint_type = run_config.blueprint._blueprint_type
+        architect_type = run_config.architect._architect_type
+        BlueprintClass = get_blueprint_from_type(blueprint_type)
+        ArchitectClass = get_architect_from_type(architect_type)
+        CrowdProviderClass = get_crowd_provider_from_type(provider_type)
+
+        if shared_state is None:
+            shared_state = BlueprintClass.SharedStateClass()
+
+        BlueprintClass.assert_task_args(run_config, shared_state)
+        ArchitectClass.assert_task_args(run_config, shared_state)
+        CrowdProviderClass.assert_task_args(run_config, shared_state)
+        # TODO special validation that _THIS TASK RUN_ can be resumed
+
+        # Make a new live run
+        live_run = self._create_live_task_run(
+            run_config,
+            shared_state,
+            task_run,
+            ArchitectClass,
+            BlueprintClass,
+            CrowdProviderClass,
+        )
+
+        try:
+            # If anything fails after here, we have to cleanup the architect
+            # Setup and deploy the server
+            built_dir = live_run.architect.prepare()
+            task_url = live_run.architect.deploy()
+
+            # TODO(#102) maybe the cleanup (destruction of the server configuration?) should only
+            # happen after everything has already been reviewed, this way it's possible to
+            # retrieve the exact build directory to review a task for real
+            live_run.architect.cleanup()
+
+            # Register the task with the provider
+            live_run.provider.setup_resources_for_task_run(
+                task_run, run_config, shared_state, task_url
+            )
+
+            live_run.client_io.launch_channels()
+        except (KeyboardInterrupt, Exception) as e:
+            logger.error(
+                "Encountered error while launching run, shutting down", exc_info=True
+            )
+            try:
+                live_run.architect.shutdown()
+            except (KeyboardInterrupt, Exception) as architect_exception:
+                logger.exception(
+                    f"Could not shut down architect: {architect_exception}",
+                    exc_info=True,
+                )
+            raise e
+
+        live_run.task_launcher.EXP_resume_assignments()
+        live_run.task_launcher.launch_units(url=task_url)
+
+        self._task_runs_tracked[task_run.db_id] = live_run
+        task_run.update_completion_progress(status=False)
+
+        return task_run.db_id
--- a/mephisto/operations/task_launcher.py
+++ b/mephisto/operations/task_launcher.py
@@ -264,3 +264,30 @@ class TaskLauncher:
             self.assignments_thread.join()
         if self.units_thread is not None:
             self.units_thread.join()
+
+    def EXP_resume_assignments(self) -> None:
+        """
+        Experimental function to go through an resume expired (or incomplete) tasks from a specific
+        task run that may have been left incomplete.
+        """
+        assignments = self.task_run.get_assignments()
+        for assignment in assignments:
+            self.assignments.append(assignment)
+            for unit in assignment.get_units():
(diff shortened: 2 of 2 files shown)


ajyl:
Can you point me towards how to created automated tests? In particular, I think I think
`Create a simple automated test that actually runs this functionality` would be a good place for me to familiarize myself with this PR.

JackUrb:
Surely! [Here's](https://github.com/facebookresearch/Mephisto/blob/42d86f2505ed3ab90625c7a75a45b39379d18741/test/core/test_operator.py#L203-L262) a good example of an operator test you could build from. The flow you'd look to build would likely need to:
1. Create a task run with 3 Assignments, each with 2 units
2. Create 3 agents using at least 2 workers (to make sure you can complete both of the first assignment, then one of the second). Lets say two for w0 and one for w1
3. Run the 3 agents to completion (with `send_agent_act` and `submit_mock_unit`). 
4. Quit the job (with `Operator.force_shutdown()`)
5. Check that one assignment is completed, one is mixed, and one is purely expired
6. Relaunch this task run
7. Complete one more unit as w0
8. Try to accept another as w0, assert that it fails (they've already done 3)
9. Complete two more units as w1
10. Assert that the operator exits (`_wait_for_runs_in_testing`)
11. Assert all three assignments are now completed.

ajyl:
Great! I'll likely take a stab at it over the weekend.
Noob question - How do I run these tests? Like any other python unittests, or do I need to do anything special?

JackUrb:
Just like any other unit test. I tend to use `pytest -k <my_test_name>` when writing a singular test case.

ajyl:
Hi - I have a new unit test created, where it's the exact copy of `test_run_job_not_concurrent`, and the only change I've made is changing `num_assignment` for `MockBlueprintArgs` from 1 to 3, and getting a huge traceback.
I think the main ones that are informative include the following:

```
___________________________________________________________________________________________________________________ ERROR at teardown of TestOperatorLocal.test_resume_job ___________________________________________________________________________________
________________________________                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                              
self = <test.core.test_operator.TestOperatorLocal testMethod=test_resume_job>                                                                                                                                                                                 
                                                                                                                                                                                                                                                              
    def tearDown(self):                                                                                                                                                                                                                                       
        if self.operator is not None:                                                                                                                                                                                                                         
            self.operator.force_shutdown(timeout=10)                                                                                                                                                                                                          
        self.db.shutdown()                                                                                                                                                                                                                                            shutil.rmtree(self.data_dir, ignore_errors=True)                                                                                                                                                                                                      
        SHUTDOWN_TIMEOUT = 10                                                                                                                                                                                                                                 
        threads = threading.enumerate()                                                                                                                                                                                                                       
        target_threads = [                                                                                                                                                                                                                                    
            t for t in threads if not isinstance(t, TMonitor) and not t.daemon                                                                                                                                                                                
        ]                                                                                                                                                                                                                                                     
        start_time = time.time()                                                                                                                                                                                                                              
        while len(target_threads) > 1 and time.time() - start_time < SHUTDOWN_TIMEOUT:                                                                                                                                                                        
            threads = threading.enumerate()                                                                                                                                                                                                                   
            target_threads = [                                                                                                                                                                                                                                
                t for t in threads if not isinstance(t, TMonitor) and not t.daemon                                                                                                                                                                            
            ]                                                                                                                                                                                                                                                             time.sleep(0.3)                                                                                                                                                                                                                                   
>       self.assertTrue(                                                                                                                                                                                                                                                  time.time() - start_time < SHUTDOWN_TIMEOUT,                                                                                                                                                                                                      
            f"Expected only main thread at teardown after {SHUTDOWN_TIMEOUT} seconds, found {target_threads}",                                                                                                                                                        )                                                                                                                                                                                                                                                     
E       AssertionError: False is not true : Expected only main thread at teardown after 10 seconds, found [<_MainThread(MainThread, started 140144099612480)>, <Thread(socket-thread-ws://localhost:3000/socket, started 140142232729344)>, <Thread(force-shut
down-cleanup-runs, started 140142213834496)>]   
```

```
ERROR    mephisto.utils.metrics:metrics.py:124 Could not launch prometheus metrics client, perhaps a process is already running on 3031? Mephisto metrics currently only supports one Operator class at a time at the moment
Traceback (most recent call last):                                                                                             
  File "/home/repos/Mephisto/mephisto/utils/metrics.py", line 122, in start_metrics_server                                                                                                                                                                    
    start_http_server(3031)                                                                                                    
  File "/home/repos/venv_mephisto/lib/python3.8/site-packages/prometheus_client/exposition.py", line 144, in start_wsgi_server                                                                                                                                
    httpd = make_server(addr, port, app, ThreadingWSGIServer, handler_class=_SilentHandler)                                                                                                                                                                   
  File "/usr/lib/python3.8/wsgiref/simple_server.py", line 154, in make_server                                                                                                                                                                                
    server = server_class((host, port), handler_class)                                                                         
  File "/usr/lib/python3.8/socketserver.py", line 452, in __init__                                                             
    self.server_bind()                                                                                                         
  File "/usr/lib/python3.8/wsgiref/simple_server.py", line 50, in server_bind                                                                                                                                                                                 
    HTTPServer.server_bind(self)                                                                                               
  File "/usr/lib/python3.8/http/server.py", line 138, in server_bind                                                           
    socketserver.TCPServer.server_bind(self)                                                                                   
  File "/usr/lib/python3.8/socketserver.py", line 466, in server_bind                                                          
    self.socket.bind(self.server_address)                                                                                      
OSError: [Errno 98] Address already in use  
```

I see that `test_run_jobs_with_restrictions` also launches 3 assignments, with 2 max units per worker. I'm able to run this test, and I'm guessing I want to move over some of the functionality in that test over as well, but I'm not sure what's missing/needed. 
I've tried some other combinations of configurations, but as far as I can tell, changing the number of assignments from 1 to 3 seem to be the culprit. Have you seen this error before?

JackUrb:
@ajyl Sorry I missed this for so long... The issue is that Mephisto in these tests expects to be running the tasks to completion in order to shutdown. If you've increased the number of assignments without any other changes, that means that the task isn't complete by the time it shuts down.

PR #911, opened 2022-10-04
Adding wrappers around Mephisto's sqlite3 DB access

JackUrb:
# Overview
Wraps the creation of `sqlite3` `Connection` objects such that, if any thread tries to access them from a different thread, we'll see this behavior occur. This covers all of the `sqlite3` connections inside of Mephisto, so @mojtaba-komeili if you still encounter the bug but aren't receiving additional logging, we'll need to look into your task-specific code.

# Testing
Automatic tests


mojtaba-komeili:
Thanks for implementing it. I will start my next task run on this branch. Hopefully this solves it, this time.

PR #918, opened 2022-10-12
Making MTurk cleanup run daily per requester

JackUrb:
# Overview
It very frequently comes up that Mephisto users and workers are left in positions where tasks seem to be broken. Usually this is the result of stale tasks, which surface from hasty shutdowns or strange expiration issues through `botocore`. The `mephisto scripts mturk cleanup` script exists to alleviate this problem, but given how frequently the issue arises I think it makes sense to build this functionality into the Mephisto run script when using an MTurk requester.

# Implementation
- Adds a new `try_prerun_cleanup` script to the `mturk_utils` that queries for presumed active (or broken HITs) and surfaces these if it has been greater than 24 hours since the last time one of these queries has been run for the given requester. It allows the user to select which if any HITs need disposing, then continues. It updates a file in `~/.mephisto/` to keep track of these times.
- Augments the `augment_config_from_db` helper function to run `try_prerun_cleanup` whenever an MTurk requester is being used.
- Adds a new option to the cleanup script to remove old tasks (> 2 weeks).
# Testing
I cleaned up some of my old tasks on one of my requesters using the new `(o)ld` option. 
I tried launching a task using one of my requesters:
**First time**
```
python static_test_script.py mephisto/provider=mturk mephisto.provider.requester_name=MYREQUESTER
...
It's been more than a day since you last ran a job with MYREQUESTER, checking for outstanding tasks...
That took a while! You may want to run `mephisto scripts cleanup mturk` later to clear out some of the older HIT types.

The requester MYREQUESTER has 2 outstanding HIT types, with 100 suspected active or broken HITs.
This may include tasks that are still in-flight, but also tasks have been improperly shut down and need cleanup.
 Please review and dispose HITs below.
HIT TITLE: ONE OF MY TASKS
LAUNCH TIME: 10/12/2022, 17:24:34
HIT COUNT: 20
Should we cleanup this hit type? (y)es or (n)o: 
>> n
HIT TITLE: ANOTHER ONE OF MY TASKS
LAUNCH TIME: 10/12/2022, 17:16:29
HIT COUNT: 80
Should we cleanup this hit type? (y)es or (n)o: 
>> y
Enter anything to confirm removal of the following HITs:
80 hits from 10/12/2022, 17:16:29 for HIT Type: ANOTHER ONE OF MY TASKS
# ctrl C here because I didn't want to remove this
```

**After first run**
```
python static_test_script.py mephisto/provider=mturk mephisto.provider.requester_name=MYREQUESTER
...
This task is going to launch live on mturk, press enter to continue: 
```

# Note

One can disable this functionality until a future date by updating the "last reviewed time" for a given requester to sometime in the future. If it becomes an issue, we will provide an option to `(d)efer` the check, however I think doing this check is important for worker quality.

### DIFF
--- a/mephisto/abstractions/providers/mturk/mturk_utils.py
+++ b/mephisto/abstractions/providers/mturk/mturk_utils.py
@@ -8,8 +8,10 @@ import boto3  # type: ignore
 import os
 import json
 import re
+import time
+from datetime import datetime
 from tqdm import tqdm  # type: ignore
-from typing import Dict, Optional, Tuple, List, Any
+from typing import Dict, Optional, Tuple, List, Any, TYPE_CHECKING
 from datetime import datetime
 
 from botocore import client  # type: ignore
@@ -19,10 +21,14 @@ from omegaconf import DictConfig
 
 from mephisto.data_model.qualification import QUAL_EXISTS, QUAL_NOT_EXIST
 from mephisto.utils.logger_core import get_logger, format_loud
-from mephisto.operations.config_handler import get_config_arg
+from mephisto.operations.config_handler import get_config_arg, DEFAULT_CONFIG_FOLDER
+
+if TYPE_CHECKING:
+    from mephisto.abstractions.database import MephistoDB
 
 logger = get_logger(name=__name__)
 
+CLEANUP_CHECK_TIME = 24 * 60 * 60  # Check every day!
 MTURK_TASK_FEE = 0.2
 MTURK_BONUS_FEE = 0.2
 SANDBOX_ENDPOINT = "https://mturk-requester-sandbox.us-east-1.amazonaws.com"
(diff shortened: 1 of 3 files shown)


mephisto/abstractions/providers/mturk/mturk_utils.py
```diff
@@ -738,3 +743,111 @@ def expire_and_dispose_hits(
...
+        print(
+            "That took a while! You may want to run `mephisto scripts "
+            "cleanup mturk` later to clear out some of the older HIT types.\n"
+        )
+    broken_hit_types = {
+        k: [h for h in v if hit_is_broken(h)]
+        for (k, v) in outstanding_hit_types.items()
+        if any([h for h in v if hit_is_broken(h)])
+    }
+    num_hit_types = len(broken_hit_types.keys())
```
mojtaba-komeili, line 793: Is it helpful to printout these types with their counts? Maybe some insight on something being wrong?

mojtaba-komeili: NM, I see it is down below.

mephisto/abstractions/providers/mturk/mturk_utils.py
```diff
@@ -738,3 +743,111 @@ def expire_and_dispose_hits(
...
+            hits_to_dispose += broken_hit_types[hit_type]
+            confirm_string += (
+                f"{hit_count} hits from {creation_time_str} for HIT Type: {cur_title}\n"
+            )
+
+    if len(hits_to_dispose) == 0:
+        print("No HITs selected for disposal. Continuing")
+        return
+
+    input(confirm_string)
```
mojtaba-komeili, line 838: Shouldn't we get user confirmation here?

mojtaba-komeili (approved):

mephisto/abstractions/providers/mturk/mturk_utils.py
JackUrb: I was a bit lazy here, using it as an opportunity to exit via `^C`. Let me put it in a `y`/`n` loop.

PR #919, opened 2022-10-12
12 hour inactivity shutdown

JackUrb:
# Overview
In the vein of #918 and trying to improve the negative impact of the long tail of failure cases, this feature sets a configurable `submission_patience`, which automatically shuts down a Mephisto `TaskRun` when no submissions have come through in (by default) 12 hours. This should capture when a task is available but can't be completed (due to a bug in the code, an onboarding that's impossible, pairing problems, etc) and take a task offline before it wastes workers' time.

# Implementation
- Adds a `no_submission_patience` argument to `TaskRunArgs`
- Adds a `last_submission_time` attribute to the `ClientIOHandler` which is updated every time a submission packet comes in.
- updates `Operator._track_and_kill_runs` to also see if the patience is exceeded as an option for shutting down a task.

# Testing
Current automated tests pass, and a new test is added for this case too.

### DIFF
--- a/test/core/test_operator.py
+++ b/test/core/test_operator.py
@@ -263,7 +263,7 @@ class OperatorBaseTest(object):
 
     @patch("mephisto.operations.operator.RUN_STATUS_POLL_TIME", 1.5)
     def test_patience_shutdown(self):
-        """Ensure that a job can be run that doesn't require connected workers"""
+        """Ensure that a job shuts down if patience is exceeded"""
         self.operator = Operator(self.db)
         config = MephistoConfig(
             blueprint=MockBlueprintArgs(num_assignments=1, is_concurrent=False),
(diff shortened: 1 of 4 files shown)


JackUrb:
@mojtaba-komeili this type of feature would have been useful for damage control on your issue (though hopefully #911 is helping on the exploration part of that front).

test/core/test_operator.py
```diff
@@ -261,6 +261,50 @@ def test_run_job_not_concurrent(self):
         assignment = task_run.get_assignments()[0]
         self.assertEqual(assignment.get_status(), AssignmentState.COMPLETED)
 
+    @patch("mephisto.operations.operator.RUN_STATUS_POLL_TIME", 1.5)
+    def test_patience_shutdown(self):
+        """Ensure that a job can be run that doesn't require connected workers"""
```
mojtaba-komeili, line 266: Not sure id this is the right docstring for this test function.

mojtaba-komeili (approved):
Thanks for adding this. It is needed a lot of times specially when running over weekends and there is hours of no activity.

test/core/test_operator.py
JackUrb: Good catch!

PR #921, opened 2022-10-25
[Bugfix] Fixing double update bug on heroku

JackUrb:
# Overview
An issue with how Heroku processes requests over sockets exposed that the current setup for dealing with duplicate messages wasn't sufficient. Running on heroku doubles _all_ messages in a way that doesn't occur on local or EC2. This PR includes updates to the frontend and backend that better check for duplicate `update_id` to fix the heroku experience.

@Kilichbek this should resolve your issue.

# Implementation
- in `socket_handler.jsx` we extract the `used_update_ids` _out_ from the state and into a `ref`, as the `parseSocketMessage` function that we register with the socket only has access to the first version of the state. Bumps to `2.0.3` based on this.
- in `ClientIOHandler` we add a new `seen_update_ids` `Set` which can be used to filter out incoming messages that are duplicate.
- Updated `./scripts/sync_package_version.sh` to include `packages`, as now some of these include `mephisto-task`
- Ran the above script to bump to `2.0.3`, and `sync_mephisto_task.py` as well to update on the python side.
- Ran `scripts/update_js_deps.sh` to push `2.0.3` to lockfiles and clear some vulns

# Testing
- Automated testing
- Deploying the `parlai_chat_task_demo` example on heroku before and after this change. Prior behavior shows duplicate messages both being sent and received, but new behavior no longer does.


pringshia:
So weird that Heroku does this and not EC2 and that a fix higher up the stack is needed. I wonder if there's anything we need to do to address a more root-cause issue, but glad to get this out since it's a severe issue.

One aside, since we are republishing the `mephisto-task` package, this might be a good opportunity to also relax the peerDependencies in it. Can I add that in to this PR?

pringshia (approved):

JackUrb:
> So weird that Heroku does this and not EC2 and that a fix higher up the stack is needed. I wonder if there's anything we need to do to address a more root-cause issue, but glad to get this out since it's a severe issue.

Also wondering about this, as I haven't been able to trace _why_ heroku specifically does the double messaging.

PR #925, opened 2022-11-07
Add additional python versions to test

### LINKED ISSUE #924: Mephisto incompatible with Python 3.11
# Overview

Thanks to a change in Python 3.11 (addressed here: https://github.com/facebookresearch/hydra/pull/2418) Mephisto currently doesn't work on Python 3.11. 

We should have some kind of grid-test for support on individual python versions (we have this somewhat) but _only_ run on `main` commits.

pringshia:
Closes #924 by introducing additional Python versions to the scheduled matrix testing.

### DIFF
--- a/.github/workflows/docker-testing-matrix.yml
+++ b/.github/workflows/docker-testing-matrix.yml
@@ -21,7 +21,7 @@ jobs:
       fail-fast: false # Run all combos, don't skip on failures
       matrix:
         # from: https://hub.docker.com/r/nikolaik/python-nodejs
-        python-version: ["3.9", "3.8"]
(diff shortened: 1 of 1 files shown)


JackUrb:
Once concern here: we'll be marked as failing entirely on `main` whenever this runs until the downstream issue in hydra<->python3.11 is resolved. It'd be nice to tie individual python versions to something we could tag on the github landing (like a badge) for each of these versions?

pringshia:
@JackUrb - we don't show tests on the main README anymore so this shouldn't be an issue. Also the downstream Hydra issue is now resolved anyway. Good to merge this in now?

PR #958, opened 2022-12-04
Annotator tracker

jayhuynh:


jayhuynh:
Sorry @JackUrb, but I'm not quite get what is the disclosure or `ActivityTrackerDisclosure`? And the task preview view you mean is the example task?

JackUrb:
The thought here is when giving instructions to a worker, I'm sure we have to _tell_ them that the entirety of their interactions with the task after they accept are being recorded. This is what I mean by a disclosure. Given the nature of collection here, it'd be useful to have a drop-in component someone can include in their instructions.

jayhuynh:
Hi @JackUrb. Thank you for suggestion. I think it's good to return a disclosure component. Just a quick question from your perspective that which appearance should I use for the disclosure? I'm thinking about a modal, tooltips or collapse. Or should it be a simple warning box and just want to make sure that I'm not overengineering it.

JackUrb:
Simple warning box should be fine!

jayhuynh:
Hi @JackUrb ,

I added ActivityTrackerDisclosure component to the package. Coud you please have a look at my pr? Thank you

JackUrb:
Hi @jayhuynh - this kind of component is **great**, only thing I'd suggest is that you include _default_ text (if no `children` are provided for instance) that inform the worker that their activity when interacting with the task is tracked over the complete duration of the task, not just the content that is submitted. Are you providing such a disclosure in your current work?

jayhuynh:
The default content for this component sounds great to me. Yes @JackUrb , we have the disclosure details in the preview panel but they are all plain text without styles. From now we can leverage ActivityTrackerDisclosure. We have also put a consent form link for supporting in this case.

I'll update the component and let you know.

jayhuynh:
Hi @JackUrb , this is the content of disclosure that we are using for our projects. Could you please have a look if we need any other adjusment? Thank you.

JackUrb (approved):
This looks great to me now, @pringshia do you have any comments?

JackUrb:
!! sorry for the delay here @jayhuynh, any cause to outright close it?

PR #966, opened 2022-12-20
Fixed Static task to not explicitly check for `objData.file1`

gzhihongwei:
I was doing some local development on a static task and for some reason the form wasn't submitting properly. Digging into the modified file, there is an implicit assumption that the user's task will have a file upload form input. Instead of explicitly checking, I changed it to see if `objData.file1` exists before accessing any further properties on that object.


JackUrb:
Ah nice catch! `StaticHTMLBlueprint` hasn't had a lot of use (or testing) compared to our other blueprints, and all the users I can think of have been leveraging file uploads in some form.

JackUrb (approved):

Issue #907, opened 2022-10-03
TypeError: cant_cancel_expirations() missing 1 required positional argument: 'frame'


jaraujo98:
I'm getting the error in the title when Ctrl-C'ing a running Mephisto process. I think the problem is that the `cant_cancel_expirations` signature should be `def cant_cancel_expirations(sig, frame):` instead of the current

https://github.com/facebookresearch/Mephisto/blob/f8bf7b1d3619d804855da0cd174ddfa16372b5f4/mephisto/operations/operator.py#L417-L421

From the [Python docs](https://docs.python.org/3/library/signal.html#signal.signal)

> The handler is called with two arguments: the signal number and the current stack frame (None or a frame object).

JackUrb:
Oh interesting, thanks for surfacing @jaraujo98. First thing of note is that you should _never_ try to ctrl-C a Mephisto process that is running unit expirations (which is what this is trying to capture), but I guess I never tried to force it. I'd be happy to accept a PR that drops the `self,` if you'd like to contribute directly, otherwise I'll address this in the next bug cleanup pass.

arjunan-k:
Hi @JackUrb I am new to Open Source. Can I work on this

PR #971, opened 2022-12-28
droped the self from cant_cancel_expirations

arjunan-k:
#907
`cant_cancel_expirations` signature should be `def cant_cancel_expirations(sig, frame):` instead of the current


JackUrb (approved):
Thanks for making this change @arjunan-k!

arjunan-k:
@JackUrb  my pleasure : )
Iooking forward to contribute more and something big in future : )

PR #989, opened 2023-01-13
Adding citation note to readme

JackUrb:
as titled


pringshia:
LGTM. I had also created a duplicate PR #990 but closing that for this. Should we remove the empty newlines in the citation or is that a part of the format? Also in my PR I also just got rid of the Changelog section as that's been empty forever.

pringshia (approved):

PR #1002, opened 2023-01-30
Reorganizing `pyproject.toml` to use Extras over Groups

### LINKED ISSUE #994: Slow install times
Dropping dependency groups in #898 is causing Mephisto installs to be very bloated. This is not great for new users and also especially noticeable in CI

- torch>=1.7.0 — **887 MB**
- nvidia-cudnn-cu11==8.5.0.96 — **557.1 MB**
- nvidia-cublas-cu11==11.10.3.66 — **317.1 MB**

are the main culprits

JackUrb:
# Summary
As Poetry doesn't easily support dependent installations with groups, I've swapped back to using extras. In this case, I still have a set of optional and non-optional installs for Mephisto roughly based on usage amount and size, however once we support many more blueprints/architects/providers we'll likely get to a point where we want other packaging. 

That, though, I'll leave as a future issue.

Resolves #994 so long as tests pass.


pringshia:
Thanks!

pringshia (approved):

PR #1005, opened 2023-02-08
fix rendering of assignment url for mock tasks

edwardguo61:
as titled, output correct urls when launched via ec2 for mock tasks


JackUrb (approved):
Great to have this change in - the hacks I've used in the past to get this URL will no longer be necessary. Thanks  @edwardguo61!

Issue #1009, opened 2023-02-17
Create `SurgeProvider`


JackUrb:
# Overview.

Like #1008, we're looking to create a new `CrowdProvider` for [SurgeAI](https://www.surgehq.ai/), relying on their [API docs](https://app.surgehq.ai/docs/api#surge-api). 

### Rough alignment so far:
Requesters => Probably just storing the Auth Key? [Authentication](https://app.surgehq.ai/docs/api#authentication)
Task/TaskRun => [Project](https://app.surgehq.ai/docs/api#projects) (roughly like we use HIT types and HITs)
Units => [Tasks](https://app.surgehq.ai/docs/api#tasks)
Workers => Roughly "Surgers", but no direct way to query. We mostly only have an ID, will need to figure out how to get it for `wrap_crowd_source.js`
Agents => Combination of [Reports](https://app.surgehq.ai/docs/api#reports), and live data from [Tasks](https://app.surgehq.ai/docs/api#tasks)
Qualifications => [Surger Groups](https://app.surgehq.ai/docs/api#surger-teams)

# Approach

The easiest way to work through this would be to clone the `MTurkProvider` and try adapting it, while also comparing to the base `CrowdProvider` class.

JackUrb:
No longer pursuing, as Surge hasn't pushed further on open access via API

PR #1011, opened 2023-03-02
Issue #1009: Add SurgeAI provider

meta-paul:


mephisto/abstractions/architects/local_architect.py
```diff
@@ -145,7 +145,12 @@ def deploy(self) -> str:
         self.running_dir = os.path.join(
             get_mephisto_tmp_dir(), f"local_server_{self.task_run_id}", "server"
         )
-        shutil.copytree(self.server_dir, self.running_dir, symlinks=True)
+
+        # TODO: Not sure that this is correct, but added `dirs_exist_ok=True`
+        #  as this line raised an exception that all these files already exist.
+        #  Maybe we need to remove it and recopy all the time.
+        #  Deployment should cleanup all directories it needs or use previous files
+        shutil.copytree(self.server_dir, self.running_dir, symlinks=True, dirs_exist_ok=True)
```
JackUrb, line 153: We should only be creating a single directory per task run, and it already existing should be considered an error.

Part of the rationale for retaining the old server was that we could examine the results for this task run in the exact setting they were collected in for debug reasons, even after the server has been shut down. 

If, for instance, you're restarting your database from scratch and using `dirs_exist_ok` you're going to potentially be populating builds that have files from completely different deploys inside.

Underneath, the issue you're adding this to resolve is one that's more fundamental and would need to be addressed at the core - Mephisto doesn't have a good `cleanup` method for `TaskRun`s (or `Requester`s for that matter), and we'd need to update a containerized version of Mephisto to entirely clean up all of the resources.

mephisto/abstractions/providers/surge_ai/surge_ai_utils.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def setup_credentials(
+    profile_name: str, register_args: Optional[ProviderArgs],
+) -> bool:
+    if not os.path.exists(os.path.expanduser(SURGE_AI_CONFIG_DIR)):
+        os.mkdir(os.path.expanduser(SURGE_AI_CONFIG_DIR))
+
+    with open(os.path.expanduser(SURGE_AI_CONFIG_PATH), 'w') as f:
+        f.write(register_args.api_key)
```
JackUrb, line 73: similar comment to the Prolific provider, we may need to support multiple concurrent surge accounts.

JackUrb (commented):
So far seems to be shaping up well. Will do a deeper dive here once I get the sandbox account setup. Have been asking their team about a few things that aren't quite straightforward (like using qualifications/teams to exclude workers, or the lack of being able to request existing budget information)

Issue #1008, opened 2023-02-17
Create `ProlificProvider`


JackUrb:
# Overview
The goal is to create a new `CrowdProvider` for [prolific](https://www.prolific.co/). Their API documentation is [here](https://docs.prolific.co/docs/api-docs/public/). While there are many things that directly correspond, some thought will need to go into the adapter on things that _don't_ directly line up.

Rough alignment so far:
Requesters => [Users](https://docs.prolific.co/docs/api-docs/public/#tag/Users)
Task/TaskRun => [Study](https://docs.prolific.co/docs/api-docs/public/#tag/Studies) (roughly like we use HIT types and HITs)
Workers => Roughly "participants", but no direct way to query. We mostly only have an ID, will need to figure out how to get it for `wrap_crowd_source.js`
Agents => Combination of [Submissions](https://docs.prolific.co/docs/api-docs/public/#tag/Submissions), [Bonuses](https://docs.prolific.co/docs/api-docs/public/#tag/Bonuses), and [Messages](https://docs.prolific.co/docs/api-docs/public/#tag/Messages)
Qualifications => [Participant Groups](https://docs.prolific.co/docs/api-docs/public/#tag/Participant-Groups)
Provider-specific Qualifications => [Requirements](https://docs.prolific.co/docs/api-docs/public/#tag/Requirements)

*Note:* we currently have `email_worker` in the codebase. We should likely also provide a `message_worker` and or thread-based interface, as while MTurk didn't have on-API messaging, many other providers seem to.

# Approach

The easiest way to work through this would be to clone the `MTurkProvider` and try adapting it.

jayhuynh:
Hi @JackUrb. We also have a plan to create `ProlificProvider` as we would like to use Prolific for some of our tasks. I'm happy to make a pr to contribute for this once it's ready.

JackUrb:
> Hi @JackUrb. We also have a plan to create `ProlificProvider` as we would like to use Prolific for some of our tasks. I'm happy to make a pr to contribute for this once it's ready.

Work is ongoing in #1012, you can view progress there.

meta-paul:
Completed in https://github.com/facebookresearch/Mephisto/pull/1012

PR #1012, opened 2023-03-13
Add Prolific provider (#1008)

meta-paul:


mephisto/abstractions/providers/prolific/prolific_utils.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def setup_credentials(
+    profile_name: str, register_args: Optional[ProviderArgs],
+) -> bool:
+    if not os.path.exists(os.path.expanduser(PROLIFIC_CONFIG_DIR)):
+        os.mkdir(os.path.expanduser(PROLIFIC_CONFIG_DIR))
+
+    with open(os.path.expanduser(PROLIFIC_CONFIG_PATH), 'w') as f:
+        f.write(register_args.api_key)
```
JackUrb, line 52: Note - one user may be working with _different_ prolific accounts at the same time. For this purpose, we may need to separate out the credentials into separate files (by the `profile_name`) or create a yaml/something to handle this.

mephisto/abstractions/providers/prolific/prolific_provider.py
```diff
@@ -0,0 +1,118 @@
...
+
+        for qualification in qualifications:
+            qualification_name = qualification['qualification_name']
+
+            if self.datastore.get_qualification_mapping(qualification_name) is None:
+                qualification['id'] = requester.create_new_qualification(qualification_name)
+
+        if hasattr(shared_state, 'prolific_specific_qualifications'):
+            # TODO(OWN) standardize provider-specific qualifications
+            qualifications += shared_state.prolific_specific_qualifications
```
JackUrb, line 99: As a note, qualifications are likely going to be the most unusual part of Prolific for Mephisto. While _we're_ happy to assign qualifications to users, and then make those into requirements or exclusions, Prolific has no way to directly do this.

They _do_ however have an allowlist/blocklist feature, and the allowlist and blocklist for a task can be updated dynamically. As such _we'll_ have to do the marginal qualification 'calculation's whenever one of the targeted qualifications has a change. I'm open to a number of directions, but I think the easiest would be:
1. Store the target qualifications that may affect a task in the `ProlificDatastore`
2. On a `Worker.<grant/revoke>_crowd_qualification(...)` call, check the datastore, and push updates to relevant allow/blocklists for active tasks.

mephisto/abstractions/providers/prolific/prolific_worker.py
```diff
@@ -0,0 +1,71 @@
...
+        """Determine if a worker is blocked"""
+        return self.datastore.get_worker_blocked(self.db_id)
+
+    def is_eligible(self, task_run: "TaskRun") -> bool:
+        """Determine if this worker is eligible for the given task run"""
+        return True
+
+    @staticmethod
+    def new(db: "MephistoDB", worker_id: str) -> "Worker":
+        return ProlificWorker._register_worker(db, worker_id + "_sandbox", PROVIDER_TYPE)
```
JackUrb, line 71: Note, I don't think Prolific [has a sandbox environment yet](https://docs.prolific.co/docs/api-docs/public/#tag/Testing), so our suggestions around this will be different.

We're getting around this by creating a regular and test account, but most won't have that ability upfront.

mephisto/abstractions/providers/prolific/prolific_utils.py
```diff
@@ -0,0 +1,136 @@
...
+        # TODO (FB-3): Implement
+        qualification = None
+    except Exception:  # TODO (FB-3): Implement
+        logger.exception(
+            f'Could not create a qualification with name "{qualification_name}" '
+            f'and description "{description}"'
+        )
+        raise
+
+    return qualification.id if qualification else None
```
JackUrb, line 112: As noted above, prolific qualifications for the most part are just going to be things _we_ keep track of locally, but that aren't really pushed out.

This in the short term will mean that teams that are working on projects together will potentially have different lists for this purpose. Next on the agenda after providers+architects is importing/exporting qualifications for Mephisto, so we'll move to resolve that issue soon enough.

pyproject.toml
```diff
@@ -84,6 +84,8 @@ tornado = { version = "^6.0", optional = false }
 parlai = { version = "^1.7.0", optional = true }
 torch = { version = "^1.4.0", optional = true }
 
+# jsonschema
```
JackUrb, line 87: Comment is generally the purpose for having the library. Here it's likely "api validation"?

mephisto/abstractions/providers/prolific/wrap_crowd_source.js
```diff
@@ -0,0 +1,119 @@
...
+    if (numErrorsCaught > 1) {
+      userDisabledErrorPrompts = confirm(
+        "We noticed you have gotten more than one of these error alerts. Would you like to hide these alerts going forward?"
+      );
+      console.log(userDisabledErrorPrompts);
+    }
+  } else {
+    // should we do auto-submission of errors?
+  }
+});
```
JackUrb, line 118: @pringshia is there a target for moving these (and the `event_emitter`) into the base `mephisto_task` library? I don't imagine these change provider-to-provider, and they're bloating these files.

JackUrb (commented):
Overall structure looks good. Big fan of having the `prolific_api` wrapped into a useful utility. I'm not sure it should sit in base `mephisto` given it should only be used in mephisto core through the `abstractions` layer.

Added some early thoughts about `Qualification`s in Prolific's API, since these will likely be the hardest to map.

mephisto/abstractions/providers/prolific/wrap_crowd_source.js
pringshia: We needed the even emitter to enable functionality for all tasks including non-react static tasks #692, and so the emitter was introduced via #879. I don't know how else we'd add it. As is now,  `mephisto-task` is mostly a React-based package for React-based tasks.

mephisto/abstractions/providers/prolific/prolific_agent.py
```diff
@@ -0,0 +1,117 @@
...
+        """Put observations into this Prolific agent's observation list"""
+        self.datastore.agent_data[self.db_id]["observed"].append(live_update)
+        super().observe(live_update)
+
+    def get_live_update(self, timeout=None) -> Optional[Dict[str, Any]]:
+        """
+        Either take an act from this mock agent's act queue (for use
+        by tests and other mock purposes) or request a regular act
+        (for use in manual testing).
+        """  # TODO (#1008)
```
JackUrb, line 70: The MTurk version will be a useful comparison

mephisto/abstractions/providers/prolific/prolific_agent.py
```diff
@@ -0,0 +1,117 @@
...
+
+        Prolific Units
+        """
+        self.update_status(AgentState.STATUS_APPROVED)
+
+    def reject_work(self, reason) -> None:
+        """
+        Reject the work done on this specific Unit
+        """
+        self.update_status(AgentState.STATUS_REJECTED)
```
JackUrb, line 92: Will need to forward this status to Prolific (and we should have a SOFT_REJECT option too)

mephisto/abstractions/providers/prolific/prolific_datastore.py
```diff
@@ -0,0 +1,390 @@
...
+    is_expired BOOLEAN
+);
+"""
+
+CREATE_WORKERS_TABLE = """
+CREATE TABLE IF NOT EXISTS workers (
+    worker_id TEXT PRIMARY KEY UNIQUE,
+    is_blocked BOOLEAN
+);
+"""
```
JackUrb, line 38: I don't think we'll need a workers table, no real worker-specific information to store (outside of qualifications, which fit in the main DB)

mephisto/abstractions/providers/prolific/prolific_datastore.py
JackUrb, line 1: Generally I think the [`MTurkDatastore`](https://github.com/facebookresearch/Mephisto/blob/e3f837baabd1a6767a556e3ba5f2c94e2cd3e24a/mephisto/abstractions/providers/mturk/mturk_datastore.py) is more analogous to the kind of information we'll be storing. The mock datastore needed to keep _everything_, as there was no external provider holding the information. We don't need to duplicate information that can be easily queried.

JackUrb (commented):
Overall solid progress here, the provider is really beginning to take shape! Some comments

mephisto/abstractions/architects/ec2/ec2_architect.py
```diff
@@ -52,6 +52,16 @@ def url_safe_string(in_string: str) -> str:
     return re.sub("[^0-9a-zA-Z-]+", "", hyphenated)
 
 
+def get_full_domain(args: DictConfig) -> str:
```
JackUrb, line 55: Is there a specific use for extracting this out? The architect still does this in `__init__`, but I see where knowing this information without setting up an `__init__` could be useful. If so, just add a comment for the function and then use it in the `__init__` call below.

Edit: found the usage, I don't think it's necessary to have used in the place it is, but the util could still be useful.

mephisto/abstractions/providers/prolific/api/README.md
```diff
@@ -0,0 +1,461 @@
+## Prolific Python SDK
```
JackUrb, line 1: 👍 This is really solid, definitely think it's quality enough to end up in the Prolific codebase too

mephisto/abstractions/providers/prolific/api/data_models/__init__.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python3
```
JackUrb, line 1: We could really use this kind of setup in the Mephisto data model. `from mephisto.data_model.agent import Agent` is definitely redundant where `from mephisto.data_model import Agent` would not be. 

Not a TODO for right now of course, but something you have me thinking about.

mephisto/abstractions/providers/prolific/api/eligibility_requirement_classes/base_eligibility_requirement.py
```diff
@@ -0,0 +1,61 @@
+import inspect
+
+from omegaconf import ListConfig
+
+
+class BaseEligibilityRequirement:
```
JackUrb, line 6: Awesome base class and usage. Should this class itself be marked as abstract?

mephisto/abstractions/providers/prolific/prolific_agent.py
```diff
@@ -0,0 +1,150 @@
...
+        self.worker: 'ProlificWorker' = cast('ProlificWorker', self.get_worker())
+
+    def _get_client(self) -> prolific_api:
+        """Get a Prolific client"""
+        requester: 'ProlificRequester' = cast('ProlificRequester', self.unit.get_requester())
+        return self.datastore.get_client_for_requester(requester.requester_name)
+
+    @property
+    def log_prefix(self) -> str:
+        return f'[Agent {self.db_id}] '
```
JackUrb, line 62: I like this inclusion a lot for debugging

mephisto/abstractions/providers/prolific/prolific_datastore.py
```diff
@@ -0,0 +1,527 @@
...
+                if db_prolific_qualification_name != prolific_participant_group_name:
+                    logger.warning(
+                        f'Prolific Qualification mapping create for {qualification_name} '
+                        f'with Prolific name {prolific_participant_group_name}, '
+                        f'already exists under {db_prolific_qualification_name}.'
+                    )
+
+                return None
+            else:
+                raise e
```
JackUrb, line 428: We'll have to revisit these as part of the qualification implementation. Generally we'll have task-run <-> (allow participant group, deny participant group) mappings, and so these functions will be adapted for that.

mephisto/abstractions/providers/prolific/prolific_datastore_tables.py
```diff
@@ -0,0 +1,68 @@
...
+        is_registered BOOLEAN
+    );
+"""
+
+CREATE_UNITS_TABLE = """
+    CREATE TABLE IF NOT EXISTS units (
+        unit_id TEXT PRIMARY KEY UNIQUE,
+        is_expired BOOLEAN
+    );
+"""
```
JackUrb, line 37: These values are stored in the main database, and can otherwise be queried from Prolific or local state, so no need to record them here too

mephisto/abstractions/providers/prolific/prolific_datastore_tables.py
```diff
@@ -0,0 +1,68 @@
...
+
+CREATE_WORKERS_TABLE = """
+    CREATE TABLE IF NOT EXISTS workers (
+        worker_id TEXT PRIMARY KEY UNIQUE,
+        is_blocked BOOLEAN
+    );
+"""
```
JackUrb, line 44: Similarly, blocked status is pushed to Prolific directly, so we don't need to store a copy.

mephisto/abstractions/providers/prolific/prolific_unit.py
```diff
@@ -0,0 +1,365 @@
...
+        # TODO (#1008): Fix this logic. It looked abstract and unclear in Mturk code.
+        if self.db_status == AssignmentState.CREATED:
+            return super().get_status()
+        elif self.db_status in [
+            AssignmentState.ACCEPTED,
+            AssignmentState.EXPIRED,
+            AssignmentState.SOFT_REJECTED,
+        ]:
+            # These statuses don't change with a get_status call
+            return self.db_status
```
JackUrb, line 123: These are final status that won't be changing, once they're stored in the db that's it.

mephisto/abstractions/providers/prolific/prolific_unit.py
```diff
@@ -0,0 +1,365 @@
...
+                    found_status = AssignmentState.REJECTED
+                elif agent_status == AgentState.STATUS_SOFT_REJECTED:
+                    found_status = AssignmentState.SOFT_REJECTED
+            else:
+                logger.warning(f'Agent for unit {self} is None')
+
+            if found_status != self.db_status:
+                self.set_db_status(found_status)
+
+            return self.db_status
```
JackUrb, line 146: These are statuses where the status of the unit is determined by what we determine the agent status. Essentially, after a task is no longer live it's units are all in one of:
`[AssignmentState.ACCEPTED, AssignmentState.EXPIRED, AssignmentState.SOFT_REJECTED, AssignmentState.COMPLETED, AssignmentState.REJECTED]`
You can think of these roughly as `ACCEPTED` and `SOFT_REJECTED` corresponding with a "paid complete" status, which cannot be changed (can't unpay!), `COMPLETED` and `REJECTED` as "unpaid complete" (wherein payment may still occur, can reverse rejections in MTurk at least), and `EXPIRED` as "unpaid incomplete" (but can not be paid out, so it's final).

As the status for `COMPLETED` and `REJECTED` may change if we approve an agent, we need to check with the agent.

mephisto/abstractions/providers/prolific/prolific_unit.py
```diff
@@ -0,0 +1,365 @@
...
+        study = prolific_utils.get_study(client, prolific_study_id)
+
+        if study is None:
+            return AssignmentState.EXPIRED
+
+        local_status = self.db_status
+        external_status = self.db_status
+
+        if study.status == StudyStatus.UNPUBLISHED:
+            external_status = AssignmentState.COMPLETED
```
JackUrb, line 172: `AssignmentState.LAUNCHED`

mephisto/abstractions/providers/prolific/prolific_unit.py
```diff
@@ -0,0 +1,365 @@
...
+        elif study.status == StudyStatus.ACTIVE:
+            external_status = AssignmentState.LAUNCHED
+        elif study.status == StudyStatus.SCHEDULED:
+            # TODO (#1008): Choose correct mapping
```
JackUrb, line 176: `AssignmentState.LAUNCHED`

mephisto/abstractions/providers/prolific/prolific_unit.py
```diff
@@ -0,0 +1,365 @@
...
+            pass
+        elif study.status == StudyStatus.PAUSED:
+            # TODO (#1008): Choose correct mapping
```
JackUrb, line 179: `AssignmentState.LAUNCHED` (for Mephisto purposes at least)

mephisto/abstractions/providers/prolific/prolific_unit.py
```diff
@@ -0,0 +1,365 @@
...
+            pass
+        elif study.status == StudyStatus.COMPLETED:
+            external_status = AssignmentState.COMPLETED
+        else:
+            raise Exception(f'Unexpected Study status {study.status}')
+
+        # -------------------------------------------------------------------------
+        # TODO (#1008): Fix this logic. It looked abstract and unclear in Mturk code.
+        if external_status != local_status:
+            if local_status == AssignmentState.ASSIGNED and external_status in [
```
JackUrb, line 193: if locally we think it's assigned, but prolific thinks it's launched, it must have been returned (or expired).

mephisto/abstractions/providers/prolific/prolific_utils.py
```diff
@@ -0,0 +1,547 @@
...
+        return None
+
+    found_workspace, workspace = _find_prolific_workspace(prolific_api, title=workspace_name)
+
+    if not found_workspace:
+        logger.error(f'Could not find a workspace with name {workspace_name}')
+        return None
+
+    try:
+        workspace_ballance: WorkspaceBalance = prolific_api.Workspaces.get_balance(id=workspace.id)
```
JackUrb, line 89: Nit: spelling

mephisto/abstractions/providers/prolific/prolific_utils.py
```diff
@@ -0,0 +1,270 @@
...
+
+    return project.id if project else None
+
+
+def delete_qualification(client: prolific_api, id: str) -> bool:
+    """
+    Delete a qualification (Prolific Participant Group) by ID
+    :param id: Prolific Participant Group's ID
+    """
+    # TODO (#1008): Implement later. There's no DELETE method for Prolific Participant Groups
```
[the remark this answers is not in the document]
JackUrb: Now that we'll be making ad-hoc participant groups, we'll likely have to ask them about adding a `DELETE` method here.

mephisto/abstractions/providers/prolific/prolific_unit.py
```diff
@@ -0,0 +1,365 @@
...
+
+        logger.debug(f'{self.log_prefix}Launching Task Run {task_run_id} with data: {args}')
+
+        # As Task Run was registered in datastore before, we can get its details from it
+        task_run_details = self.datastore.get_run(task_run_id)
+        logger.debug(f'{self.log_prefix}Task Run datastore details: {task_run_details}')
+
+        # Set up Task Run (Prolific Study)
+        logger.debug(f'{self.log_prefix}Creating Prolific Study')
+        prolific_study: Study = prolific_utils.create_study(
```
JackUrb, line 281: you have the base `task_url` here, no need to re-get it in `create_study`

JackUrb (commented):
Progress here looks really good, leaving a bunch of comments but mostly on things we knew would be updated down the road.

mephisto/abstractions/providers/prolific/prolific_datastore.py
```diff
@@ -0,0 +1,591 @@
...
+                        f"Prolific Qualification mapping create for {qualification_name} "
+                        f"with Prolific name {prolific_participant_group_name}, "
+                        f"already exists under {db_prolific_qualification_name}."
+                    )
+
+                return None
+            else:
+                raise e
+
+    def create_quailfication_mapping(
```
JackUrb, line 462: nit: sp

mephisto/abstractions/providers/prolific/prolific_datastore.py
```diff
@@ -0,0 +1,591 @@
...
+                WHERE qualification_name = ?
+                """,
+                (qualification_name,),
+            )
+            results = c.fetchall()
+            if len(results) == 0:
+                return None
+            return results[0]
+
+    def create_participant_proup_mapping(
```
JackUrb, line 389: nit: sp

mephisto/abstractions/providers/prolific/prolific_provider.py
```diff
@@ -0,0 +1,366 @@
...
+
+logger = get_logger(name=__name__)
+
+
+@dataclass
+class ProlificProviderArgs(ProviderArgs):
+    """Base class for arguments to configure Crowd Providers"""
+
+    _provider_type: str = PROVIDER_TYPE
+    requester_name: str = PROVIDER_TYPE
```
JackUrb, line 67: This should be `OmegaConf.MISSING` or None.

mephisto/abstractions/providers/prolific/prolific_worker.py
```diff
@@ -0,0 +1,262 @@
...
+
+        p_worker_id = self.get_prolific_worker_id()
+        prolific_utils.give_worker_qualification(client, p_worker_id, p_qualification_id)
+
+        logger.debug(
+            f'{self.log_prefix}Crowd qualification {qualification_name} has been granted '
+            f'for Prolific Participant "{p_worker_id}"'
+        )
+
+        return None
```
JackUrb, line 223: This whole block is the last clear remaining portion to get qualifications working. Here, we should be querying the `ProlificDatastore` to see if the given qualification (will need to query the DB for the qual ID) is being used for any allow/blocklist participant groups for a `TaskRun`. If it's not, we early exit. If it is, we need to check to see if this worker _now qualifies_ under the rules for that `TaskRun`, we edit the group here.

JackUrb (commented):
Took another pass, a few more nits and a comment on the last visible remaining step for Qualifications that I see.

examples/simple_static_task/static_test_prolific_script.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python3
```
JackUrb, line 1: Now that this is all functional, I'm not sure what the _correct_ thing for us to be doing to demonstrate different providers will be, but I don't think it's going to be to have a different script in here.

mephisto/abstractions/database.py
```diff
@@ -983,6 +984,23 @@ def delete_qualification(self, qualification_name: str) -> None:
...
+
+    @FIND_GRANT_QUALIFICATION_LATENCY.time()
+    def find_granted_qualifications(
+        self, worker_id: Optional[str] = None,
+    ) -> List[GrantedQualification]:
+        """
+        Find granted qualifications.
+        If `worker_id` is not supplied, returns all granted qualifications.
+        """
+        return self._find_granted_qualifications(worker_id=worker_id)
```
JackUrb, line 1002: This is definitely a nomenclature issue from me from way back. We should be using the signature of `check_granted_qualifications` over a `worker_id` only setup. I don't really know why I made the 'find' for `granted_qualifications` "`check`". Can you have `check_granted_qualifications` point to this as well?

mephisto/abstractions/providers/prolific/api/client.py
```diff
@@ -0,0 +1,74 @@
...
+    class Wrapper(target_cls):
+        @classmethod
+        def _base_request(cls, *args, **kwargs):
+            new_args = {k: v for k, v in kwargs.items()}
+            if new_args.get("api_key", None) is None:
+                new_args["api_key"] = api_key
+            return super()._base_request(*args, **new_args)
+
+    # Name wrapper after wrapped class (for logging)
+    Wrapper.__name__ = target_cls.__name__
```
JackUrb, line 42: Nice addition

mephisto/abstractions/providers/prolific/prolific_worker.py
```diff
@@ -0,0 +1,315 @@
...
+        logger.debug(f'{self.log_prefix}Task Run: {task_run}')
+
+        task_run_args = task_run.args
+        requester = cast('ProlificRequester', requester)
+        client = self._get_client(requester.requester_name)
+        prolific_utils.unblock_worker(client, task_run_args, self.worker_name, reason)
+        self.datastore.set_worker_blocked(self.worker_name, is_blocked=False)
+
+        # Include unblocked worker into all currently running studies, if qualified
+        self._grant_crowd_qualification(client, requester, task_run_args.provider)
```
JackUrb, line 175: it's not inherently clear what this `_grant_crowd_qualification` call is doing, given no qualification name appears provided.

mephisto/abstractions/providers/prolific/prolific_agent.py
JackUrb: This is managed by the underlying `Agent`, we mark as approved and then update our local status: https://github.com/facebookresearch/Mephisto/blob/25c2594be19100abd40037bf9bb43d8a3a87c6ca/mephisto/data_model/agent.py#L585

mephisto/abstractions/providers/prolific/prolific_unit.py
JackUrb: Ah no, I'm explaining the logic. You're saying to fix it in the TODO

mephisto/abstractions/providers/prolific/prolific_provider.py
```diff
@@ -0,0 +1,218 @@
...
+                'https://example.com?'
+                'participant_id={{%PROLIFIC_PID%}}'
+                '&study_id={{%STUDY_ID%}}'
+                '&submission_id={{%SESSION_ID%}}'
+                'where `prolific_pid`, `study_id`, `submission_id` are params we use on our side, '
+                'and `{{%PROLIFIC_PID%}}`, `{{%STUDY_ID%}}`, `{{%SESSION_ID%}}` are their '
+                'format of template variables they use to replace with their IDs'
+            ),
+        },
+    )
```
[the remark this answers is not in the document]
JackUrb: Leaving it as a setting is somewhat strange though as we expect providers to be able to know how to consume an Architect's server details and then construct the task. Leaving it a public setting means we expect people would need to change it.

mephisto/abstractions/providers/prolific/prolific_provider.py
```diff
@@ -0,0 +1,218 @@
...
+    prolific_project_name: str = field(
+        default=DEFAULT_PROLIFIC_PROJECT_NAME,
+    )
+    prolific_allow_list_group_name: str = field(
+        default=DEFAULT_PROLIFIC_GROUP_NAME_ALLOW_LIST,
+    )
+    prolific_block_list_group_name: str = field(
+        default=DEFAULT_PROLIFIC_GROUP_NAME_BLOCK_LIST,
+    )
```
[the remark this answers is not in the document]
JackUrb: I'm mostly referring here to the standard use of:
```
SharedState.qualifications = [
    ...
]
```
But I think what you're getting at is use of Prolific's own allowlist and blocklist functionality, separate from using participant groups as a qualification requirement.

mephisto/abstractions/providers/prolific/prolific_unit.py
JackUrb: From the perspective of Mephisto, I'm not sure there's a difference for an assignment being `PAUSED` and being `LAUNCHED`. Active studies under `LAUNCHED` would mean that we haven't yet pushed the unit live in Mephisto (for rate limiting reasons, usually), and paused studies being `LAUNCHED` would also warrant them being inaccessible.

Also, shouldn't UNPUBLISHED correspond to EXPIRED? I'm not sure I understand the semantics there.

mephisto/abstractions/providers/prolific/prolific_provider.py
JackUrb: `requester_name` already defaults missing here: https://github.com/facebookresearch/Mephisto/blob/25c2594be19100abd40037bf9bb43d8a3a87c6ca/mephisto/abstractions/crowd_provider.py#L25
I think that's what's making the error get thrown, though for a wrong reason.

mephisto/abstractions/providers/prolific/prolific_worker.py
JackUrb: I just mean we generally don't have a need to be creating qualification groups ad-hoc from a worker, most of the qualification granting part comes from what we store locally in order to construct the `TaskRun`-specific participant groups.

examples/simple_static_task/static_test_prolific_script.py
JackUrb: Agreed, I think we need to have another way to demonstrate provider use in a script. We can commit here for now, but it's a TODO item.

mephisto/abstractions/database.py
JackUrb: We'll have to go through deprecation steps. Make `find_granted_qualifications` a direct duplicate of `check_granted_qualifications`, and make the latter point to the former. We can then add a pending deprecation warning.

mephisto/abstractions/providers/prolific/prolific_datastore.py
JackUrb: Definitely, in that case we'll just use the local blocklist.

mephisto/abstractions/providers/prolific/prolific_worker.py
```diff
@@ -0,0 +1,315 @@
...
+            )
+            logger.debug(f'Created new Participant Group "{prolific_participant_group.id}"')
+            prolific_utils.give_worker_qualification(
+                client, prolific_participant_id, prolific_participant_group.id,
+            )
+
+        logger.debug(
+            f'{self.log_prefix}Crowd qualification {qualification_name} has been granted '
+            f'for Prolific Participant "{prolific_participant_id}"'
+        )
```
JackUrb, line 257: Moving [previous thread](https://github.com/facebookresearch/Mephisto/pull/1012#discussion_r1273686458) to up-to-date codeblock. The confusion here is asking when _this_ branch would ever be occurring. We're only calling `grant_crowd_qualification` with the purpose of altering participant groups for in-flight `TaskRun`s under the current implementation, unless the goal here is to have a route to add `ProlificWorkers` to Prolific-backed participant groups with this call.

mephisto/abstractions/providers/prolific/prolific_unit.py
```diff
@@ -0,0 +1,344 @@
...
+                    found_status = AssignmentState.SOFT_REJECTED
+            else:
+                logger.warning(f'Agent for unit {self} is None')
+
+            if found_status != self.db_status:
+                self.set_db_status(found_status)
+
+            return self.db_status
+
+        # Remaining statuses are tracking a live Study
```
JackUrb, line 147: To match our intended semantics it should be tracking a live _submission_.

mephisto/abstractions/providers/prolific/prolific_unit.py
```diff
@@ -0,0 +1,344 @@
...
+                    agent.update_status(AgentState.STATUS_EXPIRED)
+
+            # Can't determine anything else if there is no Study on this unit
+            return self.db_status
+
+        requester: 'ProlificRequester' = self.get_requester()
+        client = self._get_client(requester.requester_name)
+
+        # time.sleep(2)  # Prolific servers may take time to bring their data up-to-date
+        study = prolific_utils.get_study(client, prolific_study_id)
```
JackUrb, line 164: And you should write `get_submission`, as the statuses of a submission should align here:
Screenshot 2023-07-27 at 11 49 50 AM

PR #1018, opened 2023-04-24
Added AWS credentials pre-commit hook

meta-paul:
Implementation of https://gist.github.com/czardoz/b8bb58ad10f4063209bd

Note that correct regexp for AWS keys should be taken from here https://github.com/awslabs/git-secrets


JackUrb (approved):
Thanks for including

PR #1035, opened 2023-07-26
Experimental Maximum Wait Time for Task Launch

Guest400123064:
# __Experimental Maximum Wait Time for Task Launch__

## __Background__
In our experiment to collect conversational data between two human agents, a non-trivial portion of time is observed where one (or more) of the agents are not connected to the chatting session, resulting in the task remaining un-launched and wasting the commitment of the counterparties. To address this issue, we implemented a timeout feature that allows Mephisto to launch a task with an incomplete list of agents. Subsequently, we can add the corresponding number of chatbots and continue collecting conversations.

### __High-level Design__
The core idea is to set a maximum wait time when the first agent is onboarded. If all slots are filled (i.e., after the last agent is onboarded), the task will launch as usual. However, if there are still empty slots, the first onboarded agent will be responsible for launching the task with an incomplete list of agents.

### __Technical Details__
To set the timer, we simply use `asyncio.sleep(max_launch_timeout)` to control the maximum wait time when the first human agent connects to the session. Afterward, the control is handed back to the event loop, allowing other agents to onboard. Additionally, we track the task IDs in `WorkerPool._waiting_assignments: Set[int]`. The task ID will be removed from this set once launched. This ensures that if tasks are launched with a full list of agents (by the last joined human agent), the first agent assignment function will be notified after waking from `asyncio.sleep(max_launch_timeout)`, preventing the first agent from launching the task a second time.

We also introduce a new `TaskRunArgs` called `max_launch_timeout` (default set to zero) that can be configured as follows,
```
mephisto:
    task:
        max_launch_timeout: 42  # In seconds
```

## __Sample Use case__
We utilized this feature in combination with `parlai_chat` during the world creation stage, specifically in the `make_world` function. During this process, we checked the number of agents being passed to this function and appended chatbots to complete the list up to a pre-defined number of agents.

## __Tests__
When set to no timeout (default value), the code remains compatible with existing functionalities. Although we did not create dedicated test cases for the new functionality, we extensively employed it in our experiment and successfully collected roughly 20 conversations without encountering errors.


JackUrb:
This is awesome to see implemented! Overall I think the code makes sense, I'll take another pass later this week. I'd also like to see a test though. You should be able to copy from one of our existing tests, say this one:
https://github.com/facebookresearch/Mephisto/blob/25c2594be19100abd40037bf9bb43d8a3a87c6ca/test/core/test_operator.py#L143

Rather than submitting two agents, your test can simply register one of the agents, and assert that the task is able to complete as expected.

Ideal would be also providing a test where you set the timeout, but then supply both agents anyway and assert the behavior still works.

Guest400123064:
Hi Jack,

Thanks for your response! I think someone also mentioned the issue about `test_operator.py` in [this PR](https://github.com/facebookresearch/Mephisto/pull/910#issuecomment-1271083203). I encountered the same problem of the thread not being terminated after 10 seconds timeout when running `pytest` locally. But I don't quite understand your comment in the PR. So, if possible, would you mind elaborating more on how to avoid this issue when testing? Thanks!

JackUrb:
Sure - the basic underlying idea is that, when the test is completed, we expect resources to be properly cleaned up. You can resolve this by calling `operator.force_shutdown(timeout=0.2)` yourself before the test ends.

Guest400123064:
Hi Jack, I am a bit confused about how to avoid the `AssertionError` resulting from `tearDown`. I always get `Expected only main thread at teardown after 10 seconds...`, for example, in the `TestOperatorSingleton::test_run_job_concurrent` test case. Where should I put `operator.force_shutdown(timeout=0.2)`? Thanks!

JackUrb:
You get the `AssertionError` even when running the tests as-is without changing?

JackUrb:
Interesting, I can't reproduce on main with:
```bash
$ pytest test/core/test_operator.py 
```

And the test is also passing here for your PR. I can't tell why this wouldn't be working for you.

Guest400123064:
That is interesting... But maybe I can ignore the error for now, as it is only caused by the `tearDown` process. The actual tests do not spit any errors. Maybe it is a WSL problem.

Guest400123064:
Hi Jack, I found a challenge in testing timeout handling with `MockTaskRunner` because, unlike the `ParlAITaskRunner`, [it checks for](https://github.com/Guest400123064/mephisto-timeout/blob/d4deac91cf31ba20327ce021f9c405374172bfd3/mephisto/abstractions/blueprints/mock/mock_task_runner.py#L87) the `unit.get_assigned_agent() is not None` instead of checking `agent is not None` directly. However, my implementation will pass the reduced agent list to the task runner while leaving the units unassigned. Therefore, the `MockTaskRunner` refuse to proceed. Also, for some reason, the assertion error is not captured by the unittest when I use `self.assertRaises(AssertionError)` but the error is displayed in `Captured stdout call`. Do you have any suggestions regarding this situation? Thanks in advance!

JackUrb:
Sorry for the delay here! We should probably update the MockTaskRunner to no longer do this strange agent_id remapping, as this was a fix to a bug that no longer exists.

From: Yuxuan Wang ***@***.***>
Date: Friday, August 4, 2023 at 2:21 AM
To: facebookresearch/Mephisto ***@***.***>
Cc: Jack Urbanek ***@***.***>, Comment ***@***.***>
Subject: Re: [facebookresearch/Mephisto] Experimental Maximum Wait Time for Task Launch (PR #1035)
Hi Jack, I found a challenge in testing timeout handling with MockTaskRunner because, unlike the ParlAITaskRunner, it checks for the unit. get_assigned_agent() is not None instead of checking agent is not None directly. However, my implementation
ZjQcmQRYFpfptBannerStart
This Message Is From an External Sender
ZjQcmQRYFpfptBannerEnd

Hi Jack, I found a challenge in testing timeout handling with MockTaskRunner because, unlike the ParlAITaskRunner, it checks for<https://github.com/Guest400123064/mephisto-timeout/blob/d4deac91cf31ba20327ce021f9c405374172bfd3/mephisto/abstractions/blueprints/mock/mock_task_runner.py#L87> the unit.get_assigned_agent() is not None instead of checking agent is not None directly. However, my implementation will pass the reduced agent list to the task runner while leaving the units unassigned. Therefore, the MockTaskRunner refuse to proceed. Also, for some reason, the assertion error is not captured by the unittest when I use self.assertRaises(AssertionError) but the error is displayed in Captured stdout call. Do you have any suggestions regarding this situation? Thanks in advance!

Guest400123064:
Hi Jack,

Thank you very much for your reply! So, should I wait for the update or should I comment out the code from [L87](https://github.com/Guest400123064/mephisto-timeout/blob/d4deac91cf31ba20327ce021f9c405374172bfd3/mephisto/abstractions/blueprints/mock/mock_task_runner.py#L87) to L92, which re-generates the list of agents, and skip `None` agents within the `for loop` starting from L93? Thanks in advance!

Best,
Yuxuan Wang

Guest400123064:
Science I haven't been following the recent updates for a long time, I want to start with the latest version and submit another PR in the future to make sure the code are up to date.

PR #1039, opened 2023-07-31
quick fix: change order of checks in operator

PReithofer:
First check get_is_completed to make sure that agent and unit status is synced even if generator already finished.


JackUrb (approved):
Accepted as a quick fix. Certainly brings to question whether this state syncing should be owned by `_track_and_kill_runs` in a poll-only manner. Given recent attempts to clean up state (#1037), it's possible that we end up in a better place soon.

PR #1049, opened 2023-08-17
Updated Prolific integrations docs

meta-paul:


JackUrb (approved):
LGTM! On top of having better usage docs, I'm glad to have the oddities documented such that we don't lose that information with time.

wise-east:
I don't think this is searchable in https://mephisto.ai/ or https://mephisto.ai/python_api/mephisto.html.

PR #1058, opened 2023-08-25
Create a reusable app for task review

meta-paul:
This review app supports all functionality of the old command-line script (and adds some convenience), while providing a simple easy-to-use UI.
 
For guidelines on how to run the new Review App and make any Task code work with its interface, see `mephisto/client/review_app/README.md`.

Main features:

- Intuitive UI that includes:
  - rendered display of the performed Task unit page
  - grouping unit results by worker, with an option of batch review
  - statistics of reviewed units, for better decision-making
- Integration with any existing/future Task (see guidelines)
- Recording of reviewer actions in Mephisto DB


mephisto/client/review_app/server/api/views/stats_view.py
```diff
@@ -0,0 +1,191 @@
...
+            limit=limit,
+        )
+        all_unit_reviews = _find_units_for_worker(
+            db=app.db,
+            worker_id=worker_id,
+            task_id=task_id,
+            statuses=AssignmentState.completed(),
+            since=since,
+            limit=limit,
+        )
```
JackUrb, line 174: Would it be better to get all the units at once, then filter by status? Unsure if this is going to incur any additional search cost or not depending on the DB.

mephisto/client/review_app/server/api/views/tasks_worker_units_view.py
```diff
@@ -0,0 +1,43 @@
...
+from typing import List
+
+from flask import current_app as app
+from flask.views import MethodView
+
+from mephisto.abstractions.databases.local_database import StringIDRow
+from mephisto.data_model.unit import Unit
+
+
+class TasksWorkerUnitsView(MethodView):
```
JackUrb, line 16: This is a pretty confusing name, unsure what it's getting at by reading it.

mephisto/client/review_app/server/api/views/units_approve_view.py
```diff
@@ -0,0 +1,56 @@
...
+            if unit_ids and int(unit.db_id) not in unit_ids:
+                continue
+
+            agent = unit.get_assigned_agent()
+            if not agent:
+                raise BadRequest(f"Cound not approve Unit \"{unit.db_id}\".")
+
+            agent.approve_work()
+
+            db_queries.create_unit_review(
```
JackUrb, line 46: We'll need to do some thinking about how unit reviews will work, as they should also be applied if I just call `agent.approve_work()` outside of the context of this review flow.

mephisto/client/review_app/server/api/views/units_details_view.py
```diff
@@ -0,0 +1,63 @@
...
+            if unit_ids and int(unit.db_id) not in unit_ids:
+                continue
+
+            try:
+                unit_data = app.data_browser.get_data_from_unit(unit)
+            except AssertionError:
+                # In case if this is Expired Unit. It raises and axceptions
+                unit_data = {}
+
+            units.append(
```
JackUrb, line 52: I think that, for the most part, this should end up just being `units.append(unit_data)`. Ultimately though I see what you're saying, this is going to require a significant change to how `AgentState` specifies data, at least insofar as providing separate `inputs` (which will be what Mephisto normally sends to `initialTaskData` when a task is being done, and `outputs` (which can be _anything else_ that `AgentState.get_data` wants to include in the final data). This should be standardized across Mephisto, which unfortunately will incur us setting up a migration (either a util for going from the old formats to the standard, or temporarily dealing with both).

Then `TaskFrontend` can by default accept `initialTaskData` and `taskOutputs` or something.

mephisto/abstractions/blueprints/remote_procedure/remote_procedure_blueprint.py
```diff
@@ -101,6 +101,13 @@ class RemoteProcedureBlueprintArgs(
             "required": True,
         },
     )
+    task_source_review: str = field(
+        default=MISSING,
```
JackUrb, line 105: We should drop the required tag here - otherwise this seems like a great add and means folks can even register multiple different review flows that attach to specific runs. 

Downstream there could be value in exporting these bundles to somewhere `mephisto/tmp` such that data collected with a specific bundle could reuse that exact corresponding review bundle, but that's not necessary to do right now.

mephisto/abstractions/providers/mock/mock_agent.py
```diff
@@ -73,20 +73,45 @@ def get_live_update(self, timeout=None) -> Optional[Dict[str, Any]]:
...
         """
         self.update_status(AgentState.STATUS_APPROVED)
 
-    def reject_work(self, reason) -> None:
+        if not skip_unit_review:
+            unit = self.get_unit()
+            self.db.new_unit_review(
+                unit_id=int(unit.db_id),
+                task_id=int(unit.task_id),
+                worker_id=int(unit.worker_id),
```
JackUrb, line 94: The cast to int should occur on the db side, as the DB is what is determining that these are ints vs being random string IDs.

mephisto/client/review_app/server/api/views/units_approve_view.py
```diff
@@ -0,0 +1,68 @@
...
+from mephisto.data_model.worker import Worker
+
+
+class UnitsApproveView(MethodView):
+    def post(self) -> dict:
+        """Approve worker's input"""
+
+        data: dict = request.json
+        unit_ids: Optional[str] = data.get("unit_ids") if data else None
+        feedback = data.get("feedback") if data else None  # Optional
```
JackUrb, line 24: OK I think there's confusion here that could be improved on the documentation/nomenclature side:
`feedback` elsewhere in `AgentState` refers to a _worker_'s feedback about a task to the requester. We don't use the `feedback` field very frequently though, but use here and changes elsewhere do introduce a bit of chance for confusion.

JackUrb (commented):
Now this is really looking like the integration we really want in the long run! I'll have to update tasks from my future paper work to use this review flow rather than the setup I originally hobbled together

examples/remote_procedure/mnist_for_review/webapp/webpack.config.review.js
```diff
@@ -0,0 +1,48 @@
...
+      },
+      {
+        test: /\.css$/,
+        use: ["style-loader", "css-loader"],
+      },
+      {
+        test: /\.(svg|png|jpe?g|ttf)$/,
+        loader: "url-loader",
+        options: { limit: 100000 },
+      },
```
JackUrb: Let's add `.tsx` loader by default too
```

      {
        test: /\.tsx?$/,
        use: 'ts-loader',
        exclude: /node_modules/,
      },
```

mephisto/client/review_app/README.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+where
+
+- `-h`/`--host` - host address (optional, default: `"127.0.0.1"`)
+- `-p`/`--port` - port (optional, default: `5000`)
+- `-d`/`--debug` - debug (optional, default: `None`)
+
+
+#### Run Client application
```
JackUrb: Is there any way for us to have an option to launch both the server and client in one command?

mephisto/abstractions/blueprints/remote_procedure/remote_procedure_blueprint.py
```diff
@@ -101,6 +101,13 @@ class RemoteProcedureBlueprintArgs(
             "required": True,
         },
     )
+    task_source_review: str = field(
+        default='',
+        metadata={
+            "help": "Path to file containing javascript bundle for the task review application",
+            "required": False,
+        },
+    )
```
JackUrb: We'll also need this in the other base blueprints now

examples/remote_procedure/mnist/README.md
```diff
@@ -1,11 +1,9 @@
-<!---
-  Copyright (c) Meta Platforms and its affiliates.
```
JackUrb, line 2: Are licenses not required for `.md`?

examples/remote_procedure/mnist/webapp/cypress.config.js
```diff
@@ -1,9 +1,3 @@
-/*
- * Copyright (c) Meta Platforms and its affiliates.
```
JackUrb, line 2: Weird license drops

examples/remote_procedure/mnist/webapp/cypress/e2e/remote_procedure_mnist.cy.js
```diff
@@ -1,9 +1,3 @@
-/*
- * Copyright (c) Meta Platforms and its affiliates.
```
JackUrb, line 2: 📜

examples/remote_procedure/mnist/webapp/link_mephisto_task.sh
```diff
@@ -1,7 +1,2 @@
 #!/bin/sh
-
```
JackUrb, line 2: 📜

examples/remote_procedure/mnist/run_task.py
```diff
@@ -52,7 +52,7 @@ def validate_screening_unit(unit: Unit):
     agent = unit.get_assigned_agent()
     if agent is not None:
         data = agent.state.get_data()
-        annotation = data["outputs"]["final_submission"]["annotations"][0]
+        annotation = data["final_submission"]["annotations"][0]
```
JackUrb, line 55: Is this change intentional? I thought the standardization to having agent state's data contain `'inputs'` and `'outputs'` was part of the overall plan.

examples/simple_static_task/static_test_mturk_script.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python3
```
JackUrb, line 1: I still don't fully understand having the mturk example - it should at the very least be sandbox only

mephisto/README.md
```diff
@@ -4,12 +4,259 @@
...
+8. Launch a small batch with a chosen crowd provider to see how real workers handle your task.
+9. Iterate more.
+10. Collect some good data.
+
+---
+
+## Configure Task parameters
+
+This is a sample YAML configuration to run your Task on **AWS EC2** architect with **Prolific** provider
+
```
JackUrb, line 187: This still remains a combination that is only really relevant for FAIR. This should be an internal document on the wiki, as the advice is really quite good. For external, we'd probably still suggest heroku.

mephisto/README.md
```diff
@@ -4,12 +4,259 @@
   LICENSE file in the root directory of this source tree.
 -->
 
-# Mephisto
```
JackUrb, line 7: The new contents in the readme overall are _great_, thanks for extending this with so much detail. (hopefully enough to get people over the docker hump).

mephisto/abstractions/providers/mturk/mturk_utils.py
```diff
@@ -610,10 +610,10 @@ def approve_work(client: MTurkClient, assignment_id: str, override_rejection: bo
         )
 
 
-def reject_work(client: MTurkClient, assignment_id: str, reason: str) -> None:
+def reject_work(client: MTurkClient, assignment_id: str, review_note: Optional[str] = None) -> None:
```
JackUrb, line 613: Calling this `review_note` is a bit strange, as a note doesn't necessarily imply it will be given as feedback to the worker. In retrospect, `reason` didn't really do this well either. Perhaps `feedback`?

mephisto/abstractions/providers/prolific/prolific_agent.py
```diff
@@ -100,7 +100,12 @@ def new_from_provider_data(
 
         return super().new_from_provider_data(db, worker, unit, provider_data)
 
-    def approve_work(self) -> None:
+    def approve_work(
+        self,
+        review_note: Optional[str] = None,
+        bonus: Optional[str] = None,
```
JackUrb, line 106: having `bonus` here is a bit strange, as it implies that adding a `bonus` on this call will trigger bonusing the worker. (which I don't believe it does.

mephisto/client/review_app/README.md
```diff
@@ -0,0 +1,84 @@
+## Run TaskReview app
+
+
```
JackUrb, line 3: Good opportunity for a screenshot - demonstrates value of a UI more easily when it can be seen.

examples/remote_procedure/mnist/run_task.py
meta-paul: Yes, I had to fix that to get the Task Review front-end to work for mnist.

examples/simple_static_task/static_test_mturk_script.py
meta-paul: This is just a helpful starting point for Mephisto users to use and build their own mturk configs from. We have examples how to use YAML configs for mock and prolific providers, but not for mturk. (And I should also add "mturk_example.yml" into hydra_configs folder to better support this.)

mephisto/README.md
meta-paul: I can surely move this to internal docs. The goal here was to provide a simple guide supporting a typical workflow for our FAIR team members. Will look for a good landing place for this instruction.

mephisto/README.md
meta-paul: Without Docker it would quite painful to troubleshoot everyone's local environment :D

mephisto/abstractions/providers/mturk/mturk_utils.py
meta-paul: That is correct, I renamed it because it may or may not be sent to worker (and "feedback" implies that it does get sent).

In current Mturk implementation, `review_note` is sent to worker if HIT was rejected; but reviewer may want to leave a note for their own records for an accepted HIT as well. In that case, `review_note` will only be kept in datastore, and not sent to the worker.

mephisto/client/review_app/README.md
meta-paul: Sure, I can attach something there

examples/remote_procedure/mnist/run_task.py
JackUrb: You should update the front-end to index into outputs rather than this, such that we can have a more standard format/expectations on where to find data across different `AgentState` types: #1065

examples/simple_static_task/static_test_mturk_script.py
JackUrb: My main concern here is actually that we're beginning to couple the examples of how to use Mephisto on different providers with our `static_html` task, which is somewhat just a holdover for folks who have legacy HTML tasks they want to port.

I think maybe the route is for us to have a subfolder in examples for `provider_integrations` which would capture this. The task could be a near-duplicate to the `static_react_task`.

mephisto/README.md
JackUrb: Maryam likely has good advice on where would be appropriate to store this documentation.

mephisto/README.md
JackUrb: Agreed - it's good to get people to adopt the most reproducible route, but it's hard to get people to change! In trying to get someone to adopt a new framework (Mephisto) over whatever they're doing before, it's important for adoption to _also_ support something closer to a workflow your users are already using (notebooks, local environment, generally _not_ docker).

mephisto/abstractions/providers/mturk/mturk_utils.py
JackUrb: Right, the issue is that you'd **rather** imply that it's sent than not. Imagine someone starts using this for local notes like "this task was garbage" and that ends up pushed to the worker.

Even better would be to have _explicit_ labelling for whether the note is pushed or not.

mephisto/abstractions/providers/prolific/prolific_agent.py
meta-paul: Hi Jack, I did face a dilemma here. By design, TaskReviewApp should allow user to bonus workers during task review. But by implementation, bonusing so far has been done via launching a script.

For now I picked a middle approach: I'm saving user-provided bonuses in the datastore, so that later a script could pick them up and issue these bonuses if needed. The bonusing script will need to be adjusted then (I didn't want to touch it as it's not really part of a review app).

We can go one step further, and properly grant bonuses with the provider right away inside the "approve_work" method. I can add that code for Prolific, and probably could do that for mturk as well. 

IMO the best approach would be to support both scripted or "live" bonusing (that choice will be in TaskRun args). Bonus could be saved in the datastore, and we'd need to add an extra column "is_paid" so that live and script bonusing can work side-by-side as needed for a partuclar TaskRun.

Let me know which route makes most sense.

mephisto/abstractions/providers/mturk/mturk_utils.py
meta-paul: > Even better would be to have explicit labelling for whether the note is pushed or not.

That's a good idea, I'll add that option in the UI

examples/remote_procedure/mnist/webapp/cypress/e2e/remote_procedure_mnist.cy.js
meta-paul: Updated copyright headers (re-ran updated script, and it added a few more actually)

examples/remote_procedure/mnist/run_task.py
meta-paul: Thanks for pointing this out, I've updated this as well. I've missed screening flow when updating the code. As for the front-end, no changes needed - we either send the entire "task_data", or (as in TaskReviewApp) the UI is aware of "inputs" and "outputs".

mephisto/README.md
meta-paul: Moved that part out of the readme

mephisto/abstractions/providers/mturk/mturk_utils.py
meta-paul: I've added that option, and disabled it (along with bonusing) right away, because back-end is not ready for that yet. Once back-end logic is added (could be in v1.2.1), I'll re-enable the UI for it. So for now users can leave comment, and it will only be saved in the database.

mephisto/abstractions/providers/prolific/prolific_agent.py
meta-paul: For now I've disabled bonusing, so we can re-enable it with the next patch release (As per https://github.com/facebookresearch/Mephisto/pull/1058#discussion_r1428433620)

Issue #1060, opened 2023-09-13
__heroku_app_url empty when using heroku 


gaotianyu1350:
It seems that heroku changed their output and the current way of getting the URL with the hash will get an empty string. 

The code at [this line](https://github.com/facebookresearch/Mephisto/blob/9f1b73ae8f9b5351eb9e54cc185dec65d5e03c5d/mephisto/abstractions/architects/heroku_architect.py#L450) should be changed to 

```python
self.__heroku_app_url = domains.split("\n")[2]
```

JackUrb:
Thanks for catching @gaotianyu1350, I have a patch up that should hopefully resolve this more robustly.

PR #1062, opened 2023-09-13
Update heroku_architect.py

JackUrb:
Addressing #1060.

Uses a linear search rather than direct selection to get the first non-empty app name.

cc @gaotianyu1350


meta-paul:
@JackUrb sorry this PR flew under the radar. Should I rebase it on dev prior to merging?

JackUrb:
This fix can go straight to main

PR #1066, opened 2023-09-17
[WIP] Add unit scheduling strategies

PReithofer:
It would be nice to be able to select the unit scheduling strategy. Right now one can only manipulate this by changing the order of the assignments data input. This however has some drawbacks:

* It is not possible to schedule on an unit basis. Only the order of whole assignments can be changed.
* It is not possible to change the order after handing the assignment data to mephisto. This is a problem when using the generator pattern. Here the best you can do to get a more random schedule would be to buffer the assignments for a while and then randomize these batches.

Some scheduling strategies that could make sense:

 1. First in first out: this is the one mephisto is using right now
 2. Last in first out: this could be useful when using the generator pattern and generating new assignments based on previously submitted ones
 3. Random: also nice to have if there is some pattern in the order in which the assignments are generated
 4. Round Robin: this could be useful if one fears that not all units might be completed and prefers to have an even distribution of the completed units over the assignments.

I have implemented strategies 1-3 and it seems to work. However I'm unsure about whether my implementation makes sense and haven't written unit tests for them. 

Would greatly appreciate feedback!

### DIFF
--- a/mephisto/data_model/task_run.py
+++ b/mephisto/data_model/task_run.py
@@ -125,6 +125,26 @@ class TaskRunArgs:
             )
         },
     )
+    unit_scheduling_strategy: str = field(
+        default="FIFO",
+        metadata={
+            "help": (
+                "Strategy that determines the scheduling strategy for units."
+                "Can be 'FIFO', 'LIFO' or 'Random'."
+            )
+        }
+    )
+    scheduler_prefer_assigned_assignments: bool = field(
+        default=True,
+        metadata={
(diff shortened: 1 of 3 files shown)


JackUrb (commented):
This is a really cool feature. I'll hold off on merging until we're able to add some testing for this, but I can definitely see the value here!

meta-paul:
Moved these commits to new PR https://github.com/facebookresearch/Mephisto/pull/1107

PR #1075, opened 2023-10-12
Update copyright headers across the codebase

meta-paul:
This PR comprises a script to auto-update copyright headers `scripts/update_copyright_headers.py`, its output, and a few other manual changes.


scripts/update_copyright_headers.py
```diff
@@ -0,0 +1,216 @@
...
+        lines = f.readlines()
+
+    if len(lines) < 2:
+        raise UnsupportedFile("File has fewer than one line")
+
+    # Check copyright presence at the top of the file
+    anchor_line_number = -2
+    for i, line in enumerate(lines[:EXAMINED_LINES]):
+        text = line[:50].lower()
+        if "copyright" in text:
```
mojtaba-komeili, line 91: Why do we need this branch, could we merge it with the next one?

scripts/update_copyright_headers.py
```diff
@@ -0,0 +1,216 @@
...
+            anchor_line_number = -1
+        if anchor_line_number == -1 and ("meta" in text or "facebook" in text):
```
mojtaba-komeili: Could it be `"copyright" in text` instead  of `anchor_line_number == -1` and removing the above line?

mojtaba-komeili (approved):
The script makes sense and the outcome is what we expected. LGTM.

Feel free to ignore my other comments. It made more sense after seeing the rest of the code.

mojtaba-komeili:
Is that possible to fix the tests, maybe in a follow up PR?

scripts/update_copyright_headers.py
meta-paul: Agree, this logic could be written better (I abused integer line counter by using it as a status variable). Will push an update.

scripts/update_copyright_headers.py
meta-paul: Updating copyright is a different task from creating review app, or other tasks within release v1.2, so I've decided to create a separate branch. In addition, it has a large diff, and resolving potential conflicts is easier from a separate branch.

meta-paul:
Unittests are broken here (not because copyright script broke the code, but rather it's been a long-standing issue). They will be fixed in a separate branch, and it's not an easy task.

meganung (approved):
offline discussion, lgtm !

Issue #773, opened 2022-05-02
[Feature Request] Push exclusionary qualifications to crowdproviders when exceeding `maximum_units_per_worker`


JackUrb:
# Overview
At the moment, we have a less-than-ideal worker experience when using `maximum_units_per_worker`, as workers are able to see these tasks appear as available in their queues, but after they accept them they're unable to work on them. Instead, it would be great if we could push a qualification to these workers that prevents them from being eligible for the task altogether when they exceed this count.

# Idea
One idea we can use is a local qualification to hold the number of tasks currently completed by an individual worker on a given `task_name` (`task_id`). We can construct this at the creation of any particular `task_id` (ie, `completed_units_for_<task_name>`). Then whenever a `Unit` is completed, we can increment this qualification value. 

When a `TaskRun` is launched with `maximum_units_per_worker` set, we create a qualification for `<task_name>_<task_run_id>_exceeded_maximum`. We can then query all of the `completed_units_for_<task_name>` qualifications, and assign this disqualification to these workers.

If upon completion of a `Unit`, a worker now exceeds the `maximum_units_per_worker` value for a particular task, we can assign the qualification during the task as well.

# Possible Implementation
1. For any new `task_id`, we should generate a `completed_units_for_<task_name>` qualification. We can do this by implementing a `get_completed_units_qualification` method for the `Task` class, relying on `utils.qualifications.find_or_create_qualification`.
2. The `Operator` class should call `Task.get_completed_units_qualification()` to ensure the qualification is initialized. 
3. After creating the `TaskRun`, if `maximum_units_per_worker` is set, the `Operator` should call a new method `utils.qualifications.create_exceed_max_for_task(TaskRun, max_units)`. It should create the `<task_name>_<task_run_id>_exceeded_maximum` qualification, then query for all of the workers with the `completed_units_for_<task_name>` greater than the provided value, and assign it to them.
4. To cover any `Unit` completion, the `Agent` class in `handle_submit` should call a `WorkerPool` method to increment the value for the `completed_units_for_<task_name>` qualification for its worker. **NOTE** this qualification increment should use `bypass_crowd`. If the new value exceeds the `TaskRun`'s `maximum_units_per_worker` variable, it should assign the exceeded max qualification. The methodology for using such a method from the `Agent` class can be seen in the `update_status` method (`if self.agent_in_active_run(): ... live_run.loop_wrap.execute_coro(...`)

PR #1080, opened 2023-10-19
Prevent oversampling of study submissions by any single worker

meta-paul:
For some studies (i.e. Tasks spanning multiple TaskRuns) it may be desired to limit number of submissions received from the same worker (for example, to prevent bias in annotation tasks dataset).

The new TaskRunArgs parameter `max_submissions_per_worker` addresses this need.

### DIFF
--- a/mephisto/abstractions/providers/mturk/mturk_utils.py
+++ b/mephisto/abstractions/providers/mturk/mturk_utils.py
@@ -439,7 +439,8 @@ def create_hit_type(
                 has_locale_qual = True
         locale_requirements += existing_qualifications
 
-    if not has_locale_qual and not client_is_sandbox(client):
+    is_sandbox = client_is_sandbox(client)
+    if not has_locale_qual and not is_sandbox:
         allowed_locales = get_config_arg("mturk", "allowed_locales")
         if allowed_locales is None:
             allowed_locales = [
@@ -457,6 +458,9 @@ def create_hit_type(
                 "ActionsGuarded": "DiscoverPreviewAndAccept",
             }
         )
+
+    if is_sandbox:
+        hit_reward = 0
 
     # Create the HIT type
     response = client.create_hit_type(
--- a/mephisto/abstractions/providers/prolific/api/base_api_resource.py
+++ b/mephisto/abstractions/providers/prolific/api/base_api_resource.py
@@ -99,8 +99,6 @@ class BaseAPIResource(object):
             else:
                 result = response.json()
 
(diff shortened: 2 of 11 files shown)


meta-paul:
@JackUrb @xing-liu This is a feature to prevent oversampling. Not part of TaskReviewApp feature, but due to the size of that other PR I've used that branch as the base for now.

JackUrb:
Is this the same as [`maximum_units_per_worker`](https://github.com/facebookresearch/Mephisto/blob/642202de3db170e9bc3c4e67c6a3c245aedac800/mephisto/data_model/task_run.py#L101)?

meta-paul:
@JackUrb Similar idea, but for a different provider setup. In Prolific we don't know ahead of time who will be working on a unit before releasing it. So sidelining an overcontributing worker should be done differently (via participant group, not at the time of unit assignment because we have no control over that).

JackUrb:
I see - I think in general we should actually try to merge this concept. In #773 I noted that the current way of doing things (based on a count and a "no access" page) wasn't great, and that qualification-based would be preferred. I don't think we should have two separate arguments for this.

meta-paul:
@JackUrb I can attempt at parameter unification there. Should we do this before merging TaskReview app code, or it can be done later?

JackUrb:
I don't think it should be merged as is, and also don't think this feature should block task review.

meta-paul:
@JackUrb I've nested my logic under your paramtere name `maximum_units_per_worker`. That way one parameter will cover both use cases of Mturk and Prolific. Let me know if that makes sense.

_(By the way, disregard failing styling check. Not sure why it's failing, as if Github is running on another version of the code - couldn't find failing lines within this branch)_

JackUrb:
This consolidation now makes a lot more sense. Just be sure to test it out both on MTurk and prolific.

mojtaba-komeili (approved):
Had a quick look and it seems reasonable to me. I also see Jack is onboard with the changes. Feel free to merge after adding some tests please.

JackUrb (approved):
Overall this looks good to me. Thanks for unifying the logic between Prolific and MTurk providers, I think the flow is more obvious now for both.

One future TODO (#773) would be to push an exclusionary qualification to MTurk in the same way you currently do to prolific for qualification groups. This could be an option for a v1.3 feature, and would involve creating per-task-run exclusionary qualifications by default for each MTurk `TaskRun`. This PR makes the flow for that feature request more straightforward.

PR #1092, opened 2023-12-15
Release v1.2

meta-paul:


JackUrb:
We'll want to get doc deployment + code style squared away before this goes in. Unsure why we have some test failings here too.

meta-paul:
@JackUrb Sure thing! I usually leave fixing of failing checks to the very last stage, and I've just done that. Everything is passing now.

JackUrb (approved):
LGTM

Glad to have this together! Sorry for the review delays

PR #1107, opened 2024-03-07
Add unit scheduling strategies

meta-paul:
Added unit scheduling strategies for units while during Task Run, as per original contribution here: https://github.com/PReithofer/Mephisto/tree/add-unit-scheduling-strategies

---

It would be nice to be able to select the unit scheduling strategy. Right now one can only manipulate this by changing the order of the assignments data input. This however has some drawbacks:

- It is not possible to schedule on an unit basis. Only the order of whole assignments can be changed.
- It is not possible to change the order after handing the assignment data to mephisto. This is a problem when using the generator pattern. Here the best you can do to get a more random schedule would be to buffer the assignments for a while and then randomize these batches.

Some scheduling strategies that could make sense:

1. **First in first out:** this is the one mephisto is using right now
2. **Last in first out:** this could be useful when using the generator pattern and generating new assignments based on previously submitted ones
3. **Random:** also nice to have if there is some pattern in the order in which the assignments are generated
4. **Round Robin:** this could be useful if one fears that not all units might be completed and prefers to have an even distribution of the completed units over the assignments.

This PR implements strategies 1-3.

Would greatly appreciate feedback!


meta-paul:
Unfortunately, this feature will have to be moved onto the wishlist. After further developing the contributed code, we've realized that it would work only in one particular case: when `task.max_num_concurrent_units` parameter is set to `None`. But if that parameter has a small value (remember, value of `1` is required to efficiently work with the Prolific provider), then prioritization should be happening not (only) at the stage of reserving a unit, but (also) at the stage of composing `available_units` dict, i.e in the method `TaskLauncher._create_single_assignment`.

So in the scenario with `task.max_num_concurrent_units: 1`, the suggested code logic prioritizes unit selection from a dict of 1 element (and so the prioritization won't make any difference). And applying prioritization strategy at the stage of launching units will require quite a bit of extra work.

